PR #44, opened 2014-06-12
update setup.py for it to actually work (sorry!)

rep:
So sorry guys, the setup.py does not actually work as it does not pull in the sub-packages. Embarassing...

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -6,5 +6,5 @@ setup(name='Google Certificate Transparency',
       version='0.9',
       description='Certificate Transparency python client and monitor library',
       url='https://github.com/google/certificate-transparency',
-      packages=['ct'],
+      packages=['ct', 'ct.crypto', 'ct.crypto.asn1', 'ct.client'],
      )


mschloesser-r7:
I'm now updating our parsing setup and was checking on this PR and other issues I had previously. Sorry for not answering earlier.

The change from this PR is necessary for the setup.py file to make sense. I double checked just now that it works as I originally intended to. The setup.py itself file was pulled in earlier to the CT repo from my branch. This is really just a bugfix.

I'm not sure what the other comment is referring to with the _pb2.py files.

pphaneuf:
Once you run `make`, you'll have some generated Python source files in `python/ct/proto`, ending with `_pb2.py`. I think @eranmes was asking whether `ct.proto` should thus be in this list, and I was asking if they were to be added, if the `setup.py` file can generate those files?

But I agree that even if this PR doesn't address *all* the issues, at the minimum, it is making things better, so even if our comments are correct, it might be worth merging this PR as-is, and fix those issues in follow-up PRs?

mschloesser-r7:
Ah okay now I understand - thanks. It's actually not completely obvious whether we should add this code generation part to setup.py because then it would need the appropriate dependencies as well. As it is currently the setup.py will work even on a clean system with just pure python and it can then parse certificates with CT.

Due to the missing files there it might not support the complete normal featureset. But if that's needed one could also just go the standard ```make``` way?

pphaneuf:
If there are some useful bits that work without the proto files, I'd say we're good. Adding `protoc` calls to the `setup.py` sounds like a massive headache, if someone needs this, they can use `make`. :smiley: 

I'll merge this in shortly.

eranmes:
Ideally the code generation would be a part of setup.py (either by blatantly copying it from the Makefile or invoking make) - IMHO setup.py should fail if it can't compile the protobufs:
* Very few things will work without the compiled protobufs - the project's Python code is always written with the assumption in mind that they would be availabel.
* Poor user experience: When the code will fail a user will see import failures for an internal package without any clue that this is because the protobuf compiler missing.

@pphaneuf  - I actually prefer not to have a setup.py at all than having one that partially installs our bindings.

mschloesser-r7:
@eranmes fair enough - and I agree mostly. For my use-case i don't need the protobuf part - I can just keep it in a local branch on my side then.

I guess the best way would be to warn during installation if it fails and ask the user whether it should continue.

Feel free to close the PR - but then you should remove the setup.py alltogether.

pphaneuf:
Had a quick discussion with @eranmes, and I feel that it's true it's not a great user experience if they need those protobuf-dependent bits, but it's already there, and removing it would be an even worse experience for those already using the other bits successfully (like you). Might as well make it work a bit better...

pphaneuf:
Merged as 97036423ff0f30166b40eff724578b17a9a72a30.

PR #266, opened 2014-11-24
Ensure that the initial update is complete before sending further updates

AlCutter:


pphaneuf:
I think this pull request might no longer be relevant?

PR #297, opened 2014-12-03
Allow other threads to pause producer.

laiqu:
Without this change, if any thread other than reactors thread tries to resume production everything fails (produce is called is simply running in non reactor thread so it's output won't be handled etc.).


laiqu:
Yeah, it can be discarded

PR #359, opened 2014-12-18
Add ct_server_public_key flag

iangcarroll:
`ct` will fail to upload if you do not specify a public key for the log server.

### DIFF
--- a/cpp/client/upload_server_cert.sh
+++ b/cpp/client/upload_server_cert.sh
@@ -6,21 +6,21 @@ set -e
 
 export PYTHONPATH=${PYTHONPATH}:../python
 SERVER=$1
-
+PUBKEY=$2 # PEM encoded file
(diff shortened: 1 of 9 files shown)


eranmes:
(1) Can you please squash the commits together?
(2) When you fill in the CLA I think there's somewhere you can put down your github username so it gets linked to you. Can you make sure the author / github username match in the CLA and this change? That's the only reason I can imagine why googlebot is unhappy.

pphaneuf:
There are commits here by two different GitHub usernames, I think the system only supports one username per person?

When squashing, if you use your @iangcarroll account, this should fix this problem, I think.

iangcarroll:
Ah, forgot about this. It appears my git was configured with the wrong username, thus triggering googlebot. 

I'll squash them shortly.

iangcarroll:
That branch is messed up from my bad rebase. Just going to create a new PR...

PR #363, opened 2014-12-18
Changed way cert desc stores repeated fields.

laiqu:
Fixed todo from previous pull request


laiqu:
Solved differently with protobuf

PR #371, opened 2014-12-24
EV whitelist: Gather certs from all pending EV certs

eranmes:
This includes both Certly and Izenpe's logs.


python/utilities/ev_whitelist/create_ev_whitelist_zip.py
```diff
@@ -25,6 +25,8 @@
     "https://ct.googleapis.com/pilot",
     "https://ct.googleapis.com/aviator",
     "https://ct1.digicert-ct.com/log",
+    "https://log.certly.io",
+    "ct.izenpe.com",
```
eranmes: Good point, fixed.

pphaneuf:
LGTM

PR #373, opened 2014-12-30
Changed certificate protobuf.

laiqu:
This already existed and has more information, so why not use it! I skipped few fields (and added TODOs for those) which needed more complicated changes. Also with that protobuf CertificateDescription class is no longer needed, so cert_desc module now only contains methods for easy creation of x509certificate proto.


pphaneuf:
So you more than doubled the size of ```cert_desc.py```, but also *deleted* ```cert_desc_test.py```? Could you explain this decision? :smiley:

laiqu:
I also removed the class that was tested in cert_desc_test.py, so based on previous code there is nothing to test ;)

pphaneuf:
Ok, so you removed a class, but you sure added something else! And while the output is different, it seems to me that the test for it should be fairly similar, no? Even if you didn't test all the new features that you added (which you should!), it shouldn't take a lot of tweaking to make the existing test check that the appropriate protobuf fields have been set...

python/ct/client/db/cert_desc_test.py
```diff
@@ -1,21 +0,0 @@
-#!/usr/bin/env python
-import unittest
-from ct.client.db import cert_desc
-from ct.crypto import cert
-CERT = cert.Certificate.from_der_file("ct/crypto/testdata/google_cert.der")
-class CertificateDescriptionTest(unittest.TestCase):
-    def test_from_cert(self):
-        desc = cert_desc.CertificateDescription.from_cert(CERT)
-        self.assertEqual(desc.der, CERT.to_der())
```
pphaneuf: Like, this line could become something like this:

    self.assertEqual(desc.der_cert, CERT.to_der())

python/Makefile
```diff
@@ -1,5 +1,5 @@
 all: ct/proto/client_pb2.py ct/proto/ct_pb2.py ct/proto/tls_options_pb2.py \
-	ct/proto/test_message_pb2.py ct/proto/certificate_pb2.py
```
pphaneuf: Nothing uses this certificate.proto anymore? How about deleting it?

laiqu: Yeah, ops :(

laiqu:
Previous test mainly existed because I wanted to catch syntax errors in CertificateDescription... I feel like most errors caught by this test would be also caught in other places (mainly cert.py), so it feels redundant. On the other hand test would show the format of data in protobuf, and also maybe I should make sure that my helper functions work properly.

python/ct/client/db/sqlite_cert_db.py
```diff
@@ -66,18 +66,8 @@ def __str__(self):
...
-        # using collections.Iterable won't work properly because unicode is also
-        # iterable
-        if isinstance(obj, list):
-            return obj
-        else:
-            return [obj]
-
     def __store_cert(self, cert, index, log_key, cursor):
-        der_cert = cert.der
+        der_cert = cert.der_cert
```
pphaneuf: This local variable is kind of silly, no?

python/ct/client/db/sqlite_cert_db.py
```diff
@@ -87,19 +77,19 @@ def __store_cert(self, cert, index, log_key, cursor):
...
+        if cert.subject:
+            for name in cert.subject.common_name:
+                cursor.execute("INSERT INTO subject_names(log, cert_id, name)"
+                               " VALUES(?, ?, ?)", (log_key, index, name))
+        # TODO(laiqu) fill alternative names when those are more available in
+        # protobuf
+
 
     def store_certs_desc(self, certs, log_key):
         """Store certificates using it's descriptions.
```
pphaneuf: Descriptions, eh?

laiqu: Well those still aren't actual certificates ;) but I guess mentioning that I want protobufs would be the best

python/ct/client/db/sqlite_cert_db.py
```diff
@@ -87,19 +77,19 @@ def __store_cert(self, cert, index, log_key, cursor):
                                     ('?,' * len(self.cert_table_columns))[:-1]),
```
pphaneuf: This looks nice and flexible, but I think that while it made sense with the original more dynamic approach (```[cert[column] for column in self.cert_table_columns]```), now that you do ```str(cert.version + 1), cert.serial_number``` directly (which I kind of like better, really), it doesn't make much sense. You've got an explicit mapping, and now, changing code far away from this method will break it, which is annoying.

laiqu:
Changed to even different protobuf.

PR #375, opened 2014-12-30
Print observations as they come.

laiqu:
With this change text reporter will report observations as they come, so it will greatly reduce memory consumption.

### DIFF
--- a/python/ct/client/text_reporter.py
+++ b/python/ct/client/text_reporter.py
@@ -18,53 +18,41 @@ class TextCertificateReport(reporter.CertificateReport):
         """Report stored changes and reset report."""
         super(TextCertificateReport, self).report()
         logging.info("Report:")
-        entries_with_issues = 0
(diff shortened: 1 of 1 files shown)


python/ct/client/text_reporter.py
```diff
@@ -18,53 +18,43 @@ def report(self):
...
+        self.entries_with_issues = 0
+        self.new_entries_count = 0
 
     def _batch_scanned_callback(self, result):
-        for desc, log_index, observations in result:
-            self._observations_by_index[log_index] += observations
+        for _, log_index, cert_observations in result:
+            msg = "Cert %d:" % log_index
+            observations = []
+            for obs in cert_observations:
```
eranmes: How about:
observations = [str(obs) for obs in cert_observations]

eranmes:
LGTM, see one minor comment, you'll have to rebase and deal with merge conflicts though.

PR #377, opened 2015-01-05
Run the etcd watch callback runs on a specific executor.

pphaneuf:
As usual, testing is everything. :stuck_out_tongue:


AlCutter:
lgtm

PR #381, opened 2015-01-07
Support log private keys held in an OpenSSL ENGINE

robstradling:
Patches for issue #380.

### DIFF
--- a/cpp/util/read_private_key.cc
+++ b/cpp/util/read_private_key.cc
@@ -1,6 +1,8 @@
 #include "util/read_private_key.h"
 
+#include <openssl/engine.h>
 #include <openssl/pem.h>
+#include <stdlib.h>
 
 namespace cert_trans {
 namespace util {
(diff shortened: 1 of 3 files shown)


pphaneuf:
Could you split out the first two commits into their own pull request?

cpp/util/read_private_key.cc
```diff
@@ -24,5 +27,68 @@ KeyError ReadPrivateKey(EVP_PKEY** pkey, const std::string& file) {
...
+
+static void EngineFinish() {
+  ENGINE_finish(engine);
+}
+
+KeyError LoadEnginePrivateKey(EVP_PKEY** pkey, const std::string& engine_name,
+                              const std::string& engine_pre_cmds,
+                              const std::string& engine_post_cmds,
+                              const std::string& key_id) {
+  ENGINE_load_builtin_engines();
```
pphaneuf: Is this safe to call multiple times?

cpp/util/read_private_key.cc
```diff
@@ -24,5 +27,68 @@ KeyError ReadPrivateKey(EVP_PKEY** pkey, const std::string& file) {
 }
 
 
+static ENGINE* engine = NULL;
```
pphaneuf: Pretty much everything that touches this is crying out for a wrapper class... All those ```atexit```, eek!

cpp/util/read_private_key.cc
```diff
@@ -24,5 +27,68 @@ KeyError ReadPrivateKey(EVP_PKEY** pkey, const std::string& file) {
...
+                              const std::string& key_id) {
+  ENGINE_load_builtin_engines();
+  atexit(ENGINE_cleanup);
+
+  engine = ENGINE_by_id(engine_name.c_str());
+  if (!engine)
+    return NO_SUCH_ENGINE;
+  atexit(EngineFree);
+
+  // Temporary copy of the pre_cmds string that will be modified.
```
pphaneuf: This string manipulation is leaving me a bit unimpressed (I'd prefer to remain within the realm of ```std::string```, with things like ```find``` and ```substr```), but I'd have to read up on ```ENGINE_ctrl_cmd_string``` to know what is can or cannot do to its parameters first, maybe what I'm imagining wouldn't make sense...

iangcarroll:
Is this a dead PR? OpenSSL engine support would be great.

robstradling:
It's not dead.  Just resting.  ;-)

The code in this PR predates super_duper, so it's geared towards there being only one signing node (and hence only 1 HSM would be required).  IIRC (I can't find the thread), @pphaneuf was going to have a think about possible changes to the super_duper architecture so that there isn't a requirement to have multiple HSMs.

PR #382, opened 2015-01-08
Set default encoding to utf-8.

laiqu:
Since in every log we will have to deal with utf-8 characters it makes sense to set it to default. Also without setting it manually it's set sometimes to None, which causes exceptions while printing observations (we get UNPRINTABLE, when printing unicode strings which is horrible). I honestly couldn't find better way, but this change makes sense nonetheless.


laiqu:
Fixed by other pull request.

PR #383, opened 2015-01-08
Observations now can't raise while printing.

laiqu:
We shouldn't break whole text_reporter report because some observation is unprintable (for any reason), but we definitely should note that, so I changed default format details.

### DIFF
--- a/python/ct/cert_analysis/observation.py
+++ b/python/ct/cert_analysis/observation.py
@@ -1,3 +1,5 @@
+import logging
+
 class Observation(object):
     """Describes certificate observation."""
     def __init__(self, description, reason=None, details=None):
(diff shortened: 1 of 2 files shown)


python/ct/cert_analysis/observation.py
```diff
@@ -11,20 +13,27 @@ def __repr__(self):
...
         if self.details:
-            ret = "%s [%s]" % (ret, self._format_details())
+            ret = u"%s [%s]" % (ret, self._format_details())
         return ret
 
+    def __str__(self):
+        return unicode(self)
+
     def _format_details(self):
         """Conveience method, so it's easy to override how details have to be
```
pphaneuf: "Convenience", while you're there... :wink:

laiqu: I hate this word with all my heart.

pphaneuf:
I'm going to have to read more Python documentation to figure out what the heck is going on here. :stuck_out_tongue:

pphaneuf:
After much faffing about, LGTM.

PR #385, opened 2015-01-09
Query TempDB in more async way.

laiqu:
If database is slow for some reason, we don't want it to block whole Twisted.

### DIFF
--- a/python/ct/client/async_log_client_test.py
+++ b/python/ct/client/async_log_client_test.py
@@ -10,6 +10,7 @@ from ct.client import log_client_test_util as test_util
 from ct.client.db import database
 from twisted.internet import defer
 from twisted.internet import task
+from twisted.internet import reactor
 from twisted.python import failure
 from twisted.test import proto_helpers
 from twisted.trial import unittest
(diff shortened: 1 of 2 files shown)


python/ct/client/async_log_client_test.py
```diff
@@ -113,7 +114,7 @@ def request(self, method, uri):
 
     class FakeDB(object):
         def scan_entries(self, first, last):
-            raise database.KeyError
+            raise database.KeyError("boom!")
```
pphaneuf: Either leave the empty message, or put a better one.

python/ct/client/async_log_client_test.py
```diff
@@ -131,13 +132,15 @@ def one_shot_client(self, json_content):
                                          test_util.DEFAULT_URI,
                                          reactor=self.clock)
 
-    def default_client(self, entries_db=None):
+    def default_client(self, entries_db=None, reactor_ = None):
```
pphaneuf: Extra whitespace here.

python/ct/client/async_log_client_test.py
laiqu: I changed it, so we raise actual instance of Exception. Also it's nice to have boom! so we know that it's test exception (the same pattern is in monitor_test)

pphaneuf:
LGTM

PR #387, opened 2015-01-12
Call only one watch callback at a time, in order.

pphaneuf:
This fixes a race condition (that I believe only happens in testing at the moment, but I'm fixing it generally).


pphaneuf:
I updated this pull request, and added some passing around of the ```unique_lock```, so we can assert that it is owned (this just makes me feel better).

pphaneuf:
Here is an example of the race condition I'm trying to fix:

https://travis-ci.org/google/certificate-transparency/builds/46762140

AlCutter:
LGTM

PR #388, opened 2015-01-12
Python: Adding add-chain implementation.

eranmes:


python/ct/crypto/verify.py
```diff
@@ -11,6 +11,44 @@
...
 import ecdsa
 
+def decode_signature(signature):
+    """Decode the TLS-encoded serialized signature.
+
+    Args:
+        signature: TLS-encoded signature.
+
+    Returns:
+        the inner, ASN.1 encoded contents.
```
pphaneuf: Is this a stray comma?

eranmes: I think not - that's the original method description - but I'll be happy to change.
Also, it's wrong :) so I'll fix

pphaneuf:
It'd be nice to test an error case, but the code itself is LGTM.

eranmes:
Added tests for error cases.

PR #390, opened 2015-01-12
Increased sqlite timeout.

laiqu:
With previous change (sending reads/writes to db to different threads), we need higer timeout, because database could be easily locked for more than 5 seconds.


pphaneuf:
LGTM

PR #391, opened 2015-01-14
Add HTTP request logging to server stdout

phad:
This is useful for debugging what a client is sending to the server.
I've approximately borrowed the Apache request log format.
example:
127.0.0.1 - - [2015-Jan-14 11:53:12] "GET /ct/v1/get-sth" 200 492
127.0.0.1 - - [2015-Jan-14 11:53:26] "GET /ct/v1/get-roots" 200 1565
127.0.0.1 - - [2015-Jan-14 11:53:32] "GET /ct/v1/get-entries" 400 80
127.0.0.1 - - [2015-Jan-14 11:53:44] "GET /ct/v1/get-entries?start=1&end=1" 200 3923

### DIFF
--- a/cpp/server/handler.cc
+++ b/cpp/server/handler.cc
@@ -49,16 +49,52 @@ namespace {
 const char kJsonContentType[] = "application/json; charset=ISO-8859-1";
 
 
+void LogRequest(evhttp_request* req, int http_status, int resp_body_length) {
+  evhttp_connection* conn = evhttp_request_get_connection(req);
+  char* peer_addr;
+  ev_uint16_t peer_port;
+  evhttp_connection_get_peer(conn, &peer_addr, &peer_port);
+
+  string http_verb;
+  switch (evhttp_request_get_command(req)) {
+    case EVHTTP_REQ_DELETE:
+      http_verb = "DELETE";
+      break;
+    case EVHTTP_REQ_GET:
+      http_verb = "GET";
+      break;
+    case EVHTTP_REQ_HEAD:
+      http_verb = "HEAD";
+      break;
+    case EVHTTP_REQ_POST:
+      http_verb = "POST";
+      break;
+    case EVHTTP_REQ_PUT:
+      http_verb = "PUT";
+      break;
+    default:
+      http_verb = "UNKNOWN";
+      break;
+  }
+
+  const string uri(evhttp_request_get_uri(req));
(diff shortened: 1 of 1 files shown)


phad:
I've added a new dep on boost's date_time library.  Shout if you'd sooner I didn't do that and use <ctime> or similar.

pphaneuf:
I haven't looked closely yet, but indeed, I'd rather not add a Boost dependency, as we weaned ourselves away from Boost in ```super_duper```.

Also, not a big fan of logging to stdout, wouldn't just some regular glog logging do?

pphaneuf:
Looks a bit at what you were using, and I think ```std::chrono::system_clock``` should have everything you need?

http://en.cppreference.com/w/cpp/chrono

But I'm still thinking that while this is kind of handy, I'm thinking a ```VLOG(1)``` (or *maybe* ```LOG(INFO)```) might be sufficient?

phad:
Thanks, I'll look at std::chrono::system_clock - I'll be happy to not add new Boost deps.
No problem about logging via glog.

pphaneuf:
And if you log with glog, it takes care of the timestamp for you, so that would settle the Boost question for good, I think. :smiley:

phad:
That's true, but it's appealing to have a standard httpd-style log format for this info, don't you think?

cpp/server/handler.cc
```diff
@@ -49,16 +49,58 @@ namespace {
...
+      break;
+  }
+
+  const string uri(evhttp_request_get_uri(req));
+
+  // Omitted fields 2, 3, 4 are:
+  //  2: identity of the client determined by identd on the clients machine
+  //  3: userid of the person requesting the document as determined by HTTP authentication
+  //  4: the time that the request was received, since Glog timestamps logged lines anyway.
+  LOG(INFO) << string(peer_addr) << " - - - \"" << http_verb << " " << uri << "\" "
```
pphaneuf: Can probably do without the pretense of common log format (with the " - - - " bit), if we're doing ```LOG(INFO)```? If this later changes to writing to a common log file, we can add them back.

cpp/server/handler.cc
```diff
@@ -49,16 +49,58 @@ namespace {
 const char kJsonContentType[] = "application/json; charset=ISO-8859-1";
 
 
+// Log an incoming HTTP request in the Apache access log format.
```
pphaneuf: That's not true anymore.

pphaneuf:
LGTM, feel free to push once comments are addressed.

PR #392, opened 2015-01-14
StateKeeper now can be changed.

laiqu:
It's possible that state could be stored somewhere else than disc, so we need a way to supply it to the prober. In this way the main script (logobserver.py) can be in charge of everything.

### DIFF
--- a/python/ct/client/prober.py
+++ b/python/ct/client/prober.py
@@ -19,7 +19,7 @@ from twisted.web import client as twisted_client
 class ProberThread(threading.Thread):
     """A prober for scheduled updating of the log view."""
     def __init__(self, ct_logs, db, cert_db, temp_db_factory, monitor_state_dir,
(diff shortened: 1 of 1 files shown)


python/ct/client/prober.py
```diff
@@ -19,7 +19,7 @@
 class ProberThread(threading.Thread):
     """A prober for scheduled updating of the log view."""
     def __init__(self, ct_logs, db, cert_db, temp_db_factory, monitor_state_dir,
-                 agent=None):
+                 agent=None, state_keeper_type=None):
```
eranmes: nit: state_keeper_class?

eranmes:
LGTM modulu parameter name change.

laiqu:
Merged as 1fc2dbe1efc6bed0659cf22af6379d8c5ec45784.

PR #393, opened 2015-01-14
Python precerts scanner: Allow specifying start index.

eranmes:
This allows finding "recent" precerts.


pphaneuf:
That's the kind of pull request I like. :smiley: 

LGTM

PR #394, opened 2015-01-14
State now saves amount of parsed entries.

laiqu:
Piggied back crash if entries won't verify.

Since logs grew to 6 millions and are still growing, it would be good to not start from 0 if something goes wrong while fetching. 

PS: I hate Twisted.


pphaneuf:
LGTM

PR #399, opened 2015-01-15
Bug-fix in server where rejecting incorrect HTTP methods on some APIs continued to handle the request.

phad:
The bug was probably harmless since these were all GET requests; nonetheless, if the caller invoked an API using the wrong HTTP verb, eg. DELETE /ct/v1/get-sth, handling of the request should be terminated immediately after the SendError() call.


eranmes:
LGTM - although the merge commit should be squashed (there'll be a different one when you merge via the web interface).

pphaneuf:
LGTM, but you're in a bit of a situation... Stand by, I'll be sending you some info on sorting this out in a second.

PR #400, opened 2015-01-15
Add ClusterStateController to determine current serving STH.

AlCutter:


cpp/util/masterelection.cc
```diff
@@ -91,8 +91,14 @@ MasterElection::MasterElection(const shared_ptr<libevent::Base>& base,
...
 
+// Testing only c'tor
+MasterElection::MasterElection()
+    : client_(nullptr), proposal_state_(ProposalState::NONE) {
+}
+
+
 MasterElection::~MasterElection() {
-  VLOG(1) << my_proposal_path_ << ": Joining election";
+  VLOG(1) << my_proposal_path_ << ": Destroying election";
```
pphaneuf: I looked at that pull request SO MUCH, and that's where it got us. :stuck_out_tongue:

cpp/log/cluster_state_controller_test.cc
```diff
@@ -0,0 +1,290 @@
...
+  EXPECT_EQ(sth200_.tree_size(), sth.tree_size());
+}
+
+
+}  // namespace cert_trans
+
+
+int main(int argc, char** argv) {
+  cert_trans::test::InitTesting(argv[0], &argc, &argv, true);
+  ::testing::FLAGS_gtest_death_test_style = "threadsafe";
```
pphaneuf: What's that about?

.gitignore
```diff
@@ -9,7 +9,7 @@
 *.so
 *_pb2.py
 *~
-.*.swp
+.*.sw[lnmop]
```
pphaneuf: Oh vim, where will the madness end?!?

cpp/log/consistent_store.h
```diff
@@ -128,6 +128,8 @@ class ConsistentStore {
 
   void WatchClusterNodeStates(const ClusterNodeStateCallback& cb);
 
+  virtual std::string NodeId() const = 0;
```
pphaneuf: That's nice and all, but is anyone going to use it soon? If so, I could tolerate it here (but I'd have preferred it to be added with its user), and if not, YAGNI. :wink:

cpp/util/masterelection.cc
AlCutter: hehe :)

cpp/log/cluster_state_controller_test.cc
AlCutter: Whoops, that was dragged in from elsewhere by mistake - dinged.

cpp/log/consistent_store.h
AlCutter: I'd put that in for an earlier iteration, don't need it now (yet) - removed.

cpp/log/cluster_state_controller.h
```diff
@@ -0,0 +1,92 @@
...
+  // current state of the nodes.
+  // Returns true iff there currently is a suitable STH.
+  bool CalculateServingSTH(const std::unique_lock<std::mutex>& lock);
+
+  // Thread entry point for ServingSTH updater thread.
+  void ClusterServingSTHUpdater();
+
+  util::Executor* executor_;        // Not owned by us
+  ConsistentStore<Logged>* store_;  // Not owned by us
+  const MasterElection& election_;
```
pphaneuf: This is very evil, and Not Allowed. Const references can be temporaries, so if you are passed in a const ref, you can't keep it, as it might disappear as soon as the constructor returns. Unfortunately, the safer thing to do is to use const pointers...

http://pphaneuf.livejournal.com/138678.html

pphaneuf: Also, I'd like more ```const```, overall, in lines 69-71, which, yes, will end you up with this:

    const MasterElection* const election_;

:wink:

cpp/log/cluster_state_controller.h
```diff
@@ -0,0 +1,92 @@
...
+//
+// In particular, this class calculates the optimal STH for the cluster to
+// serve at any given time.
+template <class Logged>
+class ClusterStateController {
+ public:
+  ClusterStateController(util::Executor* executor,
+                         ConsistentStore<Logged>* store,
+                         const MasterElection& election,
+                         const int min_serving_nodes,
```
pphaneuf: Shouldn't it be pulling the min serving nodes (and fraction) out of the ```ConsistentStore```?

pphaneuf: As in, it'd be in a configuration key somewhere, so that everyone is in agreement on what these should be?

cpp/log/cluster_state_controller.h
AlCutter: Yep, thinko - been there and been bitten by this before too, thanks for catching it.
BTW, someone's hacked your blog and stuck a photo of some weird looking bloke next to your name on that link.

cpp/log/cluster_state_controller.h
AlCutter: Ah  - when I said yesterday that this is how cluster-wide config would work, I didn't mean that code to do it currently existed :)

As it happens ClusterConfig is the topic of the [probably] next PR which is in-flight on another branch...

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+
+
+template <class Logged>
+ClusterStateController<Logged>::ClusterStateController(
+    util::Executor* executor, ConsistentStore<Logged>* store,
+    const MasterElection* election, const int min_serving_nodes,
+    const double min_serving_fraction)
+    : executor_(CHECK_NOTNULL(executor)),
+      store_(CHECK_NOTNULL(store)),
+      election_(election),
```
pphaneuf: Could use a ```CHECK_NOTNULL``` here now.

AlCutter: Added.

cpp/log/cluster_state_controller.h
pphaneuf: Ok, having them as constructor parameters (and keeping them in const member variables) for now, and improving on this later, that makes sense.

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
+#ifndef CERT_TRANS_LOG_CLUSTER_STATE_CONTROLLER_INL_H_
+#define CERT_TRANS_LOG_CLUSTER_STATE_CONTROLLER_INL_H_
+
+#include "log/cluster_state_controller.h"
+#include "log/etcd_consistent_store-inl.h"
```
pphaneuf: The ```-inl``` one? I'd also put a blank line between the ```cluster_state_controller.h``` one and the rest, it happens to be at the top, but a ```-inl.h``` is sort of morally equivalent to a ```.cc``` file, I'd say...

cpp/log/cluster_state_controller.h
```diff
@@ -0,0 +1,92 @@
...
+
+  // Calculates the STH which should be served by the cluster, given the
+  // current state of the nodes.
+  // Returns true iff there currently is a suitable STH.
+  bool CalculateServingSTH(const std::unique_lock<std::mutex>& lock);
+
+  // Thread entry point for ServingSTH updater thread.
+  void ClusterServingSTHUpdater();
+
+  util::Executor* const executor_;        // Not owned by us
```
pphaneuf: Not used, as far as I can tell?

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+    update_required_cv_.notify_all();
+  }
+  cluster_serving_sth_update_thread_.join();
+}
+
+
+template <class Logged>
+void ClusterStateController<Logged>::NewTreeHead(
+    const ct::SignedTreeHead& sth) {
+  CHECK(sth.IsInitialized());
```
pphaneuf: This doesn't really check anything, as all the fields in that protobuf are optional.

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+template <class Logged>
+ClusterStateController<Logged>::ClusterStateController(
+    util::Executor* executor, ConsistentStore<Logged>* store,
+    const MasterElection* election, const int min_serving_nodes,
+    const double min_serving_fraction)
+    : executor_(CHECK_NOTNULL(executor)),
+      store_(CHECK_NOTNULL(store)),
+      election_(CHECK_NOTNULL(election)),
+      min_serving_nodes_(min_serving_nodes),
+      min_serving_fraction_(min_serving_fraction),
```
pphaneuf: Maybe add a few sanity CHECKs for ```min_serving_nodes_``` and ```min_serving_fraction_``` in the constructor below? At least 1 node, and, uh... greater than zero, maybe? Greater than or equal to zero?

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+
+  // Next calculate the newest STH we've seen which satisfies the following
+  // criteria:
+  //   - at least min_serving_nodes_ have an STH at least as large
+  //   - at least min_serving_fraction_ have an STH at least as large
+  //   - not smaller than the current serving STH
+  int num_nodes_seen(0);
+  const int current_tree_size(calculated_serving_sth_.get() != nullptr
```
pphaneuf: You can use the standard smart pointers in boolean expressions, because, well, they're properly behaved. :wink: 

    const int current_tree_size(calculated_serving_sth_ ? calculated_serving_sth_->tree_size() : 0);

cpp/log/cluster_state_controller.h
```diff
@@ -0,0 +1,92 @@
...
+  // Gets the current (if any) calculated serving STH for the cluster.
+  // If there is such an STH then return true and |sth| is populated, returns
+  // false otherwise.
+  //
+  // Note that this simply returns this node's intepretation of the optimum
+  // serving STH, the current master/contents of the servingSTH file may
+  // differ.
+  //
+  // Really only intended for testing.
+  bool GetCalculatedServingSTH(ct::SignedTreeHead* sth) const;
```
pphaneuf: This would be a good candidate for ```StatusOr```, I think.

cpp/log/cluster_state_controller.h
```diff
@@ -0,0 +1,92 @@
...
+
+  // Updates *this* node's ClusterNodeState to reflect the new tree data
+  // available.
+  void ContiguousTreeSizeUpdated(const int64_t new_contiguous_tree_size);
+
+  // Gets the current (if any) calculated serving STH for the cluster.
+  // If there is such an STH then return true and |sth| is populated, returns
+  // false otherwise.
+  //
+  // Note that this simply returns this node's intepretation of the optimum
```
pphaneuf: s/intepretation/interpretation/

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+  return true;
+}
+
+
+template <class Logged>
+void ClusterStateController<Logged>::PushLocalNodeState(
+    const std::unique_lock<std::mutex>& lock) {
+  CHECK(lock.owns_lock());
+  const util::Status status(store_->SetClusterNodeState(local_node_state_));
+  if (!status.ok()) {
```
pphaneuf: No worse than that? That said, I'm trying to think up of something better to do, and I've got nothing. :smiley:

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+  PushLocalNodeState(lock);
+}
+
+
+template <class Logged>
+bool ClusterStateController<Logged>::GetCalculatedServingSTH(
+    SignedTreeHead* sth) const {
+  CHECK_NOTNULL(sth);
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (calculated_serving_sth_.get() == nullptr) {
```
pphaneuf: if (!calculated_serving_sth_)

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+    std::unique_lock<std::mutex> lock(mutex_);
+    update_required_cv_.wait(lock, [this]() {
+      return update_required_ || exiting_;
+    });
+    if (exiting_) {
+      return;
+    }
+    CHECK(update_required_);
+    CHECK_NOTNULL(calculated_serving_sth_.get());
+    const ct::SignedTreeHead local_sth(*calculated_serving_sth_.get());
```
pphaneuf: Can drop the ```.get()```, I believe.

cpp/log/cluster_state_controller-inl.h
AlCutter: Jiggled around.

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+    if (update.exists_) {
+      all_node_states_[node_id] = update.handle_.Entry();
+    } else {
+      CHECK_EQ(1, all_node_states_.erase(node_id));
+    }
+  }
+
+  if (CalculateServingSTH(lock) && election_->IsMaster()) {
+    update_required_ = true;
+    update_required_cv_.notify_all();
```
pphaneuf: Give up the lock before the ```notify_all()```?

cpp/log/cluster_state_controller.h
AlCutter: eagle_eye_actionman
Eagle eye Pierre!

cpp/log/cluster_state_controller-inl.h
AlCutter: That's true, if any get added then it'll magically start working though - got any opinions on whether to leave it or ding it (I'm fairly easy either way)?

cpp/log/cluster_state_controller-inl.h
AlCutter: Added.

cpp/log/cluster_state_controller-inl.h
AlCutter: >_> done

cpp/log/cluster_state_controller.h
AlCutter: Hmm, that's of marginal value in this particular case I think because there's never going to be a 3rd state, but since it's Friday did it anyway.

cpp/log/cluster_state_controller.h
AlCutter: fxied

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+  // First, create a mapping of tree size to number of nodes at that size, and
+  // a mapping of the newst STH for any given size:
+  std::map<uint64_t, ct::SignedTreeHead> sth_by_size;
+  std::map<uint64_t, int> num_nodes_by_sth_size;
+  for (const auto& node : all_node_states_) {
+    if (node.second.has_newest_sth()) {
+      const uint64_t tree_size(node.second.newest_sth().tree_size());
+      num_nodes_by_sth_size[tree_size]++;
+      if (node.second.newest_sth().timestamp() >
+          sth_by_size[tree_size].timestamp()) {
```
pphaneuf: You're relying on getting an empty ```SignedTreeHead``` if that tree size didn't exist, and so hitting the default ```.timestamp()``` that gives zero, right? Maybe add a short comment to explain?

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+
+template <class Logged>
+bool ClusterStateController<Logged>::CalculateServingSTH(
+    const std::unique_lock<std::mutex>& lock) {
+  VLOG(1) << "Calculating new ServingSTH...";
+  CHECK(lock.owns_lock());
+
+  // First, create a mapping of tree size to number of nodes at that size, and
+  // a mapping of the newst STH for any given size:
+  std::map<uint64_t, ct::SignedTreeHead> sth_by_size;
```
pphaneuf: I think the ```uint64_t``` should be ```int64_t``` instead? It always feels weird to me as well to use signed for sizes, but there are some decent arguments for it...

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+    // num_nodes_seen keeps track of the number of nodes we've seen so far (and
+    // since we're working from larger to smaller size STH, they should all be
+    // able to serve this [and smaller] STHs.)
+    num_nodes_seen += it->second;
+    const double serving_fraction(static_cast<double>(num_nodes_seen) /
+                                  all_node_states_.size());
+    if (serving_fraction >= min_serving_fraction_ &&
+        num_nodes_seen >= min_serving_nodes_) {
+      LOG(INFO) << "Can serve @" << it->first << " with " << num_nodes_seen
+                << " nodes (" << (serving_fraction * 100) << "% of cluster)";
```
pphaneuf: I used to prefer percentages, but I've been at Google long enough that now I prefer 0..1 intervals. :smiley:

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+template <class Logged>
+void ClusterStateController<Logged>::OnClusterStateUpdated(
+    const std::vector<Update<ct::ClusterNodeState>>& updates) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  for (const auto& update : updates) {
+    const std::string& node_id(update.handle_.Entry().node_id());
+    if (update.exists_) {
+      all_node_states_[node_id] = update.handle_.Entry();
+    } else {
+      CHECK_EQ(1, all_node_states_.erase(node_id));
```
pphaneuf: So if we lose nodes to the point where nobody has enough entries to serve the current STH, we just won't update the serving STH (```CalculateServingSTH()``` below will return false), and indirectly stop serving (as all the nodes left would be lame ducked), I suppose?

cpp/log/cluster_state_controller-inl.h
AlCutter: Yeah, I suspect we'll probably have something to wire up to this later on...

cpp/log/cluster_state_controller-inl.h
AlCutter: done

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
+#ifndef CERT_TRANS_LOG_CLUSTER_STATE_CONTROLLER_INL_H_
+#define CERT_TRANS_LOG_CLUSTER_STATE_CONTROLLER_INL_H_
+
+#include <functional>
+
+#include "log/cluster_state_controller.h"
```
pphaneuf: Still sort of wrong, I'd say... This header should be the very first one, to ensure that it's not "fixed" by ```#include <functional>```.

cpp/log/cluster_state_controller-inl.h
AlCutter: Yeah, typo/thinko

cpp/log/cluster_state_controller-inl.h
AlCutter: Good call, done.

cpp/log/cluster_state_controller-inl.h
pphaneuf: Well, adding ```required``` fields being forbidden... :smiley:

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,203 @@
...
+#include "proto/ct.pb.h"
+
```
pphaneuf: No need for an empty line here, though. The order is "our header", the system headers, and then the project headers, I figure the protobuf generated headers still count as "project headers".

cpp/log/cluster_state_controller-inl.h
AlCutter: Yep, exactly.
Added a comment.

cpp/log/cluster_state_controller-inl.h
AlCutter: Yes, you're right - changed.

cpp/log/cluster_state_controller-inl.h
AlCutter: Heh, not only us reading these logs though :)

cpp/log/cluster_state_controller-inl.h
AlCutter: Yup, that's the plan, of course it's easy to say that as the machinery to do that doesn't exist yet.

cpp/log/cluster_state_controller_test.cc
```diff
@@ -0,0 +1,289 @@
...
+
+TEST_F(ClusterStateControllerTest, TestCalculateServingSTHAt50Percent) {
+  MockMasterElection election_is_master;
+  EXPECT_CALL(election_is_master, IsMaster()).WillRepeatedly(Return(true));
+  ClusterStateController<LoggedCertificate> controller50(store_.get(),
+                                                         &election_is_master,
+                                                         1, 0.5);
+  store_->SetClusterNodeState(cns100_);
+  sleep(1);
+  util::StatusOr<SignedTreeHead> status(
```
pphaneuf: Not a big deal, can leave it like this, but I usually put the emphasis on the "Or<Thing>" in "StatusOr<Thing>" (this is mainly a ```SignedTreeHead```, once you've got the status check out of the way).

cpp/log/cluster_state_controller-inl.h
```diff
@@ -0,0 +1,207 @@
...
+                std::placeholders::_1));
+}
+
+
+template <class Logged>
+ClusterStateController<Logged>::~ClusterStateController() {
+  {
+    std::unique_lock<std::mutex> lock(mutex_);
+    exiting_ = true;
+    update_required_cv_.notify_all();
```
pphaneuf: Argh, I missed the other ```notify_all()```, for the "release the lock" thing... Not incorrect, strictly speaking, just nicer...

cpp/log/cluster_state_controller-inl.h
pphaneuf: All right, sure... :smiley:

cpp/log/cluster_state_controller-inl.h
AlCutter: unlock'd

cpp/log/cluster_state_controller_test.cc
AlCutter: changed it over

pphaneuf:
LGTM

PR #408, opened 2015-01-16
Add a UUID4 node id, and persist it into the DB.

AlCutter:


cpp/util/uuid.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <iomanip>
+#include <random>
+#include <sstream>
+#include <string>
+
+namespace cert_trans {
+
+// Generates a type 4 (128bit random) UUID:
+std::string UUID4() {
```
pphaneuf: Can we afford a .cc file? :smiley:

cpp/server/ct-server.cc
```diff
@@ -186,7 +188,15 @@ int main(int argc, char* argv[]) {
...
     db = new FileDB<LoggedCertificate>(
         new FileStorage(FLAGS_cert_dir, FLAGS_cert_storage_depth),
-        new FileStorage(FLAGS_tree_dir, FLAGS_tree_storage_depth));
+        new FileStorage(FLAGS_tree_dir, FLAGS_tree_storage_depth),
+        new FileStorage(FLAGS_meta_dir, 0));
+
+  std::string node_id;
+  if (db->NodeId(&node_id) != Database<LoggedCertificate>::LOOKUP_OK) {
+    node_id = cert_trans::UUID4();
+    LOG(INFO) << "Initializing DB for node_id: " << node_id;
```
pphaneuf: "with UUID"? Since "node_id" isn't visible outside (as far as I could tell)...

cpp/server/ct-server.cc
```diff
@@ -34,6 +35,7 @@ DEFINE_string(trusted_cert_file, "",
               "File for trusted CA certificates, in concatenated PEM format");
 DEFINE_string(cert_dir, "", "Storage directory for certificates");
 DEFINE_string(tree_dir, "", "Storage directory for trees");
+DEFINE_string(meta_dir, "", "Storage directory for meta info");
```
pphaneuf: Maybe not in this pull request, but I've always found it a bit odd to have to specify each directory separately (instead of just having one, and ```ct-server``` figuring out how it wants to use it), and this is making it even more clear...

cpp/log/sqlite_db.h
```diff
@@ -2,6 +2,7 @@
 #ifndef SQLITE_DB_H
 #define SQLITE_DB_H
 
+#include <memory>
```
pphaneuf: What's that header for? Maybe I'm missing something, I'm not seeing it...

cpp/log/file_db-inl.h
```diff
@@ -189,6 +192,39 @@ void FileDB<Logged>::RemoveNotifySTHCallback(
...
+    std::string* node_id) {
+  std::unique_lock<std::mutex> lock(lock_);
+  return NodeId(lock, CHECK_NOTNULL(node_id));
+}
+
+
+template <class Logged>
+typename Database<Logged>::LookupResult FileDB<Logged>::NodeId(
+    const std::unique_lock<std::mutex>& lock, std::string* node_id) {
+  CHECK(lock.owns_lock());
```
pphaneuf: Why do you need that lock? That lock protects the in-memory state of entries (the vector and map), ```FileStorage``` is thread-safe and does not need locking.

pphaneuf:
I've have some thought that maybe it'd be a good idea to have a protobuf for that metadata, but should be okay for now...

Done with comments for now.

cpp/util/uuid.h
AlCutter: What, do you think I'm made of inodes?

cpp/server/ct-server.cc
AlCutter: done

cpp/server/ct-server.cc
AlCutter: Yeah, it is a bit weird, I'll pop a TODO in there.

cpp/log/file_db-inl.h
AlCutter: Removed in that case.

cpp/log/sqlite_db.h
AlCutter: I was previously using a unique_ptr for something, but then removed it - dinged.

pphaneuf:
Assuming the little name fixup, LGTM.

PR #409, opened 2015-01-16
Store issuer from root certificate.

laiqu:
It's easy to store whole certificate chain, but because parsing them is slightly exhausting for now lets just pull out the root issuer which is most interesting bit.


pphaneuf:
LGTM

PR #412, opened 2015-01-19
Python: Fix ordering of SCT fields in proto

eranmes:
The incorrect order led to the TLS encoder creating SCTs that were
incorrectly encoded (the extensions come after the timestamp).

### DIFF
--- a/python/ct/client/tls_message.py
+++ b/python/ct/client/tls_message.py
@@ -127,7 +127,9 @@ class TLSReader(object):
         """Read from the buffer into the protocol buffer message."""
         # TODO(ekasper): probably better not to modify the
         # original message until we're guaranteed to succeed?
-        for field in message.DESCRIPTOR.fields:
+        fields = message.DESCRIPTOR.fields_by_number
+        for i in sorted(fields):
+            field = fields[i]
             opts = field.GetOptions().Extensions[options.tls_opts]
             if opts.skip:
                 continue
@@ -257,7 +259,9 @@ class TLSWriter(object):
 
     def write(self, message):
         """Append a serialized message to the writer's buffer."""
-        for field in message.DESCRIPTOR.fields:
+        fields = message.DESCRIPTOR.fields_by_number
+        for i in sorted(fields):
+            field = fields[i]
             opts = field.GetOptions().Extensions[options.tls_opts]
             if opts.skip:
                 continue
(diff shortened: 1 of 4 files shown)


ekasper:
Hm, tls_message.py has this line (in both reader and writer):

for field in message.DESCRIPTOR.fields:
   # do something

It's not clear to me that message.DESCRIPTOR.fields' contract guarantees to return the fields in the order they were specified:
https://developers.google.com/protocol-buffers/docs/reference/python/index

But I think you can change the line above to

for field in message.DESCRIPTOR.fields_by_number:

which would read/write the fields in the numeric order. (I'm pretty sure this is what I intended to do...)

If you go with that, also clarify the ordering in tls_options.proto.

eranmes:
Did as you suggested (with a slight tweak as fields_by_number is a dict of field index -> FieldDescription).
Re clarification of ordering in tls_options.proto - was there anything else besides documentation changes?

python/ct/proto/tls_options.proto
```diff
@@ -51,6 +51,9 @@ package ct;
 //                             (tls_opts).select_value = 0];
 //   optional Orange orange = 3 [(tls_opts).select_field = "fruit_type",
 //                               (tls_opts).select_value = 1];
+//
+// Note that the numbering of the fields in the message determines the orderi
```
pphaneuf: s/orderi/order/

python/ct/client/tls_message.py
```diff
@@ -127,7 +127,9 @@ def read(self, message):
         """Read from the buffer into the protocol buffer message."""
         # TODO(ekasper): probably better not to modify the
         # original message until we're guaranteed to succeed?
-        for field in message.DESCRIPTOR.fields:
+        field_indices = message.DESCRIPTOR.fields_by_number.keys()
```
ekasper: nit

fields = message.DESCRIPTOR.fields_by_number
for i in sorted(fields):
  opts = fields[i].GetOptions().Extensions[options.tls_opts]

is a little easier to understand imho.

python/ct/proto/tls_options.proto
ekasper: s/orderi/order./ :)

ekasper:
LGTM

python/ct/proto/tls_options.proto
eranmes: Done and done.

python/ct/client/tls_message.py
eranmes: Done, although field is referred enough times that I kept it (field = fields[i]).

PR #413, opened 2015-01-19
Do not sign the tree from the libevent thread.

pphaneuf:
This is a blocking operation, and doing it from the libevent thread is Very Bad (tm).

### DIFF
--- a/cpp/server/ct-server.cc
+++ b/cpp/server/ct-server.cc
@@ -23,7 +23,6 @@
 #include "server/handler.h"
 #include "util/fake_etcd.h"
 #include "util/libevent_wrapper.h"
-#include "util/periodic_closure.h"
 #include "util/read_key.h"
 #include "util/thread_pool.h"
 #include "util/uuid.h"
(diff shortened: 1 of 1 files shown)


pphaneuf:
Wouldn't you like less blocking, @AlCutter? :wink:

cpp/server/ct-server.cc
```diff
@@ -223,19 +235,14 @@ int main(int argc, char* argv[]) {
                                             &log_signer);
   LogLookup<LoggedCertificate> log_lookup(db);
 
-  // Make sure that we have an STH, even if the tree is empty.
   // TODO(pphaneuf): We should be remaining in an "unhealthy state"
   // (either not accepting any requests, or returning some internal
   // server error) until we have an STH to serve. We can sign for now,
   // but we might not be a signer.
-  SignMerkleTree(&tree_signer, &log_lookup);
+  thread signer(&SignMerkleTree, &tree_signer);
```
AlCutter: Do we want/need a condition_variable/notification here to block until after the first run has completed?

AlCutter:
Looks lovely to me.

cpp/server/ct-server.cc
pphaneuf: No, if we don't want to deadlock again. :smiley:

PR #415, opened 2015-01-20
Refactor loading a LoggedCertificate from the client.

pphaneuf:


cpp/log/logged_certificate.cc
```diff
@@ -0,0 +1,65 @@
...
+              << entry.leaf.timestamped_entry().entry_type();
+    return false;
+  }
+
+  Clear();
+
+  ct::SignedCertificateTimestamp* const sct(mutable_contents()->mutable_sct());
+  sct->set_version(ct::V1);
+  sct->set_timestamp(entry.leaf.timestamped_entry().timestamp());
+  sct->set_extensions(entry.leaf.timestamped_entry().extensions());
```
AlCutter: This is kinda weird in that you'll end up with half an SCT.

cpp/log/logged_certificate.cc
```diff
@@ -0,0 +1,65 @@
...
+
+  mutable_contents()->mutable_entry()->set_type(
```
AlCutter: Can you just do `*mutable_contents()->mutable_entry() = entry.entry;` ?

cpp/log/logged_certificate.cc
pphaneuf: Any suggestion on where I can fish out the other half? :smiley:

cpp/log/logged_certificate.cc
pphaneuf: Well, LOOK AT THAT.

cpp/log/logged_certificate.cc
AlCutter: Don't tempt me :)

cpp/log/logged_certificate.cc
pphaneuf: Turns out you can't, but I added a comment.

AlCutter: How about doing the copy from `entry` thing and then filling in the missing bits from `leaf` rather than doing all of it manually? Might save on line count a bit.

pphaneuf:
Ok, one last update. :smiley:

pphaneuf:
Six more out!

AlCutter:
LGTM

PR #416, opened 2015-01-20
Fix spurious hang on notify.

AlCutter:


pphaneuf:
LGTM

Subtle, subtle, grumble...

PR #417, opened 2015-01-20
Fix burst STH generation.

AlCutter:
When signing runs are quick (i.e. sub-second), `target_run_time` can be slightly ahead of `now` after the `sleep` has returned (because of `duration_cast<seconds>` rounding down), this means that new STHs will be generated in a tight loop until a second has passed (this can be many STHs when the tree size/delta is small.)

### DIFF
--- a/cpp/server/ct-server.cc
+++ b/cpp/server/ct-server.cc
@@ -164,7 +164,7 @@ void SignMerkleTree(TreeSigner<LoggedCertificate>* tree_signer) {
       target_run_time += period;
     }
 
-    sleep(duration_cast<seconds>(target_run_time - now).count());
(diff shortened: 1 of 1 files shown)


cpp/server/ct-server.cc
```diff
@@ -164,7 +164,7 @@ void SignMerkleTree(TreeSigner<LoggedCertificate>* tree_signer) {
       target_run_time += period;
     }
 
-    sleep(duration_cast<seconds>(target_run_time - now).count());
+    std::this_thread::sleep_for(target_run_time - now);
```
pphaneuf: Ah, damn, I looked for that function, but couldn't remember its name, so I went for the "classic"...

pphaneuf:
LGTM

PR #418, opened 2015-01-20
Add a test for algorithm mismatch in X.509 certs

eranmes:


pphaneuf:
LGTM (once Travis greenlights it!).

eranmes:
PTAL and let me know if you're happy with it.
We now install our own openssl - it's neither pretty nor fast, but it guarantees a good, correct version  (also, no guarantees if somebody's running a log with our open-source code but an older version of openssl)

pphaneuf:
Taking off the LGTM, until I look at it more closely.

.travis.yml
```diff
@@ -20,7 +20,16 @@ before_install:
...
+ - cd openssl-1.0.1l
+ - ./config --prefix=/usr/local --openssldir=/usr/local/openssl
+ - make
+ - sudo make install
+ - cd ..
+
 
 script:
- - make alltests CC=clang CXX=clang++
+ - env OPENSSLDIR=/usr/local/openssl make alltests CC=clang CXX=clang++
```
pphaneuf: Put the ```OPENSSLDIR``` variable at the end, with the others, no need for ```env``` like this...

README.md
```diff
@@ -7,7 +7,7 @@ Auditing for TLS certificates.
 
 ## Dependencies ##
 
- - [OpenSSL](https://www.openssl.org/source/), at least 1.0.0, preferably 1.0.2 (and up)
+ - [OpenSSL](https://www.openssl.org/source/), at least 1.0.0q, preferably 1.0.1l or 1.0.2 (and up)
```
pphaneuf: 1.0.0q has the check we want? Hmm, I'll look into this, there might be a Ubuntu package available somewhere... (but don't wait for that)

pphaneuf:
Not too impressed with making the Travis build time even longer, but oh well...

LGTM

README.md
eranmes: Correct, but I've tested with 1.0.1l and there's no reason to point to an even older release.

PR #419, opened 2015-01-20
Initial code to pull entries from another log.

pphaneuf:
This is a fairly complicated algorithm, so starting with a single peer, and I'll be building on that to add more features (multiple peers, continuous following, etc).

### DIFF
--- a/cpp/fetcher/fetcher.cc
+++ b/cpp/fetcher/fetcher.cc
@@ -1,0 +1,251 @@
+#include "fetcher/fetcher.h"
+
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+#include <memory>
+#include <mutex>
+
+#include "base/macros.h"
+
+using cert_trans::AsyncLogClient;
+using cert_trans::LoggedCertificate;
+using cert_trans::Peer;
+using std::lock_guard;
+using std::move;
+using std::mutex;
+using std::placeholders::_1;
+using std::unique_ptr;
+using std::vector;
+using util::Task;
+using util::TaskHold;
+
+DEFINE_int32(fetcher_concurrent_fetches, 100, "number of concurrent fetch requests");
+DEFINE_int32(fetcher_batch_size, 1000, "maximum number of entries to fetch per request");
+
+namespace {
+
+
+struct Range {
+  enum State {
+    HAVE,
+    FETCHING,
+    WANT,
+  };
+
+  Range(State state, int64_t size, unique_ptr<Range>&& next = nullptr)
+      : state_(state), size_(size), next_(move(next)) {
+    CHECK(state_ == HAVE || state_ == FETCHING || state_ == WANT);
+    CHECK_GT(size_, 0);
+  };
+
+  mutex lock_;
+  State state_;
+  int64_t size_;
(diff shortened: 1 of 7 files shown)


cpp/fetcher/fetcher.cc
```diff
@@ -0,0 +1,254 @@
...
+using std::placeholders::_1;
+using std::unique_ptr;
+using std::vector;
+using util::Task;
+using util::TaskHold;
+
+namespace {
+
+
+const int kMaxFetches = 100;
```
AlCutter: Make into flags?

cpp/fetcher/fetcher.cc
```diff
@@ -0,0 +1,254 @@
...
+    return;
+  }
+
+  int64_t index(start_);
+  int num_fetch(0);
+  for (Range* current = entries_.get(); current;
+       index += current->size_, current = current->next_.get()) {
+    lock_guard<mutex> lock(current->lock_);
+
+    // Coaslesce the next entries, if possible.
```
AlCutter: "Coalesce with the next Range, if possible"?

cpp/fetcher/fetcher.cc
```diff
@@ -0,0 +1,254 @@
...
+                << status;
+      range->state_ = Range::WANT;
+      return;
+    }
+
+    VLOG(1) << "received " << retval->size() << " entries at offset " << index;
+    int64_t processed(0);
+    for (const auto& entry : *retval) {
+      LoggedCertificate cert;
+      if (!cert.CopyFromClientLogEntry(entry)) {
```
AlCutter: I think this probably merits at least a LOG(WARNING) or something, in all likelihood we're going to get wedged here on all subsequent runs (modulo a duff peer returning garbage etc.)

cpp/fetcher/remote_peer.h
```diff
@@ -0,0 +1,30 @@
...
+  // starts the asynchronous destruction).
+  RemotePeer(AsyncLogClient* client, LogVerifier* verifier, util::Task* task);
+  ~RemotePeer() override;
+
+  int64_t TreeSize() const override;
+
+ private:
+  struct Impl;
+
+  Impl* const impl_;
```
AlCutter: Add a comment to say that this will be deleted by the Task

pphaneuf:
I found a minor issue, when an issue is fetched twice, for some reason, I didn't check the error code for an existing entry, and just kept repeating... I'm running a full import of Aviator now to try to shake out issues, I'll address comments after that, in case anything else comes up.

cpp/fetcher/fetcher.cc
pphaneuf: This is where some of my current problems are, so I'll refine this a bit further, but I started out by adding log messages as suggested.

When this happens and we ```break;```, the logic below is what should avoid getting us wedged. On my TODO list for further pull requests is some handling for cases where we can't converge (say, a peer had a large tree size that we're trying to sync to, but it disappeared, so we can't possibly complete the operation).

pphaneuf:
Fixed the spelling mistake there, and addressed other comments.

AlCutter:
LGTM

PR #421, opened 2015-01-20
Extend LoadTrustedCertificates so certs can be loaded from in-mem buffers

phad:
This is an alternative to loading trusted certs from file.
I've added it because I've found it more convenient in some testing scenarios.

### DIFF
--- a/cpp/log/cert_checker.cc
+++ b/cpp/log/cert_checker.cc
@@ -18,6 +18,7 @@
 #include "util/util.h"
 
 using std::string;
+using std::vector;
 using util::ClearOpenSSLErrors;
 
 namespace cert_trans {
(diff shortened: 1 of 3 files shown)


cpp/log/cert_checker.h
```diff
@@ -100,6 +106,9 @@ class CertChecker {
   // deallocated appropriately.
   std::multimap<std::string, const Cert*> trusted_;
 
+  // Helper for LoadTrustedCertificates, whether reading from file or memory.
+  bool LoadTrustedCertificatesFromBIO(BIO* bio_in);
```
eranmes: Document that it takes ownership of bio_in and frees it.

cpp/log/cert_checker.cc
```diff
@@ -44,6 +45,29 @@ bool CertChecker::LoadTrustedCertificates(const std::string& cert_file) {
     return false;
   }
 
+  return LoadTrustedCertificatesFromBIO(bio_in);
+}
+
+bool CertChecker::LoadTrustedCertificates(const vector<string>& trusted_certs) {
+  string concat_certs;
+  for (vector<string>::const_iterator it = trusted_certs.begin();
```
eranmes: It's a shame we can't do lambdas - that loop would have been replaced with a single for_each line.

eranmes:
drive-by LGTM.

PR #424, opened 2015-01-21
Python: Encode SCTList.

eranmes:


python/ct/client/tls_message_test.py
```diff
@@ -232,5 +235,35 @@ def test_correctly_encodes_sct(self):
...
 
+    def test_correctly_encodes_sct_list_one_sct(self):
+        single_sct = ("0069616d617075626c69636b657973686174776f6669766573697864"
+                      "696765737400000000000004d20000040300097369676e6174757265"
+                      ).decode("hex")
+        sct_list = client_pb2.SignedCertificateTimestampList()
+        sct_list.sct_list.append(single_sct)
+        encoded_sct_list = tls_message.encode(sct_list)
+        self.assertEqual(encoded_sct_list[:4],
+                         "003a0038".decode("hex"))
```
ekasper: Add a comment explaining this mystery value?

python/ct/proto/client.proto
```diff
@@ -205,3 +205,9 @@ message SignedCertificateTimestamp {
                                   (tls_opts).max_length = 0xffff ];
   optional DigitallySigned signature = 5;
 }
+
+message SignedCertificateTimestampList {
+  repeated bytes sct_list = 1 [ (tls_opts).min_length = 1,
```
ekasper: Could test encode/decode failure on empty SCTs.

OTOH the advantage of the generic encoder is that  you shouldn't have to repeat those tests at the application layer, so up to you.

ekasper: Opts are correct, in any case.

python/ct/client/tls_message_test.py
```diff
@@ -232,5 +235,35 @@ def test_correctly_encodes_sct(self):
...
+        sct_proto_2 = client_pb2.SignedCertificateTimestamp()
+        sct_proto_2.CopyFrom(self._sct_proto)
+        sct_proto_2.timestamp = 1365427530000
+        second_sct = tls_message.encode(sct_proto_2)
+
+        sct_list = client_pb2.SignedCertificateTimestampList()
+        sct_list.sct_list.extend([first_sct, second_sct])
+        encoded_sct_list = tls_message.encode(sct_list)
+        self.assertEqual(encoded_sct_list[:4],
+                         "00f00076".decode("hex"))
```
ekasper: Ditto, explain the value.

ekasper:
LGTM

python/ct/client/tls_message_test.py
eranmes: Done.

python/ct/proto/client.proto
eranmes: As you point out, I think it's not necessary - I just want to make sure a few SCTLists are encoded similarly by other encoders.

python/ct/client/tls_message_test.py
eranmes: Done.

PR #427, opened 2015-01-21
Add MasterElection, ClusterStateController, and ClusterConfig to ct-server.

AlCutter:


cpp/server/ct-server.cc
```diff
@@ -222,10 +231,30 @@ int main(int argc, char* argv[]) {
   // For now, run with a dedicated thread pool as the executor for our
   // consistent store to avoid the possibility of DoS through thread starvation
   // via HTTP.
-  ThreadPool etcd_pool(2);
-  EtcdConsistentStore<LoggedCertificate> consistent_store(&etcd_pool,
+  ThreadPool internal_pool(4);
```
pphaneuf: Might as well leave it to the default number of threads, if it might get more generally used?

cpp/server/ct-server.cc
```diff
@@ -222,10 +231,30 @@ int main(int argc, char* argv[]) {
...
+  // Put a sensible single-node config into FakeEtcd. For a real clustered log
+  // we'd expect a ClusterConfig already to be present within etcd as part of
+  // the provisioning of the log.
+  {
+    ct::ClusterConfig config;
+    config.set_minimum_serving_nodes(1);
+    config.set_minimum_serving_fraction(1);
+    LOG(INFO) << "Setting default single-node ClusterConfig:\n"
+              << config.DebugString();
+    consistent_store.SetClusterConfig(config);
```
pphaneuf: Why set a cluster config at all? Wouldn't the default values work fine in the single-node case?

pphaneuf:
I think I'd also like a more explicit commit message. And a pony. :wink: 

If it had been just the various extra logging messages and small cleanups, I'd have been all right with "goodies", but now a wild ```MasterElection``` is making it's appearance, that seems a bit more substantial... :smiley:

cpp/server/ct-server.cc
AlCutter: Actually, I think a better idea would be to remove the defaults and just have everything be explicit.

cpp/server/ct-server.cc
AlCutter: It's pretty easy to just up this later on when the threads might actually be used, and at least this way we won't have pages of output from `info thread`.
I don't really feel particularly strongly about it though, so feel to try to convince me otherwise if you wish.

pphaneuf: I don't feel that strongly about it either, it just seems like a magic number lying around, and figured the automatic default is probably just as good as our guessing, but not having pages of output for ```info thread``` is a decent argument, IMHO.

cpp/server/ct-server.cc
pphaneuf: This kinds of buries the defaults in the code, though... Having the defaults in the protobuf makes them easy to find and read (you could add some comments explaining the settings there, which could serve as a reference), it keeps it all in one place, and makes it impossible for a code path to not have a setting (if the setting exists, it will absolutely have a default, and not ```!has_that_setting()```), even as they are added (would the defaults be explicitly set by the initial config tool, or when the protobuf is read? if the first, what when a setting is added, and it's not in the existing config?).

It also makes the config tool potentially really simple, taking an ASCII protobuf config file, say.

AlCutter: Buries defaults for the temporary-testing-only-known-to-be-a-single-node-instance where we *can* come up with a guaranteed good set of configs, vs. the risk that someone doesn't bother/realise they forgot to configure their For Realz cluster to their specific number of nodes and serving requirements because it starts and appears to work just fine...

What's a sensible default value for `minimum_num_serving_nodes` that works for clusters of 2, 5, and 50 nodes?

An ASCII proto buf fed into a config tool is more or less what I had in mind, and I'm not averse to shipping an example_config.txt which contains some values which would probably result in a functioning (if not optimal) cluster, but I want the log operator to have to consciously take the step of configuring the cluster during the the process of provisioning it.

Comments in the .proto file (and our as yet vapourware administration docs!) explaining in detail the knobs and levers is definitely a good thing to have - I'll add something there now, but I suspect that we'll later want to go over everything in there and make sure all the explanations fit together properly.

AlCutter:
Commit message updated too
mlp

cpp/server/ct-server.cc
pphaneuf: For the record, we discussed this a bit in person, and we might have to revisit later, but this is okay for now.

A big missing feature (which will be coming very soon) is for the cluster state controller not to operate without a config file (at the moment, it will run for a bit with the protobuf-supplied defaults of zero!).

pphaneuf:
Since I even got my pony, LGTM. :smiley:

PR #429, opened 2015-01-22
Make EtcdConsistentStore::NextAvailableSequenceNumber() use /serving_sth

AlCutter:
When there are no entries in `/sequenced_entries` then our next available sequence number is the one after the numbers covered by `/serving_sth`.

### DIFF
--- a/cpp/log/etcd_consistent_store-inl.h
+++ b/cpp/log/etcd_consistent_store-inl.h
@@ -72,10 +72,12 @@ EtcdConsistentStore<Logged>::NextAvailableSequenceNumber() const {
     return sequenced_entries.back().Entry().sequence_number() + 1;
   }
 
-  // TODO(alcutter): Implement the rest of the logic around /serving_sth too
-  // once there are methods to inspect that.
-  LOG(WARNING) << "NextAvailableSequenceNumber() not checking /serving_sth.";
-  return 0;
+  if (!serving_sth_) {
+    LOG(WARNING) << "Log has no Serving STH [new log?], returning 0";
+    return 0;
+  }
+
+  return serving_sth_->Entry().tree_size();
 }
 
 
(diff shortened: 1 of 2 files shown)


cpp/log/etcd_consistent_store-inl.h
```diff
@@ -72,10 +72,12 @@ EtcdConsistentStore<Logged>::NextAvailableSequenceNumber() const {
     return sequenced_entries.back().Entry().sequence_number() + 1;
   }
 
-  // TODO(alcutter): Implement the rest of the logic around /serving_sth too
-  // once there are methods to inspect that.
-  LOG(WARNING) << "NextAvailableSequenceNumber() not checking /serving_sth.";
-  return 0;
+  if (!serving_sth_) {
+    LOG(WARNING) << "Log has no Serving STH [new log?], returning 0";
```
pphaneuf: I wasn't 100% sure, but wouldn't returning -1 and having the signer not do anything when that happens be safer?

Even in the single node case, it should elect itself leader, then pick a serving STH based on its local database tree size (which would make a lot more sense than 0!)... If it did sign something using 0, while the database had sequenced entries in it, it would be Very Bad (tm).

pphaneuf:
Done with the review, just that one minor comment. :wink:

cpp/log/etcd_consistent_store-inl.h
AlCutter: Gotta boot-strap the first STH of the log somehow...

Entries in etcd's `/sequenced_entries` must only be removed once '/serving_sth' is updated (because sufficient replication), and `/serving_sth` is not allowed to be rewound down to an earlier STH, so if there are no entries in `/sequenced_entries` and there's no '/serving_sth' then (unless etcd is corrupt in which case All Bets Are Off) the right answer is that there are not yet any sequenced entries (i.e. bootstrapping the log.) so the next available sequence number is 0.

If there are sequenced entries in the local DB (and no STH) then they should also be in `/sequenced_entries` so we're looking at a corrupt etcd again if they're not.

The single node toy case should be identical to the cluster case in terms of how it works - i.e. via [fake]etcd.

pphaneuf:
LGTM

PR #433, opened 2015-01-22
Add a flag to set the SQLite "synchronous" pragma to OFF.

pphaneuf:
And some small cleanups (in a separate commit).

### DIFF
--- a/cpp/log/sqlite_db-inl.h
+++ b/cpp/log/sqlite_db-inl.h
@@ -75,7 +75,7 @@ typename Database<Logged>::WriteResult SQLiteDB<Logged>::CreateSequencedEntry_(
   sqlite::Statement statement(db_,
                               "INSERT INTO leaves(hash, entry, sequence) "
                               "VALUES(?, ?, ?)");
-  std::string hash = logged.Hash();
+  const std::string hash(logged.Hash());
   statement.BindBlob(0, hash);
 
   std::string data;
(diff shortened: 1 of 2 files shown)


cpp/log/sqlite_db-inl.h
```diff
@@ -4,12 +4,16 @@
...
 #include "log/sqlite_db.h"
 
+#include <gflags/gflags.h>
 #include <glog/logging.h>
 #include <sqlite3.h>
 
 #include "log/sqlite_statement.h"
 #include "util/util.h"
 
+DEFINE_bool(sqlite_synchronous_off, false,
```
AlCutter: Worth a TODO somewhere to capture the conversation we had just now about turning sync off while a fresh node is playing catchup (to let it catch up in less than two weeks), and then turning it on when it's in the game proper?

cpp/log/sqlite_db-inl.h
```diff
@@ -57,6 +61,12 @@ sqlite3* SQLiteOpen(const std::string& dbfile) {
 template <class Logged>
 SQLiteDB<Logged>::SQLiteDB(const std::string& dbfile)
     : db_(SQLiteOpen(dbfile)), tree_size_(0) {
+  if (FLAGS_sqlite_synchronous_off) {
+    sqlite::Statement statement(db_, "PRAGMA synchronous = OFF");
```
AlCutter: scary :8ball:

AlCutter:
LGTM

cpp/log/sqlite_db-inl.h
pphaneuf: And not the default! :smiley:

PR #434, opened 2015-01-22
Sanity checks of a request include "do we have one?".

pphaneuf:
See ```HttpRequest::Done``` for more "information".


cpp/client/async_log_client.cc
```diff
@@ -47,6 +47,11 @@ string UriEncode(const string& input) {
 // error if something is wrong.
 bool SanityCheck(const shared_ptr<libevent::HttpRequest>& req,
                  const AsyncLogClient::Callback& done) {
+  if (!req) {
```
AlCutter: <3 libevent

AlCutter:
lgtm

PR #436, opened 2015-01-22
Only grab the Range lock when it is needed.

pphaneuf:
Holding the lock of a ```Range``` instance while writing to the database can block other completed fetches from walking the list of ranges and starting new fetches.


AlCutter:
LGTM

PR #437, opened 2015-01-23
Ok to call StartElection/StopElection multiple times.

AlCutter:


cpp/util/masterelection.cc
```diff
@@ -118,21 +120,25 @@ void MasterElection::StartElection() {
...
 
 void MasterElection::StopElection() {
+  unique_lock<mutex> lock(mutex_);
+  if (!running_) {
+    return;
+  }
   VLOG(1) << my_proposal_path_ << ": Departing election.";
+  running_ = false;
 
   // Stop the updates from the watch. Do this before taking the lock,
```
pphaneuf: Change this to say "without having the lock" instead.

pphaneuf:
LGTM

PR #439, opened 2015-01-23
Only participate in the election when caught up

AlCutter:
This is the first half of this, also planning to add a ConsistentStore wrapper which only allows calls to modify cluster-wide state if the caller is currently master.

### DIFF
--- a/cpp/server/ct-server.cc
+++ b/cpp/server/ct-server.cc
@@ -67,6 +67,7 @@ DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
 namespace libevent = cert_trans::libevent;
 
 using ct::SignedTreeHead;
+using ct::ClusterNodeState;
 using cert_trans::CertChecker;
 using cert_trans::ClusterStateController;
 using cert_trans::EtcdClient;
@@ -79,6 +80,7 @@ using cert_trans::ReadPrivateKey;
 using cert_trans::MasterElection;
 using cert_trans::ThreadPool;
 using cert_trans::TreeSigner;
+using cert_trans::Update;
 using google::RegisterFlagValidator;
 using std::bind;
 using std::chrono::duration;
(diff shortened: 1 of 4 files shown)


pphaneuf:
Needs a bit of sorting out, I'll look again when I'm at the office. :smiley:

AlCutter:
Yah, remnants of the other PRs which I separated out, soz. :/

cpp/server/ct-server.cc
```diff
@@ -282,6 +289,31 @@ int main(int argc, char* argv[]) {
   ClusterStateController<LoggedCertificate> cluster_controller(
       &internal_pool, &consistent_store, &election_);
 
+  // Ensure we can only be the master if we've got sufficient local replication
+  util::SyncTask sth_watch_task(&internal_pool);
+  consistent_store.WatchServingSTH(
+      [&election_, &cluster_controller](const Update<SignedTreeHead>& sth) {
```
pphaneuf: I only noticed it now... There's a trailing underscore to ```election_```?

cpp/server/ct-server.cc
```diff
@@ -282,6 +289,31 @@ int main(int argc, char* argv[]) {
...
+        if (!sth.exists_) {
+          LOG(WARNING) << "Cluster has no Serving STH - leaving election.";
+          election_.StopElection();
```
pphaneuf: I don't see anything restarting the participation in the election once we get a serving STH or we catch up with the replication? Missing bit, or will be in a later pull request?

It's okay if it's for a later pull request, but I'd put a TODO until then.

cpp/server/ct-server.cc
AlCutter: Whoops, removed.

cpp/server/ct-server.cc
AlCutter: Coming up next... I'll pop a short-lived TODO in there.

pphaneuf:
LGTM

PR #440, opened 2015-01-23
Fix spurious races / hangs in tests.

AlCutter:


cpp/base/notification.cc
```diff
@@ -9,16 +9,20 @@ using std::unique_lock;
 
 namespace cert_trans {
 
+Notification::~Notification() {
+  // Ensure nobody is inside Notify() when we're destructing.
```
pphaneuf: Remove that bit, as we know it's not helping? I'd rather not artificially hide even more things...

AlCutter: It's definitely not fixing *this* problem, but ok, I'll take it out - just understand that the next time there's a 1 in 10,000 spurious hang, you've already volunteered to find it :)

pphaneuf:
LGTM, but I'll put it on the record that I'm annoyed.

PR #442, opened 2015-01-26
Join the election when we're up to date with local replication.

AlCutter:


cpp/log/cluster_state_controller-inl.h
```diff
@@ -206,6 +236,35 @@ void ClusterStateController<Logged>::CalculateServingSTH(
...
+              << " < Local contiguous_tree_size ("
+              << local_node_state_.contiguous_tree_size() << ")";
+    LOG(INFO) << "Local replication too far behind to be master - "
+                 "leaving election.";
+    election_->StopElection();
+    return;
+  }
+
+  // Otherwise, make sure we're joining in the election.
+  election_->StartElection();
```
pphaneuf: It would have been nice to know what caused us to rejoin, but I think that might be a bit too complicated...

pphaneuf:
LGTM

PR #443, opened 2015-01-26
Switch to C++11.

pphaneuf:
We've been using C++11 on the ```super_duper``` branch for a while, but there's no reason not to backport this (which helps minimize gratuitous differences).


eranmes:
LGTM, nice to see the dependencies reduced. Maybe we could even get it packaged for Debian one day.

PR #446, opened 2015-01-27
Add a 'strict' consistent store which only allows certain operations to be performed if the node is master.

AlCutter:


cpp/log/consistent_store.h
```diff
@@ -103,6 +103,8 @@ class ConsistentStore {
 
   ConsistentStore() = default;
 
+  virtual ~ConsistentStore() = default;
```
pphaneuf: Oops! Good catch.

cpp/log/strict_consistent_store.h
```diff
@@ -0,0 +1,93 @@
...
+#include "log/consistent_store.h"
+#include "util/masterelection.h"
+
+namespace cert_trans {
+
+// A wrapper around a ConsistentStore which will not allow changes to
+// the cluster state which should only be performed by the current master
+// unless this node /is/ the current master.
+//
+// Note that while this is better than just gating the start of an action
```
pphaneuf: How is this better? It seems to me that it *is* just gating the start of the mutations?

cpp/log/strict_consistent_store.h
```diff
@@ -0,0 +1,93 @@
...
+// (especially a long running action, e.g. a signing run) with a check to
```
pphaneuf: Or were you thinking at a higher level (a signing run would be longer, and this would start returning errors in the middle)?

cpp/log/strict_consistent_store.h
```diff
@@ -0,0 +1,93 @@
...
+
+  void WatchClusterConfig(
+      const typename ConsistentStore<Logged>::ClusterConfigCallback& cb,
+      util::Task* task) {
+    return peer_->WatchClusterConfig(cb, task);
+  }
+
+ private:
+  const MasterElection* const election_;  // Not owned by us
+  std::unique_ptr<ConsistentStore<Logged>> peer_;
```
pphaneuf: Const?

cpp/log/strict_consistent_store_test.cc
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+ protected:
+  bool IsMaster() const {
+    return GetParam();
+  }
+
+  NiceMock<MockMasterElection> election_;
+  // strict_store_ takes ownership of this:
+  NiceMock<MockConsistentStore<LoggedCertificate>>* peer_;
```
pphaneuf: For once, this *actually* needs to be a pointer! :smiley:

cpp/log/strict_consistent_store_test.cc
```diff
@@ -0,0 +1,134 @@
...
+  std::unique_ptr<StrictConsistentStore<LoggedCertificate>> strict_store_;
```
pphaneuf: But this doesn't. :wink:

cpp/server/ct-server.cc
```diff
@@ -244,13 +246,20 @@ int main(int argc, char* argv[]) {
           ? new FakeEtcdClient(event_base)
           : new EtcdClient(event_base, FLAGS_etcd_host, FLAGS_etcd_port));
 
+  // No real reason to let this be configurable per node; you can really
+  // shoot yourself in the foot that way by effectively running multiple
+  // distinct elections.
+  const string kLockDir("/election");
```
pphaneuf: Why isn't this just a literal, in the ```MasterElection```'s constructor arguments?

pphaneuf:
Done reviewing, nothing too serious.

cpp/log/strict_consistent_store.h
AlCutter: Yes - long operations would continue long past when mastership was lost, and it's a lot more likely that a number of operations on the cluster-wide state would be performed by a node which wasn't master.

AlCutter: (Is the comment not clear enough on the intent, should I change it?)

pphaneuf: That might be a good idea.

AlCutter: The existing comment already gives an example of a signing run being one of the long-running operations, do you have a suggestion for a better example, or a re-wording perhaps?

cpp/log/strict_consistent_store.h
AlCutter: Done

cpp/log/strict_consistent_store_test.cc
AlCutter: D'oh, you're right of course - this was a hang-over from when I was creating a new one in `SetUp()` each time while trying to narrow down the leaked Mock (which, of course, was actually the lack of the `virtual` d'tor.)
Unpointered.

cpp/log/strict_consistent_store.h
```diff
@@ -0,0 +1,93 @@
...
+
+  void WatchClusterConfig(
+      const typename ConsistentStore<Logged>::ClusterConfigCallback& cb,
+      util::Task* task) {
+    return peer_->WatchClusterConfig(cb, task);
+  }
+
+ private:
+  const MasterElection* const election_;  // Not owned by us
+  std::unique_ptr<ConsistentStore<Logged>> const peer_;
```
pphaneuf: Nitpicking: we usually put the ```const``` in front, for ```unique_ptr```.

cpp/log/strict_consistent_store_test.cc
```diff
@@ -0,0 +1,133 @@
...
+
+
+TEST_P(StrictConsistentStoreTest, TestNextAvailableSequenceNumber) {
+  if (IsMaster()) {
+    EXPECT_CALL(*peer_, NextAvailableSequenceNumber()).WillOnce(Return(123));
+  } else {
+    EXPECT_CALL(*peer_, NextAvailableSequenceNumber()).Times(0);
+  }
+
+  util::StatusOr<uint64_t> seq(strict_store_.NextAvailableSequenceNumber());
```
pphaneuf: This should be ```int64_t```, rather than ```uint64_t```.

cpp/log/strict_consistent_store.h
```diff
@@ -0,0 +1,93 @@
...
+
+namespace cert_trans {
+
+// A wrapper around a ConsistentStore which will not allow changes to
+// the cluster state which should only be performed by the current master
+// unless this node /is/ the current master.
+//
+// Note that while this is better than just gating the start of an action
+// (especially a long running action, e.g. a signing run) with a check to
+// IsMaster(), is it still necessarily racy because etcd doesn't support atomic
```
pphaneuf: "is it" should be "it is".

cpp/log/strict_consistent_store.h
pphaneuf: I'm re-reading this, and I think just adding "higher-level" to "gating the start of an action" should be sufficient to clarify this?

cpp/server/ct-server.cc
pphaneuf: Any answer on this one?

cpp/log/strict_consistent_store.h
AlCutter: added "high-level"

cpp/server/ct-server.cc
AlCutter: It's more about providing documentation - the variable name gives a local clue as to what that value is used for, and provides somewhere to anchor the comment which is intended to stop someone from helpfully changing a 'bad hardcoded constant' into a 'good FLAG' at some point in the future.

cpp/log/strict_consistent_store_test.cc
AlCutter: good spot.

cpp/log/strict_consistent_store.h
AlCutter: it is "is it", is it ?
*it is.*
it was, it is "it is" now.
*so it is.*

cpp/log/strict_consistent_store.h
AlCutter: brain fart, fixed.

pphaneuf:
LGTM

PR #448, opened 2015-01-29
Add initial cut of configs to run up SuperDuper on GKE.

AlCutter:
First cut at getting SuperDuper to run on GKE/Google Container Engine.
Few things more to do - notably stop relying on discovery.etcd.io service.


cloud/google/super_duper_replication.json
```diff
@@ -0,0 +1,67 @@
...
+          }],
+          "volumes": [{
+            "name": "log-data"
+          }]
+        }
+      },
+      "labels": {"name": "log-node"}
+    }
+  },
+  "lables": {"name": "log-node"}
```
pphaneuf: "lables"? "labels"?

Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM ubuntu
+RUN echo 'Building new SuperDuper Docker image...'
+COPY test/testdata/ca-cert.pem /tmp/
+RUN apt-get install -y software-properties-common && \
```
pphaneuf: Shouldn't that ```apt-get install``` be after the ```apt-get update```?

Makefile
```diff
@@ -6,7 +6,11 @@ all: cpp java python
 
 .PHONY: clean all alltests clean \
 	java java_test java_clean cpp cpp_test cpp_clean \
-	python python_test	
+	python python_test
+
+docker: cpp cpp_test
+	sudo docker build -t certificate_transparency/super_duper:test .
```
pphaneuf: Wouldn't that ```Dockerfile``` for ```ct-server``` itself make sense to be in that new ```cloud``` subdir as well?

cloud/google/super_duper_replication.json
```diff
@@ -0,0 +1,67 @@
...
+            }, {
+              "name": "peer",
+              "containerPort": 7001,
+              "hostPort": 7001
+            }],
+            "env": [{
+              "name": "DISCOVERY",
+              Get a fresh discovery URL from https://discovery.etcd.io/new
+              and put it in here - you'll need a fresh one EACH TIME you start
+              up a cluster - don't re-use them!
```
pphaneuf: I presume that this not being a "proper" comment is on purpose, so that you get a big red flag here?

cloud/google/super_duper_service.json
```diff
@@ -0,0 +1,12 @@
...
+  "id": "super-duper-service",
+  "apiVersion": "v1beta1",
+  "kind": "Service",
+  "selector": {
+    "name": "log-node"
+  },
+  "containerPort": "super-duper",
+  "protocol": "TCP",
+  "port": 80,
+  "createExternalLoadBalancer": true
```
pphaneuf: Sweet!

pphaneuf:
Review done, mostly good to go!

cloud/google/super_duper_replication.json
AlCutter: hmm, I guess it doesn't do much then :)
Fixed anyway FWIW.

Dockerfile
AlCutter: Well, perhaps there should be one before too, but that update is where it is in order to pull the index from the recently added repository.

Makefile
AlCutter: I put it here because it's building from the "source" that's also in this dir, seems to be a common pattern to have a `Dockerfile` at the root.  Technically I guess it could go elsewhere, but then you'd have to provide a path back to here anyway for the `ADD`/`COPY` commands in the `Dockerfile` to work (the etcd `Dockerfile` can go anywhere without worrying about the path because it doesn't use any local resources.)

cloud/google/super_duper_replication.json
AlCutter: Right.
(also, what's a "proper" JSON comment look like? AFAIK [which is not a lot when it comes to JSON] there's no provision for comments in the spec?)

Dockerfile
pphaneuf: Oh, ok, I didn't know that was the package that ```apt-add-repository``` came from... As you were, then.

Makefile
pphaneuf: Ok, makes sense.

cloud/google/super_duper_service.json
AlCutter: There's some docage on the Nginx site about this, basically the free one backs off a node if it gets an error back while serving a user request, the payware version does configurable OOB health checking.

cloud/google/super_duper_replication.json
pphaneuf: I figured maybe you were trying to get it to do something, and put the others because this one "didn't work"? But if this works okay, SGTM. :smiley:

cloud/google/super_duper_replication.json
pphaneuf: I guess the same as JavaScript? Which is to say, C/C++ style ```/* */``` and ```//```...

But it looks like the "official answer" is more or less "trololol": https://plus.sandbox.google.com/+DouglasCrockfordEsq/posts/RK8qyGVaGSr

cloud/google/super_duper_service.json
pphaneuf: And GKE uses nginx? I don't even know. :smiley:

pphaneuf:
LGTM

PR #449, opened 2015-01-29
Fix a bug, which involved clarifying the code.

pphaneuf:
When ```AsyncLogClient::GetEntries``` returned a non-OK status, we would return without calling ```WalkEntries```. If this range fetch happened to be the only one, that meant that the overall operation would get hung. I converted the range fetching to be its own ```util::Task```-driven operation, ensuring that ```WalkEntries``` would *always* get called, no matter what, and allowing me to remove the manual adding/removing of holds on the task (since child tasks have an implicit hold), making it one less thing to get wrong.

I also renamed some methods for better clarity and searchability.


AlCutter:
wot no tests? naughty boy.
LGTM.

PR #452, opened 2015-01-30
Add support for cancellation of fetches.

pphaneuf:
This will be important, as a change of cluster membership will require cancelling an existing fetch.


AlCutter:
LGTM

PR #455, opened 2015-01-30
Avoid CHECK fail when etcd changes leader.

AlCutter:


pphaneuf:
LGTM... I think? :smiley:

PR #458, opened 2015-02-02
Clean up old entries from etcd.

AlCutter:
Entries can be deleted from etcd's `/sequenced` and `/unsequenced` dirs once `/serving_sth` once they're covered by the cluster's `/serving_sth` (because that can only be updated once *sufficient* replication of those entries has occured.)

### DIFF
--- a/cpp/log/etcd_consistent_store-inl.h
+++ b/cpp/log/etcd_consistent_store-inl.h
@@ -7,6 +7,7 @@
 #include "log/etcd_consistent_store.h"
 #include "log/logged_certificate.h"
 #include "util/executor.h"
+#include "util/masterelection.h"
 #include "util/util.h"
 
 DECLARE_int32(node_state_ttl_seconds);
@@ -25,16 +26,18 @@ const char kNodesDir[] = "/nodes/";
 
 
 template <class Logged>
-EtcdConsistentStore<Logged>::EtcdConsistentStore(util::Executor* executor,
-                                                 EtcdClient* client,
-                                                 const std::string& root,
-                                                 const std::string& node_id)
+EtcdConsistentStore<Logged>::EtcdConsistentStore(
+    util::Executor* executor, EtcdClient* client,
+    const MasterElection* election, const std::string& root,
+    const std::string& node_id)
     : client_(CHECK_NOTNULL(client)),
+      election_(CHECK_NOTNULL(election)),
       sync_client_(client),
       root_(root),
       node_id_(node_id),
       serving_sth_watch_task_(CHECK_NOTNULL(executor)),
-      received_initial_sth_(false) {
+      received_initial_sth_(false),
+      exiting_(false) {
   // Set up watches on things we're interested in...
   WatchServingSTH(
       std::bind(&EtcdConsistentStore<Logged>::OnEtcdServingSTHUpdated, this,
@@ -47,6 +50,9 @@ EtcdConsistentStore<Logged>::EtcdConsistentStore(util::Executor* executor,
     std::unique_lock<std::mutex> lock(mutex_);
     serving_sth_cv_.wait(lock, [this]() { return received_initial_sth_; });
   }
+  // Now we can start a clean up thread.
+  clean_up_thread_.reset(new std::thread(
+      std::bind(&EtcdConsistentStore<Logged>::CleanUpEntriesThread, this)));
 }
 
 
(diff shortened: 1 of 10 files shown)


cpp/log/etcd_consistent_store-inl.h
```diff
@@ -56,6 +62,15 @@ EtcdConsistentStore<Logged>::~EtcdConsistentStore() {
...
   VLOG(1) << "Waiting for watch tasks to return.";
   serving_sth_watch_task_.Wait();
+  VLOG(1) << "Joining cleanup thread";
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    exiting_ = true;
+  }
+  serving_sth_cv_.notify_all();
+  clean_up_thread_->join();
+  clean_up_thread_.reset();
```
pphaneuf: Not *really* necessary, right? It's fine to leave like this, just checking that I'm not missing some subtle thing...

AlCutter: Yup, I just like tidying these up for some reason.

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -544,6 +569,96 @@ void EtcdConsistentStore<Logged>::OnEtcdServingSTHUpdated(
...
+      LOG(WARNING) << "Cleanup couldn't get unsequenced entry " << status;
+    } else {
+      status =
+          DeleteEntry(GetUnsequencedPath(unseq_entry.Entry()), &unseq_entry);
+      if (!status.ok()) {
+        LOG(WARNING) << "Cleanup couldn't delete unsequenced entry " << status;
+        return status;
+      }
+    }
+    VLOG(1) << "Cleanup deleted unsequenced entry " << entry.Entry().Hash();
```
pphaneuf: I keep having a feeling that this (and the other log messages in this method) ought to have the sequence number, not just the hash?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -544,6 +569,96 @@ void EtcdConsistentStore<Logged>::OnEtcdServingSTHUpdated(
...
+      return util::Status::OK;
+    }
+
+    // Otherwise, first we delete the corresponding unsequenced entry
+    EntryHandle<LoggedCertificate> unseq_entry;
+    status = GetPendingEntryForHash(entry.Entry().Hash(), &unseq_entry);
+    if (!status.ok()) {
+      // Log a warning here, but don't bail on the clean up as we could just be
+      // recovering from a crash halfway through a previous clean up.
+      LOG(WARNING) << "Cleanup couldn't get unsequenced entry " << status;
```
pphaneuf: When you print a ```Status```, having ": " at the end looks nicer. :smiley:

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -544,6 +569,96 @@ void EtcdConsistentStore<Logged>::OnEtcdServingSTHUpdated(
...
+        return status;
+      }
+    }
+    VLOG(1) << "Cleanup deleted unsequenced entry " << entry.Entry().Hash();
+
+    // Now delete the sequenced entry:
+    status =
+        DeleteEntry(GetSequencedPath(entry.Entry().sequence_number()), &entry);
+    if (!status.ok()) {
+      LOG(WARNING) << "Cleanup couldn't delete sequenced entry " << status;
```
pphaneuf: Same here, add ":" after "entry".

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -544,6 +569,96 @@ void EtcdConsistentStore<Logged>::OnEtcdServingSTHUpdated(
...
+      return status;
+    }
+    VLOG(1) << "Cleanup deleted sequenced entry "
```
pphaneuf: This didn't fit on one line?

pphaneuf:
LGTM, just some very uninteresting tidying up.

cpp/log/etcd_consistent_store-inl.h
AlCutter: Yeah, you're right - added.

cpp/log/etcd_consistent_store-inl.h
AlCutter: What'd be lovely is if the status code got translated to the descriptive string, too :)

cpp/log/etcd_consistent_store-inl.h
AlCutter: added

cpp/log/etcd_consistent_store-inl.h
AlCutter: Does now.

PR #459, opened 2015-02-02
Export host:port info for clustered nodes.

AlCutter:


cpp/server/ct-server.cc
```diff
@@ -298,10 +298,15 @@ int main(int argc, char* argv[]) {
     // if it's not set, which in turn causes us to not attempt to become
     // master:
     consistent_store.SetServingSTH(tree_signer.LatestSTH());
+  } else {
+    CHECK(!FLAGS_server.empty());
+    CHECK(FLAGS_server != "localhost");
```
pphaneuf: Maybe ```CHECK_NE```?

pphaneuf:
LGTM, just a nit, really.

PR #462, opened 2015-02-03
Swap two lines that were erroneously transposed.

phad:
Looks like someone hit the editor key that swaps the line under the cursor with the line above.
The code as it stands does not reset initialized_ properly at the end of Sha256Hasher::Final().

### DIFF
--- a/cpp/merkletree/serial_hasher.cc
+++ b/cpp/merkletree/serial_hasher.cc
@@ -28,8 +28,8 @@ string Sha256Hasher::Final() {
 
   unsigned char hash[SHA256_DIGEST_LENGTH];
   SHA256_Final(hash, &ctx_);
-  return string(reinterpret_cast<char*>(hash), SHA256_DIGEST_LENGTH);
   initialized_ = false;
+  return string(reinterpret_cast<char*>(hash), SHA256_DIGEST_LENGTH);
 }
 
 SerialHasher* Sha256Hasher::Create() const {
--- a/cpp/merkletree/serial_hasher.h
+++ b/cpp/merkletree/serial_hasher.h
@@ -16,8 +16,10 @@ class SerialHasher {
 
   virtual size_t DigestSize() const = 0;
 
-  // Reset the context. Must be called before the first
-  // Update() call (and again after each Final() call).
+  // Reset the context. Must be called before the first Update() call.
+  // Optionally it can be called after each Final() call; however
+  // doing so is a no-op since Final() will leave the hasher in a
+  // reset state.
   virtual void Reset() = 0;
 
   // Update the hash context with (binary) data.


AlCutter:
Are we sure that there's no code which relies on the fact that that `Update(...)`, `Update(...)`, `Final()` will hash all the data and not just the data provided by the latest `Update(...)` ?

eranmes:
(capturing an offline discussion with Paul)
Update(..), Update(..), Final() is fine. What's wrong is Update(..), Final(), Update(..), Final(..) without calling Reset() after Final().
I suggested adding a test and in Update, Final, rather than calling Reset() if initialized_ is false, CHECK that initialized_ is true. To make client code even simpler, Reset can be removed and Update/Final could reset for the client code if initialized_ is false.

phad:
I'm not sure.  My understanding is that the protocol callers should follow is:
Sha256Hasher h;
h.Reset();
[h.Update();  0 or more times]
h.Final();
--- after this h can be re-used for another hashing ---
h.Reset();
[h.Update(); 0 or more times]
h.Final();

The convenience calls being made to Reset() in the implementation of Update() and Final() allowed the protocol to be relaxed.
eg. these were legal:

Sha256Hasher h;
h.Final();
---
Sha256Hasher h;
h.Update();  // 1 or more times
h.Final();

The code as it stood would give buggy results then if the caller did:

---
Sha256Hasher h;
h.Update();  // 1 or more times
h.Final();
h.Final();   // this call would yield hashed data dependent on the earlier Updated data

The latter case is definitely a bug.

I think strictly enforcing the Reset() ...[Update()]* ... Final()  protocol is the safe and correct fix.

pphaneuf:
LGTM

Maybe adding a ```CHECK(initialized_)``` to ```Update``` and ```Final``` is a thing to do (I don't think so myself), but the change in this commit is undoubtedly good, IMHO.

phad:
initialized_ is now CHECKed in both Update() and Final(); added some Death tests for these.
PTAL

pphaneuf:
I... don't really like that, because we almost certainly do *not* have 100% test coverage, and that makes something that used to be valid crash? I'm adding enough legit crashes as it is, no? :wink:

phad:
Confused, Pierre:  I just implemented what you suggested.

eranmes:
LGTM for the original change as-is (maybe with a test to show the problem it fixed).
I agree with @pphaneuf that we don't have good enough test coverage for all client code that uses it.
My suggestion would be to leave the current behaviour, which means the Reset() call is optional, just document it.

pphaneuf:
Sorry, I wasn't totally clear there.

What I meant in my comment was that what you had in your first commit is **definitely good**, fixing an actual bug, and that while @eranmes was suggesting adding some ```CHECK``` in his first comment, you should push your existing commit right away, and discuss other changes separately. I actually had a "(I don't think so myself)" in that comment, because, well, I don't like this. :smiley:

eranmes:
Per our offline discussion, I'd be happy LGTM-ing the original change + changes to documentation in the header file that says calling Reset() after Final() is optional.

pphaneuf:
LGTM

pphaneuf:
LGTM

PR #464, opened 2015-02-04
Add mapping from LogVerifier::INVALID_MERKLE_PATH to debug string.

phad:
The VerifyResultString() function got out of sync with the enum it accepts.

### DIFF
--- a/cpp/log/log_verifier.h
+++ b/cpp/log/log_verifier.h
@@ -44,6 +44,8 @@ class LogVerifier {
         return "Inconsistent timestamps.";
       case INVALID_SIGNATURE:
         return "Invalid signature.";
+      case INVALID_MERKLE_PATH:
+        return "Invalid Merkle path.";
       default:
         assert(false);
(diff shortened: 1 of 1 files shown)


cpp/log/log_verifier.h
```diff
@@ -44,6 +44,8 @@ class LogVerifier {
         return "Inconsistent timestamps.";
       case INVALID_SIGNATURE:
         return "Invalid signature.";
+      case INVALID_MERKLE_PATH:
+        return "Invalid Merkle path.";
```
pphaneuf: And that's why I don't like the "default:" (removing it would give a compile error, with the compilers we're using).

pphaneuf:
LGTM

cpp/log/log_verifier.h
phad: Yeah, in this particular case a harmless default would have been nice.  But still at least the assertion failure alerted me to this missing case.

PR #465, opened 2015-02-04
Replace uses of assert() with the glog equivalents.

pphaneuf:
I meant to point that out in #464, and then I thought, what the heck, let's fix it all, for once and for good!

### DIFF
--- a/cpp/log/log_verifier.h
+++ b/cpp/log/log_verifier.h
@@ -2,6 +2,7 @@
 #ifndef LOG_VERIFIER_H
 #define LOG_VERIFIER_H
 
+#include <glog/logging.h>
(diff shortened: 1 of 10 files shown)


cpp/log/log_verifier.h
```diff
@@ -47,7 +48,7 @@ class LogVerifier {
       case INVALID_MERKLE_PATH:
         return "Invalid Merkle path.";
       default:
-        assert(false);
+        CHECK(false);
```
pphaneuf: You say the following ```return``` will now execute? It's *never* going to execute, you mean?

I'll add something to the ```CHECK```, though, good call.

PR #470, opened 2015-02-06
Add a tool for managing a cluster.

AlCutter:
Currently only has a method to initialise the default entries in etcd.
Laid out as it is with the intention of being able to call into it from `ct-server.cc` when that's being run in "standalone" mode in order to get its `FakeEtcd` initialised too.

### DIFF
--- a/cpp/tools/clustertool-inl.h
+++ b/cpp/tools/clustertool-inl.h
@@ -1,0 +1,60 @@
+#ifndef CERT_TRANS_TOOLS_CLUSTERTOOL_INL_H_
+#define CERT_TRANS_TOOLS_CLUSTERTOOL_INL_H_
+#include <event2/thread.h>
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+#include <google/protobuf/text_format.h>
+#include <memory>
+#include <openssl/err.h>
+#include <string>
+
+#include "util/etcd.h"
+#include "log/etcd_consistent_store.h"
+#include "log/logged_certificate.h"
+#include "log/log_signer.h"
+#include "log/sqlite_db.h"
+#include "log/strict_consistent_store.h"
+#include "log/tree_signer.h"
+#include "proto/ct.pb.h"
+#include "tools/clustertool.h"
+#include "util/masterelection.h"
+#include "util/read_key.h"
+#include "util/status.h"
+#include "util/thread_pool.h"
+
+namespace cert_trans {
+
+// TODO(alcutter): have a way of passing in other configs
+const char kDefaultClusterConfig[] =
+    "minimum_serving_nodes: 2\n"
+    "minimum_serving_fraction: 0.75\n";
+
+
+// static
+template <class Logged>
+util::Status ClusterTool<Logged>::InitLog(
+    TreeSigner<Logged>* tree_signer,
+    ConsistentStore<Logged>* consistent_store) {
+  if (tree_signer->UpdateTree() != TreeSigner<LoggedCertificate>::OK) {
+    return util::Status(util::error::UNKNOWN, "Failed to Update Tree");
+  }
+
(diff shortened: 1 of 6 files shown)


cpp/tools/clustertool-inl.h
```diff
@@ -0,0 +1,60 @@
...
+  util::Status status(
+      consistent_store->SetServingSTH(tree_signer->LatestSTH()));
+  if (!status.ok()) {
+    return status;
+  }
+
+  ct::ClusterConfig config;
+  if (!google::protobuf::TextFormat::ParseFromString(kDefaultClusterConfig,
+                                                     &config)) {
+    return util::Status(util::error::FAILED_PRECONDITION,
```
pphaneuf: More like ```INVALID_ARGUMENT```, no?

> INVALID_ARGUMENT indicates arguments that are problematic regardless of the state of the system (e.g., a malformed file name).

> Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed.

cpp/tools/clustertool.h
```diff
@@ -0,0 +1,24 @@
+#ifndef CERT_TRANS_TOOLS_CLUSTERTOOL_H_
+#define CERT_TRANS_TOOLS_CLUSTERTOOL_H_
+
+namespace cert_trans {
+
+
+template <class Logged>
```
pphaneuf: That *this* has to be templated makes me cry a little...

cpp/tools/clustertool_main.cc
```diff
@@ -0,0 +1,110 @@
...
+    status = ClusterTool<LoggedCertificate>::InitLog(&tree_signer,
+                                                     &consistent_store);
+  } else {
+    Usage();
+  }
+
+  LOG(INFO) << status;
+  election.StopElection();
+
+  return status.ok() ? 0 : 1;
```
pphaneuf: Suggestion:

    return status.error_code();

You could also return ```util::error::INVALID_ARGUMENT``` at line 63, to match.

cpp/tools/clustertool_main.cc
```diff
@@ -0,0 +1,110 @@
...
+  CHECK_EQ(pkey.status(), util::Status::OK);
+  LogSigner log_signer(pkey.ValueOrDie());
+
+  const shared_ptr<libevent::Base> event_base(make_shared<libevent::Base>());
+  std::unique_ptr<libevent::EventPumpThread> pump(
+      new libevent::EventPumpThread(event_base));
+
+  EtcdClient etcd_client(event_base, FLAGS_etcd_host, FLAGS_etcd_port);
+
+  const string kLockDir("/election");
```
pphaneuf: Doesn't have to be in this PR, but we should consider putting this somewhere that's shared with ```ct-server.cc```.

pphaneuf:
Minor thing: I'd rename this ```ct-clustertool```, so it'd still make sense when installed in ```/usr/bin```?

Otherwise done with this review.

cpp/tools/clustertool-inl.h
AlCutter: Yeah, good call - currently it's kinda `FAILED_PRECONDITION` because the only possible `ClusterConfig` is hardcoded :D  But it should really be `INVALID_ARGUMENT` as we will support custom configs RSN.

cpp/tools/clustertool.h
AlCutter: It doesn't *have* to be, it *could* totally be a `clustertool` which only works with instances of `LoggedCertificate` servers :)

cpp/tools/clustertool_main.cc
AlCutter: Cool, nice idea - thanks.

pphaneuf:
LGTM

PR #471, opened 2015-02-06
Python client: custom timeout for POST

eranmes:
That's where the timeout is really needed...


phad:
Code change looks fine, can you check the travis failure is not caused by your change before merging?

eranmes:
Seems like flakiness in the C++ tests - all Python tests pass.

PR #473, opened 2015-02-09
Python client: Plumb ca_bundle to requests

eranmes:
In order to verify the SSL certificate received for https requests, the
ca_bundle has to be passed into the requests module. Needed when the
ca_bundle is not in the standard place.

### DIFF
--- a/python/ct/client/log_client.py
+++ b/python/ct/client/log_client.py
@@ -196,8 +196,9 @@ def _parse_consistency_proof(response, servername):
 # A class that we can mock out to generate fake responses.
 class RequestHandler(object):
     """HTTPS requests."""
-    def __init__(self, connection_timeout=60):
+    def __init__(self, connection_timeout=60, ca_bundle=True):
       self._timeout = connection_timeout
+      self._ca_bundle = ca_bundle
 
     def __repr__(self):
         return "%r()" % self.__class__.__name__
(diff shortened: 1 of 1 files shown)


python/ct/client/log_client.py
```diff
@@ -196,8 +196,9 @@ def _parse_consistency_proof(response, servername):
 # A class that we can mock out to generate fake responses.
 class RequestHandler(object):
     """HTTPS requests."""
-    def __init__(self, connection_timeout=60):
+    def __init__(self, connection_timeout=60, ca_bundle=True):
```
phad: ca_bundle doesn't sound like something you'd pass as a boolean value.
is it really a flag?

eranmes: The ca_bundle is being passed into verify, which can take one of the following values:
True: SSL certificate validation will take place using the default CA root certs set.
False: SSL certificate validation will not take place.
Path to a file: SSL certificate validation will take place using the CA root set specified in this file.

See requests documentation:
http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification

phad: Yeuch.  An argument that can be given a boolean value, or string.

benlaurie:
Hmm ... given the weirdness (boolean vs. string):

a) Documentation?

b) Should we decline a false setting? We should not encourage disabling cert checking!

PR #474, opened 2015-02-09
Remove contiguous_tree_size from CNS

AlCutter:


proto/ct.proto
```diff
@@ -206,7 +206,7 @@ message SSLClientCTData {
 
 message ClusterNodeState {
   optional string node_id = 1;
-  optional int64 contiguous_tree_size = 2;
+  optional int64 _DEPRECATED_0 = 2;
```
pphaneuf: Could put a ```[deprecated=true]``` annotation instead?

AlCutter: Didn't know about that - this is the old school way :)

cpp/server/ct-server.cc
```diff
@@ -317,17 +317,23 @@ int main(int argc, char* argv[]) {
...
   PeriodicClosure node_state_refresh(
       event_base, std::chrono::seconds(FLAGS_node_state_refresh_seconds),
       [&db, &cluster_controller, &internal_pool]() {
         // Actually run this on the thread pool, since this periodic closure is
         // calling back on the eventloop thread.
         internal_pool.Add([&db, &cluster_controller]() {
-          cluster_controller.ContiguousTreeSizeUpdated(db->TreeSize());
+          SignedTreeHead latest_sth;
+          if (db->LatestTreeHead(&latest_sth) !=
+              Database<LoggedCertificate>::LOOKUP_OK) {
```
pphaneuf: Isn't this going to compete with the tree signer? If I'm not mistaken, the STH in the database is the serving STH, and the tree signer puts up the locally signed STH, no?

pphaneuf:
Not 100% sure that it's necessary, but rebasing this to the newest HEAD might be a good idea?

cpp/server/ct-server.cc
AlCutter: Dammit, yes.
Annoying because I was halfway through wiring in the `TreeSigner` here and then had an idea that just using the DB would be "easier" (not realising that it came with a free side of "completely wrong".)
I've re-done and finished off what I was going to do originally.

AlCutter:
It was at HEAD when I sent it :)

pphaneuf:
Gotta keep up, man! :wink:

pphaneuf:
LGTM, pending Travis.

PR #476, opened 2015-02-09
Python client: Document ca_bundle parameter

eranmes:
Also disallow false values that would disable SSL certificate checking.

### DIFF
--- a/python/ct/client/log_client.py
+++ b/python/ct/client/log_client.py
@@ -253,9 +253,23 @@ class RequestHandler(object):
 class LogClient(object):
     """HTTP client for talking to a CT log."""
 
+    """Create a new log client.
+
+    Args:
+        uri: The CT Log URI to communicate with.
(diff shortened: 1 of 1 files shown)


python/ct/client/log_client.py
```diff
@@ -253,9 +253,23 @@ def post_response_body(self, uri, post_data=None):
...
+        ca_bundle: True or a file path containing a set of CA roots. See
+        Requests documentation for more information:
+        http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification
+        Note that a false-y value is not allowed.
+    """
     def __init__(self, uri, handler=None, connection_timeout=60,
                  ca_bundle=True):
         self._uri = uri
+        if not ca_bundle:
+          raise ClientError("Refusing to turn of SSL certificate checking.")
```
phad: s/of/off/

phad:
This is better, thanks for adding the docs.

PR #477, opened 2015-02-09
Add a set_config command to the cluster_tool

AlCutter:


cpp/tools/clustertool.h
```diff
@@ -10,9 +18,15 @@ class ClusterTool {
...
   //  - Creates the /cluster_config entry.
-  static util::Status InitLog(TreeSigner<Logged>* tree_signer,
+  static util::Status InitLog(const ct::ClusterConfig& cluster_config,
+                              TreeSigner<Logged>* tree_signer,
                               ConsistentStore<Logged>* consistent_store);
 
+  // Sets the cluster config
+  static util::Status SetClusterConfig(
+      const ct::ClusterConfig& cluster_config,
+      ConsistentStore<Logged>* consistent_store);
```
pphaneuf: I was already feeling that in the previous pull request: either those should be free functions, or that class should start pulling its weight a bit and have something (like the tree signer and the consistent store)? Even then, that wouldn't really be state, just all const stuff, so I'd probably favour free functions? But I'll leave it to you...

cpp/tools/clustertool_main.cc
```diff
@@ -47,64 +66,121 @@ void Usage() {
...
+            << "  initlog     Initialise a new log.\n"
+            << "  set_config  Set/Change a cluster's config.\n";
 }
 
 
+unique_ptr<LogSigner> BuildLogSigner() {
+  CHECK(!FLAGS_key.empty());
+  util::StatusOr<EVP_PKEY*> pkey(ReadPrivateKey(FLAGS_key));
+  CHECK_EQ(pkey.status(), util::Status::OK);
+  return unique_ptr<LogSigner>(new LogSigner(pkey.ValueOrDie()));
```
pphaneuf: I really wish there was a ```make_unique<LogSigner>(pkey.ValueOrDie())```, this is some bad case of stuttering... The "solution" is Go, I suppose? :wink:

pphaneuf:
LGTM

cpp/tools/clustertool.h
AlCutter: Yeah, fair point - I've removed their shackles and they're free to gambol through the `.text SECTION` as they please.

pphaneuf:
Looks even gooder to me!

PR #478, opened 2015-02-09
Keep the etcd key with EntryHandle, so that we know which entry was deleted when watching a directory.

AlCutter:


cpp/log/etcd_consistent_store-inl.h
```diff
@@ -448,15 +457,15 @@ util::Status EtcdConsistentStore<Logged>::CreateEntry(const std::string& path,
 
 template <class Logged>
 template <class T>
-util::Status EtcdConsistentStore<Logged>::ForceSetEntry(
-    const std::string& path, EntryHandle<T>* t) {
+util::Status EtcdConsistentStore<Logged>::ForceSetEntry(EntryHandle<T>* t) {
   CHECK_NOTNULL(t);
-  CHECK(!t->HasHandle());
```
pphaneuf: Wouldn't you want to keep that check too?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -467,15 +476,14 @@ util::Status EtcdConsistentStore<Logged>::ForceSetEntry(
 template <class Logged>
 template <class T>
 util::Status EtcdConsistentStore<Logged>::ForceSetEntryWithTTL(
-    const std::string& path, const std::chrono::seconds& ttl,
-    EntryHandle<T>* t) {
+    const std::chrono::seconds& ttl, EntryHandle<T>* t) {
   CHECK_NOTNULL(t);
-  CHECK(!t->HasHandle());
```
pphaneuf: Same here?

pphaneuf:
Other than that, LGTM (feel free to submit after putting those two checks back).

cpp/log/etcd_consistent_store-inl.h
AlCutter: I'd convinced myself that it wasn't necessary, but then we spoke about it offline and you reminded me of my arguments for having it there when I put it in, so I've re-instated it and added a comment explaining why it's there [to remind me, at least, next time I go to delete it.]

cpp/log/etcd_consistent_store-inl.h
AlCutter: yup

PR #481, opened 2015-02-10
Fix two small mistakes.

pphaneuf:
The ```CHECK_NOTNULL``` only improves the logs, but the missing ```public``` keyword has an impact on an upcoming pull requests.


AlCutter:
LGTM

PR #482, opened 2015-02-10
Make TreeHasher thread-safe.

pphaneuf:
Also, a few other cleanups: formatting, hiding constants away, improved const-correctness, etc...

### DIFF
--- a/cpp/merkletree/tree_hasher.cc
+++ b/cpp/merkletree/tree_hasher.cc
@@ -10,8 +10,8 @@ using std::string;
 
 namespace {
 
-const string kLeafPrefix(1, '\x00');
-const string kNodePrefix(1, '\x01');
+const char kLeafPrefix('\x00');
+const char kNodePrefix('\x01');
 
 std::string EmptyHash(SerialHasher* hasher) {
   hasher->Reset();
@@ -27,7 +27,7 @@ TreeHasher::TreeHasher(SerialHasher* hasher)
 string TreeHasher::HashLeaf(const string& data) const {
   lock_guard<mutex> lock(lock_);
   hasher_->Reset();
-  hasher_->Update(kLeafPrefix);
+  hasher_->Update(string(1, kLeafPrefix));
   hasher_->Update(data);
   return hasher_->Final();
 }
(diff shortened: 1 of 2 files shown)


cpp/merkletree/tree_hasher.cc
```diff
@@ -1,31 +1,31 @@
...
+using std::mutex;
 using std::string;
 
-const string TreeHasher::kLeafPrefix(1, '\x00');
-const string TreeHasher::kNodePrefix(1, '\x01');
+namespace {
 
-TreeHasher::TreeHasher(SerialHasher* hasher) : hasher_(hasher) {
-}
+const string kLeafPrefix(1, '\x00');
```
AlCutter: Could you make these `const char` to avoid the styleguide forbidden global var of a complex class?

cpp/merkletree/tree_hasher.h
```diff
@@ -10,28 +13,28 @@ class TreeHasher {
...
   // Takes ownership of the SerialHasher.
   TreeHasher(SerialHasher* hasher);
-  ~TreeHasher();
 
   size_t DigestSize() const {
     return hasher_->DigestSize();
   }
 
-  std::string HashEmpty();
+  const std::string& HashEmpty() const {
```
AlCutter: `EmptyHash()` to match the var name?

AlCutter:
All done.
`SerialHasher` instances aren't shared anywhere, are they?

cpp/merkletree/tree_hasher.h
pphaneuf: I thought about it, but it causes a cascade of other changes, didn't really want to get into that for now...

cpp/merkletree/tree_hasher.cc
pphaneuf: Done.

pphaneuf:
I haven't verified that ```SerialHasher``` instances aren't shared, but as TreeSigner takes ownership, that would be unlikely? I've pushed an update now, I'll look around while waiting for your LGTM...

cpp/merkletree/tree_hasher.h
AlCutter: oke doke.

AlCutter:
LGTM

PR #484, opened 2015-02-10
Make LogLookup thread-safe.

pphaneuf:
This might belong on ```master```, but I wasn't sure about the ramifications...


AlCutter:
`master` is all single-threaded and blocking isn't it?
LGTM anyway

pphaneuf:
It's been improved a bit, so it's not completely blocking, but it's not heavily threaded either. I believe that this part of it is single threaded (signing is only done from one place).

PR #485, opened 2015-02-10
Add a continuous fetcher.

pphaneuf:


pphaneuf:
Added a commit.

cpp/log/cluster_state_controller_test.cc
```diff
@@ -103,6 +107,7 @@ class ClusterStateControllerTest : public ::testing::Test {
 
   ThreadPool pool_;
   shared_ptr<libevent::Base> base_;
+  StrictMock<MockContinuousFetcher> fetcher_;
```
pphaneuf: I think this thing is actually a bit racy (Travis failed because of that), maybe I should use a NiceMock instead? That's not really the bit being tested here, maybe just ignoring everything is the right thing to do?

cpp/fetcher/continuous_fetcher.cc
```diff
@@ -0,0 +1,157 @@
...
+using std::move;
+using std::mutex;
+using std::placeholders::_1;
+using std::shared_ptr;
+using std::string;
+using std::unique_lock;
+using std::unique_ptr;
+using util::Task;
+
+DEFINE_int32(delay_between_fetches_secs, 30, "delay between fetches");
```
AlCutter: /secs/seconds/ (just consistency because there are other flags which use "seconds")

cpp/fetcher/continuous_fetcher.cc
```diff
@@ -0,0 +1,157 @@
...
+void ContinuousFetcherImpl::FetchDone(Task* task) {
+  if (!task->status().ok()) {
+    LOG(WARNING) << "error while fetching: " << task->status();
+  }
+
+  lock_guard<mutex> lock(lock_);
+  fetch_task_.reset();
+
+  if (restart_fetch_) {
+    base_->Add(bind(&ContinuousFetcherImpl::FetchDelayDone, this, nullptr));
```
AlCutter: does this `nullptr` not just immediately blow up on the `CHECK_NOTNULL` on L137?

pphaneuf: Jeez, I added the ```CHECK_NOTNULL``` by instinct *just* before sending this for review. :stuck_out_tongue: 

I'll replace it with a comment...

AlCutter:
LGTM

PR #486, opened 2015-02-10
Fiddle with compute stuff

AlCutter:
- split the pods out so that etcd and the log server are logically separate entities.
- add some handy turnup/down scripts
- increase etcd timeouts


pphaneuf:
LGTM

PR #489, opened 2015-02-11
Keep track of the etcd index early on when handling responses, and apply a timeout to HTTP requests

AlCutter:


cpp/util/etcd.cc
```diff
@@ -46,6 +46,9 @@ DEFINE_bool(etcd_consistent, true, "Add consistent=true param to all requests. "
             "Do not turn this off unless you *know* what you're doing.");
 DEFINE_bool(etcd_quorum, true, "Add quorum=true param to all requests. "
             "Do not turn this off unless you *know* what you're doing.");
+DEFINE_int32(etcd_connection_timeout_seconds, 10,
+             "Number of seconds after "
+             "which to timeout etcd connections.");
```
pphaneuf: Is it me, or clang-format struck again?

cpp/util/etcd.cc
```diff
@@ -607,16 +612,22 @@ Status EtcdClient::WatchState::HandleSingleValueRequestDone(
...
+                                         const shared_ptr<JsonObject>& json,
+                                         int64_t etcd_index) {
+  // TODO(alcutter): doing this here works around some etcd 401 errors, but in
+  // the case of sustained high qps we could miss updates entirely (e.g. if
+  // this new index is already past the 1000 entry horizon by the time we make
+  // the new watch request.)  One way to address this might be to have the
+  // watcher re-do an "initial" get on the target, and, in the case of
+  // directory watches, maintain a set of known keys so that it can synthesise
+  // 'delete' updates.
+  highest_index_seen_ = max(highest_index_seen_, etcd_index);
```
pphaneuf: Could port the ```CHECK_LT``` that was in ```HandleSingleValueRequestDone```? This really should be higher, if it starts going backward, something is Very Wrong (tm).

cpp/util/etcd.cc
AlCutter: d'oh, fixed.

cpp/util/etcd.cc
```diff
@@ -46,6 +46,9 @@ DEFINE_bool(etcd_consistent, true, "Add consistent=true param to all requests. "
             "Do not turn this off unless you *know* what you're doing.");
 DEFINE_bool(etcd_quorum, true, "Add quorum=true param to all requests. "
             "Do not turn this off unless you *know* what you're doing.");
+DEFINE_int32(etcd_connection_timeout_seconds, 10,
```
pphaneuf: We might have had a split of "secs" vs "seconds" in our flags (look at line 43). :wink:

cpp/util/etcd.cc
```diff
@@ -628,6 +639,8 @@ void EtcdClient::WatchState::RequestDone(Status status,
   {
     // This is probably due to a timeout, just retry.
     if (!status.ok()) {
+      LOG(INFO) << "Watch request failed (" << status << "), will retry after "
```
pphaneuf: INFO? WARNING? One of those is evil because of flushes, right? I guess this is INFO enough (since we're retrying it)...

It'd be nice to have the log messages consistent for these errors, though. I don't think we need to mention when we'll retry?

cpp/util/etcd.cc
AlCutter: Yah, as commented IRL, I'll make it `CHECK_LE` here because I'm not 100% sure that there couldn't be some sort of etcd spasm where we get multiple `RequestDone()` coming back at the same index. But as you say, if someone's going to violate causality, they can do it somewhere else.

```
We don't allow faster than light neutrinos in here, said the bartender.
A neutrino walks into a bar.
```

cpp/util/etcd.cc
AlCutter: not any more :)

cpp/util/etcd.cc
AlCutter: Yeah, WARNING and above flush, changed all these to INFO and unified the format a bit.

pphaneuf:
LGTM

PR #492, opened 2015-02-12
Don't try to track AddPeer calls in this test.

pphaneuf:
It is futile, and just makes the test flaky for no interesting reason.


AlCutter:
LGTM

PR #494, opened 2015-02-12
Add a pool of evhttp_connection.

pphaneuf:


cpp/net/connection_pool.h
```diff
@@ -0,0 +1,53 @@
...
+  evhttp_connection_unique_ptr Get(const URL& url);
+  void Put(evhttp_connection_unique_ptr&& conn);
+
+ private:
+  libevent::Base* const base_;
+
+  std::mutex lock_;
+  // We get and put connections from the back of the deque, and when
+  // there are too many, we prune them from the front (LIFO).
+  std::map<std::pair<std::string, uint16_t>,
```
AlCutter: Could typedef `std::pair<std::string, uint16_t>` to a `HostPortPair` or similar for a small increase in readability and longer life for our `:`, `<` and `>` keys.

cpp/net/connection_pool.cc
```diff
@@ -0,0 +1,62 @@
...
+using std::mutex;
+using std::pair;
+using std::string;
+
+namespace cert_trans {
+namespace internal {
+
+namespace {
+
+const int kMaxConnPerHostPort = 4;
```
AlCutter: flag?

cpp/net/connection_pool.cc
```diff
@@ -0,0 +1,62 @@
...
+  it->second.pop_back();
+
+  return retval;
+}
+
+
+void ConnectionPool::Put(evhttp_connection_unique_ptr&& conn) {
+  char* host;
+  uint16_t port;
+  evhttp_connection_get_peer(conn.get(), &host, &port);
```
AlCutter: `CHECK(conn)` ?
Shouldn't really happen, but...

pphaneuf: I think silently ignoring that is the best course of action.

pphaneuf:
Comments addressed.

AlCutter:
LGTM

PR #496, opened 2015-02-12
Add a URL fetcher class.

pphaneuf:
Also convert AsyncLogClient to use it.

Still many things can be improved on it (deadlines, retries, cancellations, etc), but at least it provides a reasonable shell for these features to be added.

### DIFF
--- a/cpp/client/async_log_client.cc
+++ b/cpp/client/async_log_client.cc
@@ -1,22 +1,20 @@
 #include "client/async_log_client.h"
 
 #include <algorithm>
+#include <event2/http.h>
 #include <glog/logging.h>
 #include <iterator>
 #include <memory>
-#include <sstream>
 
 #include "log/cert.h"
 #include "proto/serializer.h"
 #include "util/json_wrapper.h"
-#include "util/libevent_wrapper.h"
-
-namespace libevent = cert_trans::libevent;
 
 using cert_trans::AsyncLogClient;
 using cert_trans::Cert;
 using cert_trans::CertChain;
 using cert_trans::PreCertChain;
+using cert_trans::UrlFetcher;
 using ct::DigitallySigned;
 using ct::MerkleAuditProof;
 using ct::SignedCertificateTimestamp;
@@ -24,12 +22,11 @@ using ct::SignedTreeHead;
 using std::back_inserter;
 using std::bind;
 using std::copy;
-using std::function;
 using std::make_shared;
-using std::ostringstream;
 using std::placeholders::_1;
 using std::shared_ptr;
 using std::string;
+using std::to_string;
 using std::unique_ptr;
 using std::vector;
 
@@ -46,19 +43,12 @@ string UriEncode(const string& input) {
 
 // Do some common checks, calls the callback with the appropriate
 // error if something is wrong.
-bool SanityCheck(const shared_ptr<libevent::HttpRequest>& req,
-                 const AsyncLogClient::Callback& done) {
-  if (!req) {
-    done(AsyncLogClient::UNKNOWN_ERROR);
-    return false;
-  }
-
-  if (req->GetResponseCode() < 1) {
-    done(AsyncLogClient::CONNECT_FAILED);
-    return false;
-  }
-
-  if (req->GetResponseCode() != HTTP_OK) {
+bool SanityCheck(UrlFetcher::Response* resp,
+                 const AsyncLogClient::Callback& done, util::Task* task) {
+  // TODO(pphaneuf): We should report errors better. The easiest way
+  // would be for this to use util::Task as well, so it could simply
+  // pass on the status.
+  if (!task->status().ok() || resp->status_code != HTTP_OK) {
     done(AsyncLogClient::UNKNOWN_ERROR);
     return false;
   }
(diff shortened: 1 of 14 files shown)


cpp/client/async_log_client.cc
```diff
@@ -67,12 +57,17 @@ bool SanityCheck(const shared_ptr<libevent::HttpRequest>& req,
...
-                SignedTreeHead* sth, const AsyncLogClient::Callback& done) {
-  if (!SanityCheck(req, done))
+void DoneGetSTH(UrlFetcher::Response* resp, SignedTreeHead* sth,
+                const AsyncLogClient::Callback& done, util::Task* task) {
+  unique_ptr<UrlFetcher::Response> resp_deleter(CHECK_NOTNULL(resp));
+  unique_ptr<util::Task> task_deleter(CHECK_NOTNULL(task));
+
+  LOG_IF(INFO, !task->status().ok()) << "DoneGetSTH: " << task->status();
+
+  if (!SanityCheck(resp, done, task))
```
AlCutter: Add `{...}` here and elsewhere

cpp/client/async_log_client.h
```diff
@@ -38,7 +38,12 @@ class AsyncLogClient {
 
   typedef std::function<void(Status)> Callback;
 
-  AsyncLogClient(const std::shared_ptr<libevent::Base>& base,
+  // The "executor" will be used to run callbacks.
+  // TODO(pphaneuf): Thee executor would not be necessary if we
```
AlCutter: /Thee/The/

cpp/net/url_fetcher.cc
```diff
@@ -0,0 +1,200 @@
...
+
+  map<string, string, ci_less<string>> headers;
+  for (evkeyval* ptr = evhttp_request_get_input_headers(req)->tqh_first; ptr;
+       ptr = ptr->next.tqe_next) {
+    map<string, string>::iterator it(headers.find(ptr->key));
+
+    if (it == headers.end()) {
+      headers.insert(make_pair(ptr->key, ptr->value));
+    } else {
+      it->second.append(", ");
```
AlCutter: What's the thinking behind this over a `std::multimap` ?

cpp/net/url_fetcher.cc
```diff
@@ -0,0 +1,200 @@
...
+  string body(reinterpret_cast<const char*>(evbuffer_pullup(
+                  evhttp_request_get_input_buffer(req), body_length)),
+              body_length);
+  state->response_->body.swap(body);
+
+  state->task_->Return();
+}
+
+
+UrlFetcher::Request AdjustRequest(UrlFetcher::Request req) {
```
AlCutter: `AdjustRequest` is a bit vague, how about `SanitiseRequest` (or something better you might be able to come up with)?

cpp/net/url_fetcher.cc
```diff
@@ -0,0 +1,200 @@
...
+        Status(util::error::INVALID_ARGUMENT,
+               "UrlFetcher: unsupported protocol: " + req.url.Protocol()));
+    return;
+  }
+
+  evhttp_add_header(evhttp_request_get_output_headers(http_req_), "Host",
+                    request_.url.Host().c_str());
+
+  // Remember to use the parameter here, because the copy we keep in
+  // request_ has the body stripped out to save space.
```
AlCutter: Nice comment, thanks!

cpp/net/url_fetcher.cc
pphaneuf: The ordering of fields with the same name is important, but I think a multimap would break that?

https://tools.ietf.org/html/rfc7230#section-3.2.2

I had forgotten about the damned Set-Cookie header, though, but my general intent was to keep it simple to use: need a header, look it up, and you're done.

I could change it to an ```std::multimap```.

cpp/net/url_fetcher.cc
pphaneuf: I've got upcoming things that are maybe more adjustments? Missing deadline set to a default value, for example? Or would you call that "sanitisation" as well?

cpp/net/url_fetcher.cc
pphaneuf: It might go away soon, I just realised that I'll need to keep the body, in case we need to retry the request. There's a way to add data to an ```evbuffer``` without it taking ownership, so I'll use that instead, and keep it in the ```std::string```.

cpp/net/url_fetcher.cc
AlCutter: Since C++11 (<3 C++11) the order of values associated with the same key into a multimap is order of insertion (http://en.cppreference.com/w/cpp/container/multimap), so I think you're ok from that PoV, but I'll leave it up to you to decide.

cpp/net/url_fetcher.cc
AlCutter: Normalise?
It's just `AdjustRequest` sounds like you're potentially changing it in a more-that-superficial fashion.

cpp/net/url_fetcher.cc
pphaneuf: I :heart: C++11 too. :smiley:

pphaneuf:
Comments addressed.

AlCutter:
LGTM

PR #497, opened 2015-02-13
Hex encode keys for unsequenced entries.

AlCutter:
Because the regular base64 charset includes "/" which will lead to interesting times, plus hex is not entirely unprecedented for representing hashes...


pphaneuf:
The little details... :wink: 

LGTM

PR #501, opened 2015-02-13
Add an operator<< for UrlFetcher::Response.

pphaneuf:
Also clean up some unnecessary ```using``` statements that were left behind.


AlCutter:
LGTM

pphaneuf:
Tweaking the way the body is output a bit before pushing, if you don't like it, can revisit.

PR #502, opened 2015-02-16
Fix a few issues with libevent::Base::Delay.

pphaneuf:
If we delay the cancellation callback (by using ```Base::Add```), it could race with the real completion, and the task could go into the ```DONE``` state, which means that both ```timer``` and ```task``` itself could be invalid by the time ```DelayCancel``` would get called! But the cancellation callback is stated to always complete before the task is allowed to go into the ```DONE``` state, and ```event_del``` is actually okay to call on other threads (as long as you haven't done ```event_free``` first, of course!).

So not only is bouncing the call to ```event_del``` to the libevent thread unnecessary, but doing so might cause a bug. :smiley: 

FYI, if we still needed to do the ```event_del``` on the libevent thread, that could still be achieved, by having the cancellation callback add a hold on the task before scheduling the call to ```event_del``` (in a function that would release the hold, of course), to prevent the task from going into the ```DONE``` state.

### DIFF
--- a/cpp/util/libevent_wrapper.cc
+++ b/cpp/util/libevent_wrapper.cc
@@ -111,8 +111,8 @@ void Base::Add(const function<void()>& cb) {
 
 
 void Base::Delay(const duration<double>& delay, util::Task* task) {
-  // If the delay is zero, what the heck, we're done!
-  if (delay == duration<double>::zero()) {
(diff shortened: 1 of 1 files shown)


AlCutter:
LGTM

PR #504, opened 2015-02-16
Add some more VLOG messages.

pphaneuf:


pphaneuf:
This is the first in a sequence of (at least) six, that ends at ```fetcher_libevent_thread```, if you want to look at where this is going...

AlCutter:
LGTM

PR #505, opened 2015-02-16
When a key exist, but there's no connection, make a new one.

pphaneuf:
Oops!


cpp/net/connection_pool.cc
```diff
@@ -26,7 +26,7 @@ evhttp_connection_unique_ptr ConnectionPool::Get(const URL& url) {
   lock_guard<mutex> lock(lock_);
 
   auto it(conns_.find(key));
-  if (it == conns_.end()) {
+  if (it == conns_.end() || it->second.empty()) {
```
AlCutter: :)

AlCutter:
LGTM

PR #506, opened 2015-02-17
Refactor the URL fetcher request callback a bit.

pphaneuf:
No functional change, this is only to make later pull requests clearer.


AlCutter:
LGTM

PR #507, opened 2015-02-17
Python: Move tls_message to its own package

eranmes:
The tls_message will be useful in ct/crypto later on, so to make
the dependency structure a bit clearer the tls_message package has
been moved.


AlCutter:
L"as-good-as-it-gets"TM

AlCutter:
Travis is having a bad time, but doesn't look like it's your fault.

PR #508, opened 2015-02-17
Keep the request body in the request.

pphaneuf:
Keeping the body around is necessary to be able to retry the request.


AlCutter:
LGTM

PR #509, opened 2015-02-17
No need to keep the evhttp_request* around.

pphaneuf:


cpp/net/url_fetcher.cc
```diff
@@ -67,29 +65,30 @@ UrlFetcher::Request NormaliseRequest(UrlFetcher::Request req) {
...
-                        "UrlFetcher: unsupported protocol: " +
-                            request_.url.Protocol()));
+    task_->Return(Status(util::error::INVALID_ARGUMENT,
+                         "UrlFetcher: unsupported protocol: " +
+                             request_.url.Protocol()));
     return;
   }
 
-  evhttp_add_header(evhttp_request_get_output_headers(http_req_), "Host",
+  evhttp_request* const req(
```
AlCutter: nit: keep this as `http_req` (or `ev_req`, or ...) to provide a little more distinguishability between this and `request`?

AlCutter:
LGTM with the slight name change.

PR #511, opened 2015-02-17
Make the request from the libevent dispatch thread.

pphaneuf:


cpp/net/url_fetcher.cc
```diff
@@ -186,8 +198,10 @@ void UrlFetcher::Post(const Request& req, Response* resp, Task* task) {
   TaskHold hold(task);
 
   State* const state(
-      new State(&impl_->pool_, EVHTTP_REQ_GET, req, resp, task));
+      new State(&impl_->pool_, EVHTTP_REQ_POST, req, resp, task));
```
pphaneuf: Oops! :smiley:

cpp/net/url_fetcher.cc
```diff
@@ -40,9 +42,13 @@ struct State {
     CHECK(!conn_) << "request state object still had a connection at cleanup?";
   }
 
+  // The following methods are only called on the libevent dispatch
```
AlCutter: /are only/should (must?) only be/ ?

AlCutter: Also, there's `Base::OnEventThread()` if you want to ensure it's so (or just log if it isn't).

pphaneuf: Updated the comment and used ```Base::OnEventThread()```, as suggested.

AlCutter:
LGTM

PR #513, opened 2015-02-18
Use gmock for etcd_test.

pphaneuf:


pphaneuf:
This allows catching missing calls (which I've been experiencing in my refactoring of the etcd client).

AlCutter:
Nice one.
LGTM

PR #515, opened 2015-02-19
Use a Notification to make sure we get the callback.

pphaneuf:
The current version of ```EtcdClient``` is synchronous (sometimes to disastrous effect!), but I'm fixing that too. :wink:


AlCutter:
LGTM

PR #516, opened 2015-02-19
Use util::Task in EtcdClient (internally).

pphaneuf:


cpp/util/etcd.cc
```diff
@@ -155,36 +163,44 @@ void GetRequestDone(Status status, const shared_ptr<JsonObject>& json,
        EtcdClient::Node::InvalidNode(), -1);
     return;
   }
-  cb(status, EtcdClient::Node(createdIndex.Value(), modifiedIndex.Value(),
-                              key.Value(), value.Value()),
-     etcd_index);
+
+  cb(task->status(),
```
AlCutter: Could be `Status::OK` to be clear when reading the code (`GetAllRequestDone` below does that)

AlCutter:
LGTM

PR #518, opened 2015-02-19
Improve URL and UrlFetcher::Response a bit.

pphaneuf:


pphaneuf:
I do YAGNI, but when I need it, I make it. :wink:

AlCutter:
LGTM

PR #520, opened 2015-02-23
Clear the HTTP headers in a response before filling them in.

pphaneuf:
This caused a bug if you re-used an UrlFetcher::Response.


AlCutter:
LGTM

PR #522, opened 2015-02-24
A few improvements to UrlFetcher

pphaneuf:
Having the HTTP verb in the ```UrlFetcher::Request``` makes retries a lot easier, and also everything is a bit more general (no ```switch``` statement to call different ```Get```/```Post```/etc methods).

### DIFF
--- a/cpp/client/async_log_client.cc
+++ b/cpp/client/async_log_client.cc
@@ -354,9 +354,9 @@ AsyncLogClient::AsyncLogClient(util::Executor* const executor,
 
 void AsyncLogClient::GetSTH(SignedTreeHead* sth, const Callback& done) {
   UrlFetcher::Response* const resp(new UrlFetcher::Response);
-  fetcher_->Get(GetURL("get-sth"), resp,
-                new util::Task(bind(DoneGetSTH, resp, sth, done, _1),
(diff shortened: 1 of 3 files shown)


cpp/client/async_log_client.cc
```diff
@@ -464,12 +464,13 @@ void AsyncLogClient::InternalAddChain(const CertChain& cert_chain,
   jsend.Add("chain", jchain);
 
   UrlFetcher::Request req(GetURL(pre_cert ? "add-pre-chain" : "add-chain"));
+  req.verb = UrlFetcher::Verb::POST;
```
AlCutter: I wonder if `verb` should be passed to the c'tor...

pphaneuf: I wondered the same, but it seems to me that once you start using other verbs, you'll also want to set a body, maybe some headers, and so on... So I went toward keeping this closer to a "pure ```struct```" (even though it's not all that "pure"!).

AlCutter: Yeah, I take your point, just seems like the verb might be one of the easier ones to forget to set resulting in some head scratching.  But it's your call.

AlCutter:
LGTM

Issue #405, opened 2015-01-16
Add support for monitoring


AlCutter:

eranmes:
Where? For what?

pphaneuf:
We're using the issues here to remember things. First step is going to be "figure out what and where", I think. :smiley:

pphaneuf:
You've got some sort of support for monitoring... You're clearly on top of this, so assigning to you, but I'll let you decide if you're sufficiently done to close this one. :smiley:

AlCutter:
We have support for prometheus now, need to build on it a bit - e.g. pre-packaged config with alerts etc. but we'll open separate issues for those things.

PR #525, opened 2015-02-25
Run up prometheus in GKE too.

AlCutter:
Issue #405

### DIFF
--- a/cloud/prometheus/prometheus.conf
+++ b/cloud/prometheus/prometheus.conf
@@ -1,0 +1,33 @@
+# Global settings and defaults.
+global {
+  scrape_interval: "10s"
+  evaluation_interval: "10s"
+  labels {
+    label {
+      name: "service"
+      value: "ct"
+    }
+  }
+}
+
+# Monitor Prometheus itself.
+job {
+  # This job will be named "prometheus", so a job="prometheus" label will be
+  # added to all time series scraped from it.
+  name: "prometheus"
+  # Scrape this job every 15s, overriding the global default.
+  scrape_interval: "15s"
+  # Configure a group of static HTTP targets
+  target_group {
+    target: "http://localhost:9090/metrics"
+  }
+}
+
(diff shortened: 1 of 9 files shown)


cloud/prometheus/prometheus.conf
```diff
@@ -0,0 +1,33 @@
+# Global settings and defaults.
+global {
+  scrape_interval: "10s"
+  evaluation_interval: "10s"
```
pphaneuf: That's pretty hot, no? Or am I just used to, er, a different scale? :smiley:

cloud/prometheus/prometheus.conf
```diff
@@ -0,0 +1,33 @@
...
+  labels {
+    label {
+      name: "service"
+      value: "super_duper"
```
pphaneuf: Shouldn't that be something more like "ct-server" or "ct_server"?

pphaneuf: It's not like you can monitor "master", anyway... :wink:

pphaneuf:
LGTM, my comments are possibly irrelevant, so I'll let you decide.

cloud/prometheus/prometheus.conf
AlCutter: Might be a bit hot for long-term production use, but kinda handy for now while it's barely working :)
We can certainly crank it down later on once it's all settled in.

pphaneuf: SGTM

cloud/prometheus/prometheus.conf
AlCutter: Well, it's the overall service [i.e. all the jobs get that label too], so and "...server" isn't quite right, but I've changed it to "ct" for now.

pphaneuf:
All right, fixed a few things I missed, too, looks even better than what I can tell. :wink:

PR #526, opened 2015-02-25
Convert EtcdClient to use UrlFetcher.

pphaneuf:


pphaneuf:
I added a commit to fix the build (the tests passed just fine once they were building again!).

cpp/util/etcd.cc
```diff
@@ -452,123 +437,74 @@ static const EtcdClient::Node kInvalidNode(-1, -1, "", "");
...
+      params.insert(make_pair("quorum", "true"));
+    } else {
+      LOG_EVERY_N(WARNING, 100) << "Sending request without 'quorum=true'";
     }
 
-    {
-      lock_guard<mutex> lock(lock_);
-      CHECK(!req_);
-      req_ = req;
+    req_.url.SetPath("/v2/keys" + key);
```
AlCutter: re-add the:
```
CHECK(!key.empty());
CHECK_EQ(key[0],'/');
```
here ?

cpp/util/etcd.cc
```diff
@@ -452,123 +437,74 @@ static const EtcdClient::Node kInvalidNode(-1, -1, "", "");
...
+    switch (req_.verb) {
+      case UrlFetcher::Verb::POST:
+      case UrlFetcher::Verb::PUT:
+        req_.headers.insert(
+            make_pair("Content-Type", "application/x-www-form-urlencoded"));
+        req_.body = UrlEscapeAndJoinParams(params);
+
```
AlCutter: `break;` ? :)

cpp/util/etcd.cc
```diff
@@ -452,123 +437,74 @@ static const EtcdClient::Node kInvalidNode(-1, -1, "", "");
...
+      default:
+        req_.url.SetQuery(UrlEscapeAndJoinParams(params));
     }
-    conn_->MakeRequest(req, verb_, uri.c_str());
+    req_.url.SetQuery(UrlEscapeAndJoinParams(params));
```
AlCutter: I guess this one should be deleted?

cpp/util/etcd.cc
pphaneuf: Done, albeit a bit higher.

cpp/util/etcd.cc
pphaneuf: Done. :smiley:

cpp/util/etcd.cc

cpp/util/etcd.cc
```diff
@@ -689,8 +615,8 @@ void EtcdClient::WatchState::RequestDone(GenericResponse* gen_resp,
   // helper function.
   {
     // This is probably due to a timeout, just retry.
-    if (!task->status().ok()) {
-      LOG(INFO) << "Watch request failed: " << task->status();
+    if (!child_task->status().ok()) {
+      LOG(INFO) << "Watch request failed: " << child_task->status();
```
pphaneuf: I made this one a bit less chatty (with ```LOG_EVERY_N```), since it happens on every timeout.

pphaneuf: (but I think it could still be useful, as it's a kind of catch-all)

cpp/util/etcd.cc
```diff
@@ -579,17 +515,17 @@ EtcdClient::WatchUpdate::WatchUpdate()
...
+  WatchInitialGetAllDone(state, status, {node}, etcd_index);
 }
 
 
-void EtcdClient::WatchState::InitialGetAllDone(Status status,
-                                               const vector<Node>& nodes,
-                                               int64_t etcd_index) {
-  if (task_->CancelRequested()) {
-    task_->Return(Status::CANCELLED);
+void EtcdClient::WatchInitialGetAllDone(WatchState* state, util::Status status,
```
AlCutter: Can this (and `WatchInitialGetDone`) be `const` ?

cpp/util/etcd.cc
```diff
@@ -669,18 +592,21 @@ void EtcdClient::WatchState::RequestDone(GenericResponse* gen_resp,
 
   VLOG(2) << "etcd_index: " << gen_resp->etcd_index;
 
-  // TODO(alcutter): doing this here works around some etcd 401 errors, but in
-  // the case of sustained high qps we could miss updates entirely (e.g. if
-  // this new index is already past the 1000 entry horizon by the time we make
-  // the new watch request.)  One way to address this might be to have the
-  // watcher re-do an "initial" get on the target, and, in the case of
-  // directory watches, maintain a set of known keys so that it can synthesise
+  // TODO(alcutter): doing this here works around some etcd 401
```
AlCutter: Did clang-format get all aggressive on wrapping this comment? looks like there's loads of space left on these lines !

cpp/util/etcd.cc
pphaneuf: No, because it makes their ```this``` ```const```, and they need to ```bind``` or call non-```const``` methods...

cpp/util/etcd.cc
```diff
@@ -803,101 +737,81 @@ EtcdClient::~EtcdClient() {
...
-  CHECK(uri);
-  LOG(INFO) << "etcd leader: " << evhttp_uri_get_host(uri.get()) << ":"
-            << evhttp_uri_get_port(uri.get());
-
-  // Update the last known leader, and retry the request on the new
-  // leader.
-  etcd_req->Run(UpdateLeader(evhttp_uri_get_host(uri.get()),
-                             evhttp_uri_get_port(uri.get())));
+  if (!task->status().ok()) {
+    // TODO(pphaneuf): If there is a connection problem, we should
```
AlCutter: same weird early wrapped comment here?

cpp/util/etcd.cc
```diff
@@ -803,101 +737,81 @@ EtcdClient::~EtcdClient() {
...
+                 "etcd returned a redirect without a Location header?"));
+      return;
+    }
 
-void EtcdClient::RequestDone(const shared_ptr<libevent::HttpRequest>& req,
-                             Request* etcd_req) {
-  // The HttpRequest object will be invalid as soon as we return, so
-  // forget about it now. It's too late to cancel, anyway.
-  etcd_req->Reset();
+    URL url(it->second);
```
AlCutter: constable?

cpp/util/etcd.cc
```diff
@@ -803,101 +737,81 @@ EtcdClient::~EtcdClient() {
...
+    if (url.Host().empty() || url.Port() == 0) {
+      etcd_req->task_->Return(
+          Status(util::error::INTERNAL,
+                 "could not parse Location header from etcd: " + it->second));
```
AlCutter: /c/C/

cpp/util/etcd.cc
pphaneuf: It was emacs being a bit aggressive, I've let clang-format rewrap it, and it makes it a no-op.

cpp/util/etcd.cc
pphaneuf: I did the same thing, but it's a bit far in the diff, we'll see how that goes...

cpp/util/etcd.cc
pphaneuf: Indeed!

cpp/util/etcd.cc
AlCutter: Ah, yes, I missed the bind to a non-const.

cpp/util/etcd.cc
pphaneuf: We haven't been totally consistent, but as those strings aren't proper sentence (and can be put inside of other messages, etc), I prefer to keep them un-capitalised (sort of following internal Google style).

AlCutter: Oke doke, Go has a similar recommendation for error message etc., so I can see where you're coming from.

AlCutter:
LGTM

PR #527, opened 2015-02-25
Update highest_index_seen_ from the nodes received.

pphaneuf:
It appears etcd sometimes sends a slightly out of date index in the ```X-Etcd-Index``` header? This causes double notifications, which isn't the end of the world, but are easy to fix.

### DIFF
--- a/cpp/util/etcd.cc
+++ b/cpp/util/etcd.cc
@@ -541,6 +541,10 @@ void EtcdClient::WatchInitialGetAllDone(WatchState* state, util::Status status,
 
   vector<WatchUpdate> updates;
   for (const auto& node : nodes) {
+    LOG_IF(WARNING, etcd_index < node.modified_index_)
+        << "X-Etcd-Index (" << etcd_index
+        << ") smaller than node modifiedIndex (" << node.modified_index_
+        << ") for key \"" << node.key_ << "\"";
     updates.emplace_back(WatchUpdate(node, true /*exists*/));
   }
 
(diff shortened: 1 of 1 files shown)


cpp/util/etcd.cc
```diff
@@ -541,6 +541,10 @@ void EtcdClient::WatchInitialGetAllDone(WatchState* state, util::Status status,
 
   vector<WatchUpdate> updates;
   for (const auto& node : nodes) {
+    LOG_IF(INFO, etcd_index < node.modified_index_)
+        << "X-Etcd-Index (" << etcd_index
+        << ") smaller than node modifiedIndex (" << node.modified_index
+        << ") for key \"" << node.key_ << "\"";
```
pphaneuf: I thought of adding a ```CHECK``` here, when I was reviewing the code that touches ```highest_index_seen_```, but figured this would just cause crashes when we could continue just fine.

Maybe I could make this a ```WARNING```, though? Considering it really shouldn't happen, and I was about to make the server crash instead of logging, maybe a flush to disk isn't the worst thing? :wink:

cpp/util/etcd.cc
```diff
@@ -541,6 +541,10 @@ void EtcdClient::WatchInitialGetAllDone(WatchState* state, util::Status status,
 
   vector<WatchUpdate> updates;
   for (const auto& node : nodes) {
+    LOG_IF(INFO, etcd_index < node.modified_index_)
+        << "X-Etcd-Index (" << etcd_index
+        << ") smaller than node modifiedIndex (" << node.modified_index_
+        << ") for key \"" << node.key_ << "\"";
```
pphaneuf: I thought of adding a ```CHECK``` here, when I was reviewing the code that touches ```highest_index_seen_```, but figured this would just cause crashes when we could continue just fine.

Maybe I could make this a ```WARNING```, though? Considering it really shouldn't happen, and I was about to make the server crash instead of logging, maybe a flush to disk isn't the worst thing? :wink:

AlCutter: yeah, WARNING might be ok; we don't watch things which change super frequently it shouldn't be too much of a hit anyway.

AlCutter:
LGTM

PR #529, opened 2015-02-26
Keep track of known etcd keys, to recover an outdated watch.

pphaneuf:


cpp/util/etcd.cc
```diff
@@ -597,24 +624,36 @@ void EtcdClient::WatchRequestDone(WatchState* state, GenericResponse* gen_resp,
...
+      Get(state->key_,
+          bind(&EtcdClient::WatchInitialGetDone, this, state, _1, _2, _3));
+    }
+
+    return;
+  }
+
+  if (child_task->status().ok()) {
+    VLOG(2) << "X-Etcd-Index: " << gen_resp->etcd_index;
+    CHECK_LE(state->highest_index_seen_, gen_resp->etcd_index);
```
AlCutter: We've seen etcd give us a crazy low `X-Etcd-Index` when the real is much higher (in fact, more than the watch window of 1000 entries higher!), is it worth changing this to a LOG(WARNING) ?

pphaneuf: Yeah, I've made this a bit more resilient...

AlCutter:
LGTM

PR #533, opened 2015-02-27
Pure-Go Merkle Tree implementation

mpalmer:
This is just my first draft interface specification.  Actual implementation will, naturally, be a little longer...

I'm keen to get feedback on whether the interface as a whole is "Go-ish", and whether it fits with the larger style of the project.  Commentary on the documentation is also greatly appreciated.

### DIFF
--- a/go/merkletree_new/merkle_tree_interface.go
+++ b/go/merkletree_new/merkle_tree_interface.go
@@ -1,0 +1,65 @@
+package merkletree
+
+// MerkleTreeInterface is the methods which are required to manipulate the
+// contents of a Merkle Hash Tree (that is, add new entries), and to query
+// the tree for inclusion and transparency proofs.
+type MerkleTreeInterface interface {
+	// CurrentRootHash returns the hash of the entire current tree.
+	CurrentRootHash func () ([]byte, error)
+
+	// InclusionProof returns a list of the hashes of the "sibling"
+	// nodes to the leaf entry located at the (0-based) index specified.
+	// The returned list is presented in order, starting from the
+	// sibling to the item itself, and ending with a hash that is an
+	// immediate child of the root.
+	InclusionProof func (entry uint64) ([]byte, error)
+
+	// ConsistencyProof returns a list of the hashes that make up the
+	// consistency proof between the hash trees with size |tree1| and
+	// |tree2|.
+	ConsistencyProof func (sz uint64) ([]byte, error)
+}
+
+// MerkleTreeDataInterface defines the means by which a merkle hash tree can
+// retrieve data from an arbitrary data store, without tying the tree
+// implementation to any particular backend storage technology.
+type MerkleTreeDataInterface interface {
+	// EntryAt takes a (0-based) |index| into the list of leaf entries in
+	// the tree, and returns the contents of the leaf.  For a given
+	// |index| in a given tree, the value returned **MUST NOT** ever
+	// change, or the whole point of a merkle tree is defeated.
+	EntryAt func (index uint64) []byte
+
+	// Size returns the number of entries currently in the tree.  The
+	// value returned by this function must never decrease between
+	// calls; it is OK for the value to increase over time.
+	Size func () uint64
+}
+
+// MerkleTreeCacheInterface provides a means for a merkle tree to cache
+// hash values in the tree.  Whilst all hash values *can* be recalculated
+// on the fly, it is far more efficient to be able to cache recently-used
+// values, to avoid needing to rehash everything.
+type MerkleTreeCacheInterface interface {
+	GetNode func ([]byte) []byte
+	SetNode func ([]byte) []byte
+}
+
+type MerkleTree struct {
+	// If a merkle tree you wish to be, you must implement these
+	// functions three.
+	MerkleTreeInterface
+
+	dao   *MerkleTreeDataInterface
+	cache *MerkleTreeCacheInterface
+
+	hasher func ([]byte) []byte
+}
+
+// New creates a new merkle hash tree.  The number of "leaves" of the tree,
+// as well as their contents, are retrieved through |dao|.  If you wish to
+// have acceptable performance on non-trivial tree sizes, you'll want to
+// provide |cache| (otherwise, pass `nil`).  The hash function used for all
+// nodes in the tree is specified by |hasher|.
+func New(dao *MerkleTreeDataInterface, cache *MerkleTreeCacheInterface, hasher func ([]byte) []byte) *MerkleTree {
+}
--- a/go/merkletree_new/merkle_tree_interface.go
+++ b/go/merkletree_new/merkle_tree_interface.go
@@ -1,23 +1,25 @@
 package merkletree
 
-// MerkleTreeInterface is the methods which are required to manipulate the
-// contents of a Merkle Hash Tree (that is, add new entries), and to query
-// the tree for inclusion and transparency proofs.
+type Hash []byte
+
+// MerkleTreeInterface is the set of methods which are required to
+// manipulate the contents of a Merkle Hash Tree (that is, add new entries),
+// and to query the tree for inclusion and transparency proofs.
 type MerkleTreeInterface interface {
 	// CurrentRootHash returns the hash of the entire current tree.
-	CurrentRootHash func () ([]byte, error)
+	CurrentRootHash() (Hash, error)
 
 	// InclusionProof returns a list of the hashes of the "sibling"
 	// nodes to the leaf entry located at the (0-based) index specified.
 	// The returned list is presented in order, starting from the
 	// sibling to the item itself, and ending with a hash that is an
 	// immediate child of the root.
-	InclusionProof func (entry uint64) ([]byte, error)
+	InclusionProof(entry uint64) ([]Hash, error)
 
 	// ConsistencyProof returns a list of the hashes that make up the
 	// consistency proof between the hash trees with size |tree1| and
 	// |tree2|.
-	ConsistencyProof func (sz uint64) ([]byte, error)
+	ConsistencyProof(tree1 uint64, tree2 uint64) ([]Hash, error)
 }
 
 // MerkleTreeDataInterface defines the means by which a merkle hash tree can
@@ -41,8 +43,8 @@ type MerkleTreeDataInterface interface {
 // on the fly, it is far more efficient to be able to cache recently-used
 // values, to avoid needing to rehash everything.
 type MerkleTreeCacheInterface interface {
-	GetNode func ([]byte) []byte
-	SetNode func ([]byte) []byte
+	GetNode func ([]byte) Hash
+	SetNode func ([]byte, Hash)
 }
 
 type MerkleTree struct {
--- a/go/new_merkletree/merkle_tree.go
+++ b/go/new_merkletree/merkle_tree.go
@@ -1,0 +1,177 @@
+package new_merkletree
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"hash"
+	"math"
+)
+
+// New creates a new merkle hash tree.  The number of "leaves" of the tree,
+// as well as their contents, are retrieved through |dao|.  If you wish to
+// have acceptable performance on non-trivial tree sizes, you'll want to
+// provide |cache| (otherwise, pass `nil`).  The hash function used for all
+// nodes in the tree is specified by |hasher|.
+func New(dao NewMerkleTreeDataInterface, cache NewMerkleTreeCacheInterface, hasher hash.Hash) *NewMerkleTree {
+	return &NewMerkleTree{dao: dao, cache: cache, hasher: hasher}
+}
+
+func (mt NewMerkleTree) CurrentRoot() (Hash, error) {
+	if mt.dao.Size() == 0 {
+		// Special case: empty trees get hashes of empty strings
+		mt.hasher.Reset()
+		return mt.hasher.Sum([]byte{}), nil
+	} else {
+		return mt.subtreeRoot(0, mt.dao.Size()-1)
+	}
+}
+
+func (mt NewMerkleTree) InclusionProof(leaf uint64) ([]Hash, error) {
+	if mt.dao.Size() == 0 {
+		return nil, errors.New("NewMerkleTree: Can't calculate an inclusion proof on an empty tree")
+	}
+
+	if leaf >= mt.dao.Size() {
+		return nil, fmt.Errorf("NewMerkleTree: Invalid leaf entry ID: %v", leaf)
+	}
+
+	return mt.inclusionSubtree(leaf, 0, mt.dao.Size()-1)
+}
+
+func (mt NewMerkleTree) ConsistencyProof(from, to uint64) ([]Hash, error) {
+	if from == 0 {
+		// There's no algorithmic basis for this that I know of, but it is how
+		// existing implementations do it
+		return []Hash{}, nil
+	}
+
+	if to > mt.dao.Size() {
+		return nil, fmt.Errorf("NewMerkleTree.ConsistencyProof: Value for 'to' greater than tree size (to=%v, tree size=%v)", to, mt.dao.Size())
+	}
+
+	if from > to {
+		return nil, fmt.Errorf("NewMerkleTree.ConsistencyProof: 'to' greater than 'from'")
+	}
+
+	return mt.subproof(from, 0, to-1, true)
+}
+
+func (mt NewMerkleTree) hash(s []byte) Hash {
+	mt.hasher.Reset()
+	mt.hasher.Write(s)
+	return mt.hasher.Sum([]byte{})
+}
+
+func (mt NewMerkleTree) leafHash(s []byte) Hash {
+	return mt.hash(bytes.Join([][]byte{{0x0}, s}, []byte{}))
+}
+
+func (mt NewMerkleTree) nodeHash(h1, h2 Hash) Hash {
+	return mt.hash(bytes.Join([][]byte{{0x1}, h1, h2}, []byte{}))
+}
+
+func (mt NewMerkleTree) subtreeRoot(n1, n2 uint64) (Hash, error) {
+	if n1 == n2 {
+		l, err := mt.dao.EntryAt(n1)
+		if err != nil {
+			return nil, err
+		}
+		return mt.leafHash(l), nil
+	} else {
+		pivot := largestPowerOfTwoLessThan(n2 - n1 + 1)
+
+		s1, err1 := mt.subtreeRoot(n1, n1+pivot-1)
+		if err1 != nil {
+			return nil, err1
+		}
+
+		s2, err2 := mt.subtreeRoot(n1+pivot, n2)
+		if err2 != nil {
+			return nil, err2
+		}
+
+		return mt.nodeHash(s1, s2), nil
+	}
+}
+
+func (mt NewMerkleTree) inclusionSubtree(leaf, n1, n2 uint64) ([]Hash, error) {
+	if n1 == n2 {
+		// Inclusion proof of a single element is the empty list
+		return []Hash{}, nil
+	} else {
+		pivot := largestPowerOfTwoLessThan(n2 - n1 + 1)
+
+		if leaf < pivot {
+			h, err := mt.subtreeRoot(n1+pivot, n2)
+			if err != nil {
+				return nil, err
+			}
+
+			sp, err := mt.inclusionSubtree(leaf, n1, n1+pivot-1)
+			if err != nil {
+				return nil, err
+			}
+
+			return append(sp, h), nil
+		} else {
+			h, err := mt.subtreeRoot(n1, n1+pivot-1)
+			if err != nil {
+				return nil, err
+			}
+
+			sp, err := mt.inclusionSubtree(n1+leaf-pivot, n1+pivot, n2)
+			if err != nil {
+				return nil, err
+			}
+
+			return append(sp, h), nil
+		}
+	}
+}
+
+func largestPowerOfTwoLessThan(n uint64) uint64 {
+	if n < 2 {
+		panic("Can't calculate tiny powers of two")
+	} else {
+		return uint64(math.Floor(math.Pow(2, math.Floor(math.Log2(float64(n)-1)))))
+	}
+}
+
+func (mt NewMerkleTree) subproof(from, t1, t2 uint64, b bool) ([]Hash, error) {
+	if t2 == from-1 {
+		if b {
+			return []Hash{}, nil
+		} else {
+			h, err := mt.subtreeRoot(t1, t2)
+
+			return []Hash{h}, err
+		}
+	} else if t1 == t2 {
+		h, err := mt.subtreeRoot(t1, t2)
+		return []Hash{h}, err
+	} else {
+		pivot := largestPowerOfTwoLessThan(t2 - t1 + 1)
+		var (
+			sp            []Hash
+			h             Hash
+			err_sp, err_h error
+		)
+
+		if from <= pivot+t1 {
+			sp, err_sp = mt.subproof(from, t1, t1+pivot-1, b)
+			h, err_h = mt.subtreeRoot(t1+pivot, t2)
+		} else {
+			sp, err_sp = mt.subproof(from, t1+pivot, t2, false)
+			h, err_h = mt.subtreeRoot(t1, t1+pivot-1)
+		}
+
+		if err_sp != nil {
+			return nil, err_sp
+		}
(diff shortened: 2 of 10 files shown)


pphaneuf:
Maybe @AlCutter should take a look, as the closest we have to a resident Go expert. :smiley: 

For my part, I don't think this should go in a new directory (it should match the package name, at any rate)? Maybe have a "New" prefix on the types, while you're building them, and we switch the existing code over to them, then we could do a single commit deleting the old code and renaming the types to drop the "New"?

Any idea what's going on with this on Travis?

```
go/merkletree_new/merkle_tree_interface.go:8: syntax error: unexpected func, expecting (
go/merkletree_new/merkle_tree_interface.go:15: non-declaration statement outside function body
go/merkletree_new/merkle_tree_interface.go:20: non-declaration statement outside function body
go/merkletree_new/merkle_tree_interface.go:31: syntax error: unexpected func, expecting (
go/merkletree_new/merkle_tree_interface.go:36: non-declaration statement outside function body
go/merkletree_new/merkle_tree_interface.go:44: syntax error: unexpected func, expecting (
go/merkletree_new/merkle_tree_interface.go:45: non-declaration statement outside function body
```

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
+package merkletree
+
+// MerkleTreeInterface is the methods which are required to manipulate the
```
AlCutter: /is the methods/is the set of methods/ ?

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
...
+// contents of a Merkle Hash Tree (that is, add new entries), and to query
+// the tree for inclusion and transparency proofs.
+type MerkleTreeInterface interface {
+	// CurrentRootHash returns the hash of the entire current tree.
+	CurrentRootHash func () ([]byte, error)
```
AlCutter: Should this (and elsewhere) be `CurrentRootHash() ([]byte, error)` ?

AlCutter: Might be worth having a `type Hash []byte` ?

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// InclusionProof returns a list of the hashes of the "sibling"
+	// nodes to the leaf entry located at the (0-based) index specified.
+	// The returned list is presented in order, starting from the
+	// sibling to the item itself, and ending with a hash that is an
+	// immediate child of the root.
+	InclusionProof func (entry uint64) ([]byte, error)
```
AlCutter: ... then this could be `InclusionProof(entry uint64) ([]Hash, error)`

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// ConsistencyProof returns a list of the hashes that make up the
+	// consistency proof between the hash trees with size |tree1| and
+	// |tree2|.
+	ConsistencyProof func (sz uint64) ([]byte, error)
```
AlCutter: The comment talks about `tree1` and `tree2` but the func only takes `sz`.
Is the intention that this would give you a consistency proof between *this* tree and one of a smaller size?  If so, how will we handle the case where a request comes in for a consistency proof between two historical trees?  In practice a larger tree can answer that (see the C++ merkle_tree code), but this interface would scupper that (unless you have a cunning plan for this already?)

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// MerkleTreeDataInterface defines the means by which a merkle hash tree can
+// retrieve data from an arbitrary data store, without tying the tree
+// implementation to any particular backend storage technology.
+type MerkleTreeDataInterface interface {
```
AlCutter: These two interfaces are an interesting idea, I'd only ever imagined the tree being all in-memory.
Will be interesting to see what the trade-off will be in terms of space efficiency, but this could be very nice for insane tree sizes.

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
...
+	// EntryAt takes a (0-based) |index| into the list of leaf entries in
+	// the tree, and returns the contents of the leaf.  For a given
+	// |index| in a given tree, the value returned **MUST NOT** ever
+	// change, or the whole point of a merkle tree is defeated.
+	EntryAt func (index uint64) []byte
```
AlCutter: `(Hash, error)`: someone could pass in an `index` > tree size

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
...
+	// calls; it is OK for the value to increase over time.
+	Size func () uint64
+}
+
+// MerkleTreeCacheInterface provides a means for a merkle tree to cache
+// hash values in the tree.  Whilst all hash values *can* be recalculated
+// on the fly, it is far more efficient to be able to cache recently-used
+// values, to avoid needing to rehash everything.
+type MerkleTreeCacheInterface interface {
+	GetNode func ([]byte) []byte
```
AlCutter: Not following how these are going to work, could you explain the params and return values a bit?

I can imagine something along the lines of
```
GetNode(level uint64, index uint64) (Hash, error)  
SetNode(level uint64, index uin64, hash Hash) (error)
```
So with:
``` 
      R
     /  \
    /     \
   /        \
  n10        n11
 /   \      /   \    
l0   l1    l2   l3
```
Where `R` is level 0, index 0, `n10` is level 1, index 0, and `n11` is level 1, index 1, and so on.

But perhaps there's a better way.

go/merkletree_new/merkle_tree_interface.go
```diff
@@ -0,0 +1,65 @@
...
+	SetNode func ([]byte) []byte
+}
+
+type MerkleTree struct {
+	// If a merkle tree you wish to be, you must implement these
+	// functions three.
+	MerkleTreeInterface
+
+	dao   *MerkleTreeDataInterface
+	cache *MerkleTreeCacheInterface
```
AlCutter: Could/should the cache 'wrap' the DataInterface so that there's just one path to getting node info?
i.e. if the cache doesn't have the requested data then it'll go and populate that cell before returning the value or does that not work for your plan?

mpalmer:
re: the CLA, the copyright on this work is actually held by Google, so yeah...

@pphaneuf, I didn't actually wave the compiler over this code, since it was entirely speculative.

go/merkletree_new/merkle_tree_interface.go
mpalmer: Whups.  Fixed.

go/merkletree_new/merkle_tree_interface.go
mpalmer: Yep, I can certainly see the value in having a separate `Hash` type.  The syntax issues are just my inexperience with Go shining through.

go/merkletree_new/merkle_tree_interface.go
mpalmer: That was just me making a copy-paste failure, I suspect.

go/merkletree_new/merkle_tree_interface.go
mpalmer: Yes, I don't think an all-memory tree scales very well, and I think it is important to at least provide for the possibility of not holding the entire tree, and all the leaf node data, in memory.

go/merkletree_new/merkle_tree_interface.go
mpalmer: Good catch.

go/merkletree_new/merkle_tree_interface.go
mpalmer: By making the key an opaque string, the cache interface can be kept very simple and generic.  The key structure for leaf node hashes is simply the entry number, while the keys for intermediate hashes is simply `"M..N"`, where `M` is the first leaf entry number and `N` is the last leaf entry number.  I find this way of reasoning about the "meaning" of an intermediate node to be much simpler than talking about levels and indexes.

go/merkletree_new/merkle_tree_interface.go
mpalmer: That would require the cache interface to know an awful lot about how to construct intermediate nodes -- essentially, half of the merkle tree implementation would end up inside the cache interface.  My desire is that the cache interface be "pluggable".  The first cache interface would be a simple in-memory LRU cache, however a shared cache (using, say, Redis or memcached) would be far more CPU efficient and allow new nodes to be immediately effective, rather than having to pre-calculate the entire tree.

mpalmer:
First draft of complete implementation done.  Feedback appreciated.

go/new_merkletree/merkle_tree.go
```diff
@@ -0,0 +1,177 @@
+package new_merkletree
```
AlCutter: Can you add a package comment here?

go/new_merkletree/merkle_tree.go
```diff
@@ -0,0 +1,177 @@
...
+func New(dao NewMerkleTreeDataInterface, cache NewMerkleTreeCacheInterface, hasher hash.Hash) *NewMerkleTree {
+	return &NewMerkleTree{dao: dao, cache: cache, hasher: hasher}
+}
+
+func (mt NewMerkleTree) CurrentRoot() (Hash, error) {
+	if mt.dao.Size() == 0 {
+		// Special case: empty trees get hashes of empty strings
+		mt.hasher.Reset()
+		return mt.hasher.Sum([]byte{}), nil
+	} else {
```
AlCutter: No need for an `else` block here, just have the next return directly below the `if` block.

go/new_merkletree/merkle_tree.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func (mt NewMerkleTree) InclusionProof(leaf uint64) ([]Hash, error) {
+	if mt.dao.Size() == 0 {
+		return nil, errors.New("NewMerkleTree: Can't calculate an inclusion proof on an empty tree")
+	}
+
+	if leaf >= mt.dao.Size() {
+		return nil, fmt.Errorf("NewMerkleTree: Invalid leaf entry ID: %v", leaf)
```
AlCutter: /entry ID/index/ ?

go/new_merkletree/merkle_tree.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	return mt.inclusionSubtree(leaf, 0, mt.dao.Size()-1)
+}
+
+func (mt NewMerkleTree) ConsistencyProof(from, to uint64) ([]Hash, error) {
+	if from == 0 {
```
AlCutter: This might be a good candidate for Go's switch statement:
```
switch {
       from == 0:
               ...
               return []Hash{}, nil
        to > mt.dao.Size():
               ...
               return ...
        from > to:
               ...
}
return mt.subproof(...)

File: go/new_merkletree/merkle_tree.go [hunk 13]
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (mt NewMerkleTree) subtreeRoot(n1, n2 uint64) (Hash, error) {
+	if n1 == n2 {
+		l, err := mt.dao.EntryAt(n1)
+		if err != nil {
+			return nil, err
+		}
+		return mt.leafHash(l), nil
+	} else {
```
AlCutter: no need for the `else` block (and elsewhere, I'll stop commenting on them individually)

go/new_merkletree/merkle_tree.go
```diff
@@ -0,0 +1,177 @@
...
+			return append(sp, h), nil
+		}
+	}
+}
+
+func largestPowerOfTwoLessThan(n uint64) uint64 {
+	if n < 2 {
+		panic("Can't calculate tiny powers of two")
+	} else {
+		return uint64(math.Floor(math.Pow(2, math.Floor(math.Log2(float64(n)-1)))))
```
AlCutter: Rather than messing with all the floats, could you do something like:
```
  p := uint64(1)
  for ; n > 1; n >>= 1 {
    p <<= 1
  }
  return p
```
?

You may need to adjust slightly depending on what you want returning when `n` is itself a power of two; right now you'll get `n` back, if you don't want that then you can probably just pre-decrement it before the loop.

go/new_merkletree/merkle_tree.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func (mt NewMerkleTree) subproof(from, t1, t2 uint64, b bool) ([]Hash, error) {
+	if t2 == from-1 {
```
AlCutter: Use Go's `switch` ?

go/new_merkletree/merkle_tree_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (h *NullHash) BlockSize() int {
+	return 1
+}
+
+// Real testing methods start here.
+
+func TestNew(t *testing.T) {
+	var tree *NewMerkleTree
+	tree = New(nil, nil, sha256.New())
```
AlCutter: Can you write this more concisely as:
```
if tree := New(nil, nil, sha256.New()); tree == nil {
  t.Fail()
}
```

?

go/new_merkletree/merkle_tree_test.go
```diff
@@ -0,0 +1,228 @@
...
+	if tree == nil {
+		t.Fail()
+	}
+}
+
+func TestEmptyTree(t *testing.T) {
+	var tree *NewMerkleTree
+	// Let's use sha256 for this one, just to show we can
+	tree = New(NewDummyDAO(0), nil, sha256.New())
```
AlCutter: `tree := New...` and ding the `var ...` above?

go/new_merkletree/merkle_tree_test.go
```diff
@@ -0,0 +1,228 @@
...
+	case 5:
+		return []byte("\x01\x01\x01\x00A\x00B\x01\x00C\x00D\x00E")
+	case 6:
+		return []byte("\x01\x01\x01\x00A\x00B\x01\x00C\x00D\x01\x00E\x00F")
+	case 7:
+		return []byte("\x01\x01\x01\x00A\x00B\x01\x00C\x00D\x01\x01\x00E\x00F\x00G")
+	default:
+		log.Fatalf("Unexpected numLeaves: %v", numLeaves)
+	}
+	return nil
```
AlCutter: do you need this final `return`?

go/new_merkletree/merkle_tree.go
mpalmer: Yes... yes I can.

go/new_merkletree/merkle_tree.go
mpalmer: I blame my Ruby upbringing... implicit returns, and all that fun stuff.

go/new_merkletree/merkle_tree.go
mpalmer: Indeed.

go/new_merkletree/merkle_tree.go
mpalmer: Done.

go/new_merkletree/merkle_tree.go
mpalmer: The casts do make for an extremely ugly piece of code, however I avoided the iterative shift implementation out of efficiency concerns (this function is called quite often).  I'll benchmark the two implementations, and if there's not much difference (I guess there's the possibility that the loop form might be faster, if floating-point operations aren't real quick) I'll switch over.

go/new_merkletree/merkle_tree.go
mpalmer: This is a bit less cut-and-dried than the other suggested use of `switch`; I've changed it over, but I'm not ecstatic about the readability difference.  Large chunks of code in a switch statement "feel" less readable to me.  However, if it's more idiomatic Go, then I assume it's worth it for everyone else's benefit.

go/new_merkletree/merkle_tree_test.go
mpalmer: I've removed one extraneous line, and added a comment about why the var is there.

go/new_merkletree/merkle_tree_test.go
mpalmer: Done.

go/new_merkletree/merkle_tree_test.go
mpalmer: Yes, although it doesn't have to be hanging out the bottom all on its own.

mpalmer:
New diff on its way.

go/new_merkletree/merkle_tree.go
AlCutter: Did you come up with any conclusions from the benchmark?

go/new_merkletree/merkle_tree_interface.go
```diff
@@ -0,0 +1,61 @@
+package new_merkletree
+
+import "hash"
+
+type Hash []byte
+
+// NewMerkleTreeInterface is the set of methods which are required to
+// manipulate the contents of a Merkle Hash Tree (that is, add new entries),
+// and to query the tree for inclusion and transparency proofs.
+type NewMerkleTreeInterface interface {
```
ThomasHabets: Are we expecting more implementations to satisfy this interface? If not, skip it.

go/new_merkletree/merkle_tree_interface.go
```diff
@@ -0,0 +1,61 @@
...
+// NewMerkleTreeCacheInterface provides a means for a merkle tree to cache
+// hash values in the tree.  Whilst all hash values *can* be recalculated
+// on the fly, it is far more efficient to be able to cache recently-used
+// values, to avoid needing to rehash everything.
+type NewMerkleTreeCacheInterface interface {
+	GetNode([]byte) Hash
+	SetNode([]byte, Hash)
+}
+
+type NewMerkleTree struct {
```
ThomasHabets: This is not part of the interface, but the specific implementation.

go/new_merkletree/merkle_tree.go
```diff
@@ -0,0 +1,173 @@
...
+// New creates a new merkle hash tree.  The number of "leaves" of the tree,
+// as well as their contents, are retrieved through |dao|.  If you wish to
+// have acceptable performance on non-trivial tree sizes, you'll want to
+// provide |cache| (otherwise, pass `nil`).  The hash function used for all
+// nodes in the tree is specified by |hasher|.
+func New(dao NewMerkleTreeDataInterface, cache NewMerkleTreeCacheInterface, hasher hash.Hash) *NewMerkleTree {
+	return &NewMerkleTree{dao: dao, cache: cache, hasher: hasher}
+}
+
+func (mt NewMerkleTree) CurrentRoot() (Hash, error) {
```
ThomasHabets: Looks like these should all have a pointer receiver  (func (mt *MerkleTree)…)

go/new_merkletree/merkle_tree_interface.go
```diff
@@ -0,0 +1,61 @@
...
+	// If a merkle tree you wish to be, you must implement these
+	// functions three.
+	NewMerkleTreeInterface
+
+	dao   NewMerkleTreeDataInterface
+	cache NewMerkleTreeCacheInterface
+
+	hasher hash.Hash
```
ThomasHabets: There's no reason to keep this state, right? So how about "hasher func() hash.Hash" ?

go/new_merkletree/merkle_tree_interface.go
mpalmer: OK.

go/new_merkletree/merkle_tree_interface.go
mpalmer: I assume you're referring specifically to the `NewMerkleTree` struct there?  (Github doesn't appear to indicate which line(s) you're commenting on, specifically).  If so, I'm not sure what you're objecting to.  Is it the placement of the type within the `_interface.go` file, or something more fundamental, like its being exported?

go/new_merkletree/merkle_tree.go
mpalmer: Updated.  I'm still getting my head around exactly when pointer receivers should be used, so thanks for the, uh, pointer.  (grin)

go/new_merkletree/merkle_tree_interface.go
mpalmer: Ooooh, thanks for that.  I originally tried to take the New method as part of the constructor, but I couldn't quite work out how to do it, so I took the lazy way out.  This way is *far* cleaner.

go/new_merkletree/merkle_tree.go
mpalmer: Oops, forgot to comment here.  Turns out the iterative shifting is a lot quicker.  Floating point arithmetic is slooooooooow, apparently.  New diff on its way.

go/new_merkletree/merkle_tree_interface.go
ThomasHabets: On Tue, 12 May 2015 20:18:42 -0700, Matt Palmer <notifications@github.com> said:
> > +type NewMerkleTree struct {
>
> I assume you're referring specifically to the `NewMerkleTree` struct there?
> (Github doesn't appear to indicate which line(s) you're commenting on,
> specifically).  If so, I'm not sure what you're objecting to.  Is it the
> placement of the type within the `_interface.go` file, or something more
> fundamental, like its being exported?

Huh, yeah github doesn't seem to be awesome in code reviewing, to say
the least. :-(

I'm referring to the specific implementation being in the interface go
file. But I think in a separate comment I was asking why it has an
interface at all.

mpalmer: OK, I've fixed that up in the latest version, removing the `NewMerkleTreeInterface` and moving the `NewMerkleTree` struct into `merkle_tree.go`.

AlCutter:
On the module name front, I'd quite like to keep the C++ peer tree around for a little bit longer; it'll give us the ability to A:B test for correctness & performance etc., but we could move that to some other package name and have yours be called merkletree.

AlCutter:
BTW, the C++ `merkle_tree_test.cc` file has got a whole bunch of known-answer tests, it'd be really nice to transcribe those over into your `merkle_tree_test.go` file.

AlCutter:
The C++ test also has a small "reference" implementation which it uses to compare against the "real" implementation, I wonder if we could a similar thing here using the C++ peered tree as the reference?

mpalmer:
@AlCutter, I might stick with calling the *package* `gomerkletree` for now, but drop any "new" stuff from the type and function names.  I ported over the tests from the Go wrapper code previously; I'll take a look at the C++ tests and wrestle them in as well.  I'm not sure what you mean by your last comment; are you thinking of keeping the C++ reference implementation around to generate trees, heads, and proofs for use in test cases?

AlCutter:
@mpalmer, oke doke - the naming sgtm.
On the reference subject - check out the "FUZZ TESTS AGAINST REFERENCE IMPLEMENTATIONS" section in `merkle_tree_test.cc`, we could do something similar in `merkle_tree_test.go`, but using the  C++ merkle tree as the reference IYSWIM?

mpalmer:
ISWYM.

pphaneuf:
On fuzzing, this has been bouncing around recently, might be worth a look?

https://github.com/dvyukov/go-fuzz

mpalmer:
@AlCutter, I've enhanced the test suite with a bunch of extra tests (turns out I *hadn't* ported the tests from the existing Go merkletree module, I just intended to...), so there's now more realistic test coverage, based on existing tests.

I'm not sure what to do with the C++ reference implementation and the fuzzing.  Should I extract the C++ reference implementation, write a Go module around it, and then do fuzz testing against them both side-by-side?

AlCutter:
Ah, I was thinking more of using the existing Go-wrapped-C++-merkletree as the reference for the fuzzing against your new pure Go tree, based on the assumption that the existing-go-wrapped-C++-merkletree is known working (and of course is built using the C++ tree which is already fuzz tested against the C++ "reference" implementation methods in the C++ test.)

mpalmer:
Dear Googlebot,

I don't have copyright over the contents of this PR, Google does.  So harassing me (*repeatedly*) about signing a CLA that *I can't legally agree to* isn't going to get you very far.

RJPercival:
There is now a pure-Go Merkle tree implementation here: https://github.com/google/trillian/blob/v1.2.1/merkle/memory_merkle_tree.go

PR #534, opened 2015-02-27
Fix a couple of sequencer issues

AlCutter:


pphaneuf:
Any background on why returning entries in sequenced order is helpful?

AlCutter:
Well, it seems like `GetSequencedEntries()` should really return them in sequence order, but the way I discovered that it wasn't doing that was that it caused `NextAvailableSequenceNumber()` to return something way too low (because it just checks the last entry in the returned list.)  I could've fixed it there too by making that O(N) and scanning.

pphaneuf:
LGTM

PR #536, opened 2015-02-27
Do not crash when a peer group doesn't have enough entries.

pphaneuf:
This could happen if we lost the peer that had the most entries during a fetch.


AlCutter:
LGTM

PR #537, opened 2015-02-27
Use the database contiguous tree size for get-entries.

pphaneuf:
Also use ```int64_t``` instead of ```int``` for indices more comprehensively.


AlCutter:
LGTM

PR #538, opened 2015-02-27
Convert EtcdClient::Delete to use util::Task.

pphaneuf:
I'll do them one at a time, so it's not a big blob of a pull request, and so that I can make progress in-between other things. :wink:


AlCutter:
LGTM

PR #540, opened 2015-03-02
Clean up old entries in etcd on request only.

AlCutter:


Dockerfile
```diff
@@ -36,7 +36,8 @@ CMD cd /mnt/ctlog/ && \
         --trusted_cert_file=/usr/local/etc/ctlog_ca_roots.pem \
         --stderrthreshold=0 \
         --log_dir=/mnt/ctlog/logs \
-        --tree_signing_frequency_seconds=60 \
+        --tree_signing_frequency_seconds=30 \
+        --guard_window_seconds=10 \
```
pphaneuf: That's for testing only, no? Sure you want to push this, especially in this PR (that looks like this hasn't got much to do with it)?

AlCutter: Yeah, I'll pull it out into a separate PR though and merge it in anyway, then we can remove it once things have settled down a bit.

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -625,14 +619,18 @@ void EtcdConsistentStore<Logged>::OnEtcdServingSTHUpdated(
...
-util::Status EtcdConsistentStore<Logged>::RunOneCleanUpIteration(
-    const int64_t clean_up_to_sequence_number) {
-  CHECK_LE(0, clean_up_to_sequence_number);
+util::Status EtcdConsistentStore<Logged>::CleanupOldEntries() {
   if (!election_->IsMaster()) {
     return util::Status(util::error::PERMISSION_DENIED,
                         "Non-master node cannot run cleanups.");
   }
 
+  if (!serving_sth_) {
```
pphaneuf: Here and at line 641, you're accessing ```serving_sth_``` without holding ```mutex_```? Isn't that wrong? Could probably deal with this by calling ```GetServingSTH```, which would give you a copy (or tell you that there's no serving STH).

AlCutter: Grr, yep - that was a bug in the original version and I didn't notice it here.

pphaneuf: In the original, I think this checking was done with the lock held in the ```CleanUpEntriesThread``` loop?

AlCutter: Nope, checkout line 712 in the old version...

pphaneuf:
LGTM, but I have a question on why splitting the sequencing and signing... This is so we put out STHs in a more timely manner? If I'm not mistaken, if there's a huge pile of new certs, this will now issue STHs while sequencing is done (so more incrementally, instead of once after everything is sequenced)?

AlCutter:
Yeah, that's exactly it.

PR #541, opened 2015-03-02
Only send an empty watch update for the initial get.

AlCutter:


cpp/util/etcd.cc
```diff
@@ -514,6 +515,7 @@ struct EtcdClient::WatchState {
 
   int64_t highest_index_seen_;
   map<string, int64_t> known_keys_;
+  bool first_update_;
```
pphaneuf: Could figure that out locally at line 551, if ```highest_index_seen_``` is still -1?

cpp/util/etcd.cc
```diff
@@ -581,8 +588,11 @@ void EtcdClient::WatchInitialGetAllDone(WatchState* state, util::Status status,
 
   state->known_keys_.swap(new_known_keys);
 
-  state->task_->executor()->Add(
-      bind(&EtcdClient::SendWatchUpdates, this, state, move(updates)));
+  if (!updates.empty() || state->first_update_) {
+    state->first_update_ = false;
+    state->task_->executor()->Add(
+        bind(&EtcdClient::SendWatchUpdates, this, state, move(updates)));
```
pphaneuf: Can't do this! ```SendWatchUpdates``` sends the update... *and* starts the next request!

I was going to point out that there's two spots calling it, as well, so if you're going to do it this way, should do it centrally, in ```SendWatchUpdates```, but as it turns out, it's the only way to go.

cpp/util/etcd.cc
AlCutter: Nice, thanks.

cpp/util/etcd.cc
AlCutter: Ah - yes, thank you!
Cool, updated.

cpp/log/cluster_state_controller-inl.h
```diff
@@ -446,11 +449,14 @@ void ClusterStateController<Logged>::DetermineElectionParticipation(
...
 void ClusterStateController<Logged>::ClusterServingSTHUpdater() {
   while (true) {
+    LOG(INFO) << "ClusterServingSTHUpdater going again.";
     std::unique_lock<std::mutex> lock(mutex_);
     update_required_cv_.wait(lock, [this]() {
       return update_required_ || exiting_;
     });
+    LOG(INFO) << "ClusterServingSTHUpdater got ping.";
     if (exiting_) {
+      LOG(INFO) << "ClusterServingSTHUpdater thread returning.";
```
pphaneuf: Not sure if those should be ```VLOG``` or not, but this is fine for now, we can change it later.

pphaneuf:
LGTM, modulo adding detail to that log message (and Travis).

cpp/log/cluster_state_controller-inl.h
AlCutter: They kinda slipped in there, but they might be useful again the in future so I've mostly cranked them down to `VLOG(1)`, a couple remain `LOG(INFO)` though.

PR #543, opened 2015-03-03
Reduce default num followers, select peers at random.

AlCutter:


Makefile
```diff
@@ -8,7 +8,7 @@ all: cpp java python
 	java java_test java_clean cpp cpp_test cpp_clean \
 	python python_test
 
-docker: cpp cpp_test
+docker: cpp
```
pphaneuf: I'm not against this, but you shouldn't be so miserly with your pull requests. :wink:

cpp/fetcher/peer_group.cc
```diff
@@ -87,14 +87,19 @@ shared_ptr<Peer> PeerGroup::PickPeer(const int64_t needed_size) const {
...
     const int64_t tree_size(peer.first->TreeSize());
     group_tree_size = max(group_tree_size, tree_size);
     if (tree_size >= needed_size) {
-      return peer.first;
+      capable_peers.push_back(peer.first);
     }
   }
 
+  if (!capable_peers.empty()) {
+    return capable_peers[std::rand() % capable_peers.size()];
```
pphaneuf: It's not a huge problem if we have a deterministic seed here, but do we seed this somewhere (with ```std::srand```)?

pphaneuf:
LGTM

cpp/fetcher/peer_group.cc
AlCutter: Not explicitly, no.

Makefile
AlCutter: it was getting hairy already having them 4 deep :)

PR #548, opened 2015-03-03
Centralise the Get/GetAll in the watch logic.

pphaneuf:


cpp/util/etcd.cc
```diff
@@ -978,12 +983,7 @@ void EtcdClient::Watch(const string& key, const WatchCallback& cb,
   WatchState* const state(new WatchState(key, cb, task));
   task->DeleteWhenDone(state);
 
-  if (KeyIsDirectory(key)) {
-    GetAll(key,
-           bind(&EtcdClient::WatchInitialGetAllDone, this, state, _1, _2, _3));
-  } else {
-    Get(key, bind(&EtcdClient::WatchInitialGetDone, this, state, _1, _2, _3));
-  }
+  WatchRequestDone(state, nullptr, nullptr);
```
AlCutter: Maybe add a comment here [too] to the effect that this will cause the first request to be fired off?
It could be a little bit confusing given the `...Done()` name.

AlCutter:
LGTM with an extra lil' comment.

PR #551, opened 2015-03-04
Fixed three (!!!) bugs in the GetAll handling.

pphaneuf:


AlCutter:
blimey(!)
LGTM

pphaneuf:
The two missed ```return``` statements look quite bad, but they're really in "can't happen" branches, so they probably don't affect much, really (there's the case where etcd sometimes returns a raw 404 from the Go HTTP library, but since the body isn't JSON, it would bail earlier).

I think the -1 index bit in the case of an empty directory, though, might have gotten some stuck watches? If you do a watch on an empty directory, you might get stuck in some busy loop, until something appears in that directory. I don't think it would have been incorrect, but wouldn't help... :wink:

PR #552, opened 2015-03-05
Set WatchUpdate::exists_ from Node::deleted_.

pphaneuf:
This pull request is specifically small to act as a stepping stone where we are confident in each of the steps.

I'm trying to cut down the number of code paths and combine some parsing code, and I'm thinking ```WatchUpdate``` doesn't need to exist, but I want to make sure this is palatable.


AlCutter:
LGTM

PR #554, opened 2015-03-05
Proxy requests to other nodes when local node is stale.

AlCutter:


cpp/server/proxy.h
```diff
@@ -0,0 +1,170 @@
...
+                                   "and status code."));
+}  // namespace
+
+
+template <class Logged>
+class ClusterStateController;
+class UrlFetcher;
+
+
+template <class Logged>
```
pphaneuf: Could we not make this a template, and "bake in" ```LoggedCertificate```? As in, make the first constructor parameter a ```const ClusterStateController<LoggedCertificate>*```?

I know my little project of de-templating isn't even started, but we're not much closer to starting to instantiate this for other classes than ```LoggedCertificate```, I'd say? Since this bit here doesn't even really care about the template parameter, we could avoid having to live like animals, and have a normal ```.cc``` file, and if we need to re-template it, it'd be really easy?

cpp/server/proxy.h
```diff
@@ -0,0 +1,170 @@
...
+class Proxy {
+ public:
+  Proxy(const ClusterStateController<Logged>* controller, UrlFetcher* fetcher,
+        util::Executor* executor);
+
+  virtual ~Proxy() = default;
+
+  virtual util::Status ProxyRequest(evhttp_request* req) const;
```
pphaneuf: I *really* like how much ```const``` is going on in this class. :heart:

cpp/server/proxy.h
```diff
@@ -0,0 +1,170 @@
...
+
+template <class Logged>
+class ClusterStateController;
+class UrlFetcher;
+
+
+template <class Logged>
+class Proxy {
+ public:
+  Proxy(const ClusterStateController<Logged>* controller, UrlFetcher* fetcher,
```
pphaneuf: Or... maybe you could take an ```std::function``` to give you what you want form the ```controller```? Since it's just this one thing? With a clever signature, you could just ```bind``` it straight on?

cpp/server/proxy.h
AlCutter: I'm not sure I'm entirely behind the de-templating idea; quite a lot of work has gone in to making the code be [relatively] *thing* agnostic (despite there only being a `LoggedCertificate` *thing* currently), and as you know there's interest in things like binary transparency, key transparency, etc. and I know of at least one currently active project who are interested in re-using this codebase for non-`LoggedCertificate` *things*.

That said, there is a lot of code here which is templated to actual *thing* where it really only needs access to a common `LoggedEntry` interface, so there's still a project for a rainy day :)

cpp/server/proxy.h
AlCutter: I'm down with this though, ... and the template is history :)

cpp/server/proxy.h
pphaneuf: I agree with being "thing agnostic", but not so much with templating being used to achieve it (a common ```LoggedEntry``` interface is what I have in mind), but yeah, the rainy days... :wink:

cpp/server/ct-server.cc
```diff
@@ -430,7 +432,12 @@ int main(int argc, char* argv[]) {
   thread signer(&SignMerkleTree, &tree_signer, &consistent_store,
                 &cluster_controller, &election);
   ThreadPool pool;
-  HttpHandler handler(&log_lookup, db, &checker, &frontend, &pool);
+  Proxy proxy(
+      std::bind(&ClusterStateController<LoggedCertificate>::GetFreshNodes,
```
pphaneuf: Could skip the ```std::``` here.

cpp/log/cluster_state_controller-inl.h
```diff
@@ -205,6 +205,41 @@ void ClusterStateController<Logged>::SetNodeHostPort(const std::string& host,
...
+
+template <class Logged>
+std::vector<ct::ClusterNodeState>
+ClusterStateController<Logged>::GetFreshNodes() const {
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (!actual_serving_sth_) {
+    LOG(WARNING) << "Cluster has no ServingSTH, all nodes are stale.";
+    return {};
+  }
+  std::vector<ct::ClusterNodeState> fresh;  // for 1983.
```
pphaneuf: Heh. :smiley:

cpp/server/handler.cc
```diff
@@ -237,38 +241,81 @@ int64_t GetIntParam(const multimap<string, string>& query,
...
+    // automatically mean we're unable to answer this request.
+    if (controller_->NodeIsStale()) {
+      const util::Status status(proxy_->ProxyRequest(request));
+      if (!status.ok()) {
+        LOG(INFO) << "Failed to proxy request: " << status;
+        SendError(request, HTTP_INTERNAL, "Failed to proxy request");
+      }
+      return;
+    } else {
+      local_handler(request);
```
pphaneuf: Uh oh... This is no-go, as the handler is then going to (indirectly) call ```evhttp_send_reply```, which I'm pretty sure needs to be on the libevent thread.

I'm thinking that the check of ```NodeIsStale``` could be done before spinning this off?

This could be a little simpler (just something like ```pool_->Add(bind(&Proxy::ProxyRequest, proxy_, request));```) if ```ProxyRequest``` would take over the request *completely* (and send the error if there's one).

cpp/server/handler.cc
```diff
@@ -237,38 +241,81 @@ int64_t GetIntParam(const multimap<string, string>& query,
...
+    }
+  });
+}
+
+
+void HttpHandler::AddProxyWrappedHandler(
+    libevent::HttpServer* server, const std::string& path,
+    const libevent::HttpServer::HandlerCallback& local_handler) {
+  CHECK(server->AddHandler(path, bind(&HttpHandler::ProxyInterceptor, this,
```
pphaneuf: image

cpp/server/handler.cc
```diff
@@ -278,6 +325,7 @@ void HttpHandler::GetEntries(evhttp_request* req) const {
     return SendError(req, HTTP_BADMETHOD, "Method not allowed.");
   }
 
+
```
pphaneuf: That's in a method, man!

cpp/server/ct-server.cc
AlCutter: skipped

cpp/log/cluster_state_controller-inl.h
AlCutter: Glad you got the reference :)

cpp/server/handler.cc
AlCutter: The inception of this PR predates that knowledge, and I forgot to update it after we learnt that last night :) (Note that the `Blocking...()` methods all originally executed on the pool too.)

Can't check `NodeIsStale()` before spinning off because that's on the libevent thread, and that method can block if the internal lock is held.  I had put a comment there about that, I've updated it to be even more explicit.

I've updated the Proxy to send the error message though, that's a good idea.

cpp/server/handler.cc
AlCutter: We gotta go deeper, man :)

AlCutter:
rebased onto head

cpp/server/json_output.cc
```diff
@@ -0,0 +1,91 @@
...
+  const string uri(evhttp_request_get_uri(req));
+  return string(peer_addr) + " \"" + http_verb + " " + uri + "\" " +
+         std::to_string(http_status) + " " + std::to_string(resp_body_length);
+}
+
+
+}  // namespace
+
+
+JsonOutput::JsonOutput(libevent::Base* base) : base_(base) {
```
pphaneuf: Add ```CHECK_NOTNULL```?

cpp/server/json_output.cc
```diff
@@ -0,0 +1,91 @@
...
+                               resp_body.c_str()),
+           0);
+
+  const string log(LogRequest(req, http_status, resp_body.size()));
+  const auto send_reply([req, http_status, log]() {
+    evhttp_send_reply(req, http_status, /*reason*/ NULL, /*databuf*/ NULL);
+    LOG(INFO) << log;
+  });
+
+  if (!libevent::Base::OnEventThread()) {
```
pphaneuf: Kicking myself for having forgotten about this!

cpp/server/json_output.cc
```diff
@@ -0,0 +1,91 @@
...
+                               const JsonObject& json) {
+  CHECK_EQ(evhttp_add_header(evhttp_request_get_output_headers(req),
+                             "Content-Type", kJsonContentType),
+           0);
+  const string resp_body(json.ToString());
+  CHECK_GT(evbuffer_add_printf(evhttp_request_get_output_buffer(req), "%s",
+                               resp_body.c_str()),
+           0);
+
+  const string log(LogRequest(req, http_status, resp_body.size()));
```
pphaneuf: The fact that GitHub's syntax highlighter marks ```log``` in the same colour as functions makes my Spidey sense tingle... Maybe best to stick with ```logstr``` or something similar, to avoid conflicts? It's a rather common name.

cpp/server/json_output.h
```diff
@@ -0,0 +1,41 @@
...
+struct evhttp_request;
+class JsonObject;
+
+namespace cert_trans {
+namespace libevent {
+class Base;
+}  // namespace libevent
+
+
+static const char kJsonContentType[] = "application/json; charset=ISO-8859-1";
```
pphaneuf: This could be moved to the ```.cc``` wholesale, I think? If not, move the definition?

And "ISO-8859-1"? What is this, the 90's? UTF-8! :wink:

cpp/server/json_output.cc
AlCutter: added

cpp/server/json_output.cc
AlCutter: I re-did that from memory since I'd already split it out before rebasing, and I guess my memory isn't  photographic :)
Changed.

cpp/server/json_output.h
AlCutter: hehe, done, dunno where that came from!

cpp/server/proxy.h
```diff
@@ -0,0 +1,57 @@
+#ifndef CERT_TRANS_SERVER_PROXY_H_
+#define CERT_TRANS_SERVER_PROXY_H_
+
+#include <functional>
+#include <vector>
+
+#include "base/macros.h"
+
+
+struct evhttp_request;
```
pphaneuf: I'd suggest not using forward declarations for external things like this?

cpp/server/proxy.h
```diff
@@ -0,0 +1,57 @@
...
+ public:
+  typedef std::function<std::vector<ct::ClusterNodeState>()>
+      GetFreshNodesFunction;
+  Proxy(libevent::Base* base, JsonOutput* output,
+        const GetFreshNodesFunction& get_fresh_nodes, UrlFetcher* fetcher,
+        util::Executor* executor);
+
+  virtual ~Proxy() = default;
+
+  virtual void ProxyRequest(evhttp_request* req) const;
```
pphaneuf: What's up with all the ```virtual```? I don't see any derived class? Mock?

cpp/server/proxy.h
```diff
@@ -0,0 +1,57 @@
...
+
+ private:
+  libevent::Base* base_;
```
pphaneuf: Make that ```const```?

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+    std::string body(reinterpret_cast<const char*>(evbuffer_pullup(
+                         evhttp_request_get_input_buffer(req), body_length)),
+                     body_length);
+    fetcher_req.body.swap(body);
+  }
+  VLOG(1) << "Proxying request to " << url.Host() << ":" << url.Port()
+          << url.PathQuery();
+  UrlFetcher::Response* resp(new UrlFetcher::Response);
+  fetcher_->Fetch(fetcher_req, resp,
+                  new util::Task(std::bind(&ProxyRequestDone, base_, output_,
```
pphaneuf: Ditch that ```std::``` too? :smiley:

pphaneuf: Well, I now see that you don't actually have any ```using``` statements in this file, so you could generally afford some, now that it's a proper ```.cc```, like for ```std::string```, ```std::vector```, ```ct::ClusterNodeState```, maybe ```util::Task``` and ```std::rand```?

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+#include "net/url_fetcher.h"
+#include "proto/ct.pb.h"
+#include "server/json_output.h"
+#include "util/libevent_wrapper.h"
+
+
+namespace cert_trans {
+
+namespace {
+static Counter<std::string>* total_proxied_requests(
```
pphaneuf: Would add (at least) one empty line after the ```namespace```, but I would feel all right if you clumped the ```namespace```s together.

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+  if (!task->status().ok()) {
+    return output->SendError(request, HTTP_INTERNAL,
+                             "Proxied request failed.");
+  }
+
+  // TODO(alcutter): Consider retrying the proxied request some number of times
+  // in the case where the request fails.
+  for (auto it(response->headers.begin()); it != response->headers.end();
+       ++it) {
+    CHECK_EQ(evhttp_add_header(evhttp_request_get_output_headers(request),
```
pphaneuf: Oh, oops, didn't catch this at first, but this isn't HTTP/1.1... You have to drop any headers that might be mentioned in a "Connection" header (and the "Connection" header itself, IIRC).

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+                                   "status_code",
+                                   "Number of proxied API requests by path "
+                                   "and status code."));
+
+
+void ProxyRequestDone(libevent::Base* base, JsonOutput* output,
+                      evhttp_request* request, const std::string& path,
+                      UrlFetcher::Response* response, util::Task* task) {
+  CHECK_NOTNULL(request);
+  CHECK_NOTNULL(response);
```
pphaneuf: I think ```response``` is leaked?

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+  // TODO(alcutter): Consider retrying the proxied request some number of times
+  // in the case where the request fails.
+  for (auto it(response->headers.begin()); it != response->headers.end();
+       ++it) {
+    CHECK_EQ(evhttp_add_header(evhttp_request_get_output_headers(request),
+                               it->first.c_str(), it->second.c_str()),
+             0);
+  }
+  CHECK_GT(evbuffer_add_printf(evhttp_request_get_output_buffer(request), "%s",
+                               response->body.c_str()),
```
pphaneuf: You could use ```evbuffer_add_reference```, and use its cleanup function to delete ```response```?

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+                     body_length);
+    fetcher_req.body.swap(body);
+  }
+  VLOG(1) << "Proxying request to " << url.Host() << ":" << url.Port()
+          << url.PathQuery();
+  UrlFetcher::Response* resp(new UrlFetcher::Response);
+  fetcher_->Fetch(fetcher_req, resp,
+                  new util::Task(std::bind(&ProxyRequestDone, base_, output_,
+                                           req, url.Path(), resp,
+                                           std::placeholders::_1),
```
pphaneuf: Ooh, it was in my face the whole time, but I forgot ```std::placeholders::_1```! :smiley:

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+
+
+Proxy::Proxy(libevent::Base* base, JsonOutput* output,
+             const GetFreshNodesFunction& get_fresh_nodes, UrlFetcher* fetcher,
+             util::Executor* executor)
+    : base_(base),
+      output_(output),
+      get_fresh_nodes_(get_fresh_nodes),
+      fetcher_(fetcher),
+      executor_(executor) {
```
pphaneuf: I'd pepper all of this with ```CHECK_NOTNULL```.

pphaneuf: And, I guess, a ```CHECK(get_fresh_nodes_);``` in the body, to make sure we're not being set up to fail later?

cpp/server/proxy.cc
pphaneuf: Note that ```response->headers``` is mutable, shouldn't be too hard to do in place?

cpp/log/cluster_state_controller.h
```diff
@@ -61,6 +61,12 @@ class ClusterStateController {
 
   void RefreshNodeState();
 
+  bool NodeIsStale() const;
+
+  // Returns a vector of nodes in the cluster which are able to serve the
+  // cluster's current ServingSTH.
```
pphaneuf: This always exclude the current node, even if it *is* funky fresh, maybe mention that here?

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+}
+
+
+void Proxy::ProxyRequest(evhttp_request* req) const {
+  CHECK_NOTNULL(req);
+
+  const std::vector<ct::ClusterNodeState> fresh_nodes(get_fresh_nodes_());
+  if (fresh_nodes.empty()) {
```
pphaneuf: We can't get here if we're up-to-date, right? Since we skip ourselves, if no other node is fresh, this would be true... I *think* this cannot happen?

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+  }
+  const ct::ClusterNodeState& target(
+      fresh_nodes[std::rand() % fresh_nodes.size()]);
+
+  URL url(evhttp_request_uri(req));
+  url.SetProtocol("http");
+  url.SetHost(target.hostname());
+  url.SetPort(target.log_port());
+
+  UrlFetcher::Request fetcher_req(url);
```
pphaneuf: Could pass in ```evhttp_request_uri(req)``` here, and do the twiddling on ```fetcher.url``` instead, but sure.

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+      fetcher_req.verb = UrlFetcher::Verb::PUT;
+      break;
+    default:
+      return output_->SendError(req, HTTP_BADMETHOD, "Bad method requested.");
+      break;
+  }
+
+  for (evkeyval* ptr = evhttp_request_get_input_headers(req)->tqh_first; ptr;
+       ptr = ptr->next.tqe_next) {
+    fetcher_req.headers.insert(std::make_pair(ptr->key, ptr->value));
```
pphaneuf: Again, shouldn't copy *all* the headers, otherwise problems... Should be able to make a helper function for that.

cpp/server/proxy.cc
```diff
@@ -0,0 +1,134 @@
...
+  }
+  if (fetcher_req.verb == UrlFetcher::Verb::PUT ||
+      fetcher_req.verb == UrlFetcher::Verb::POST) {
+    const size_t body_length(
+        evbuffer_get_length(evhttp_request_get_input_buffer(req)));
+    std::string body(reinterpret_cast<const char*>(evbuffer_pullup(
+                         evhttp_request_get_input_buffer(req), body_length)),
+                     body_length);
+    fetcher_req.body.swap(body);
```
pphaneuf: Could ```evbuffer_drain``` the buffer from the request? I was a bit careful that ```UrlFetcher::Request``` contains *all* the bits necessary for a request, so if you want to retry it, you can just keep that around...

cpp/server/handler.cc
```diff
@@ -488,10 +468,10 @@ void HttpHandler::BlockingGetEntries(evhttp_request* req, int64_t start,
     json_entries.Add(&json_entry);
   }
 
-  shared_ptr<JsonObject> json_reply(make_shared<JsonObject>());
-  json_reply->Add("entries", json_entries);
+  JsonObject json_reply;
+  json_reply.Add("entries", json_entries);
```
pphaneuf: Fix a thing one day, re-fix it better the next. :wink:

pphaneuf:
Okay, done reviewing as of 94213d1120a6bb1e47b8fa5ad319784908abb08f.

cpp/server/proxy.cc
pphaneuf: This is just a suggestion for avoiding a copy, but if you want to keep it simple, make the copy and delete ```response``` here, that's fine by me as well.

cpp/server/proxy.h
AlCutter: removed.

cpp/server/proxy.h
AlCutter: Yeah, it's to leave the door open for Mocks, I didn't add one in this PR though.

cpp/server/proxy.h
AlCutter: dammit, I thought I had.

cpp/server/proxy.cc
AlCutter: Yeah, a hangover from it originally being a template.
added a bunch of `using`s.

cpp/server/proxy.cc
AlCutter: Jiggled around a bit.

cpp/server/proxy.cc
AlCutter: grumble grumble, ok - done.

cpp/server/proxy.cc
AlCutter: Good catch

cpp/server/proxy.cc
AlCutter: I'd rather leave it simple for now, we can come back and add that optimisation later on if need be.

cpp/server/proxy.cc
AlCutter: damn, nearly managed to get one past you.

cpp/server/proxy.cc
AlCutter: Liberally seasoned.

cpp/log/cluster_state_controller.h
AlCutter: Yup, comment updated.

cpp/server/proxy.cc
AlCutter: Yeah, if we're up-to-date then we wouldn't attempt to proxy the request (well, there's a slight race because we could decide we're fresh at the `NodeIsStale()` call and then immediately get an etcd watch update making us stale, but that's kinda what the comment about being smarter about when to proxy is about - if we look at the params of the request we'll know whether we can serve it with the data we have locally - in practice the worst that'll happen is a 400.)

If everyone is stale (perhaps because of an insane ClusterConfig and the signer blowing up) then we could end up here. (This was an advantage of all nodes pulling sequenced entries from etcd during their local signing.)

cpp/server/proxy.cc
AlCutter: done here too

cpp/server/proxy.cc
AlCutter: done.

cpp/server/proxy.cc
```diff
@@ -0,0 +1,181 @@
...
+using std::getline;
+using std::make_pair;
+using std::pair;
+using std::placeholders::_1;
+using std::rand;
+using std::string;
+using std::stringstream;
+using std::unique_ptr;
+using std::unordered_set;
+using std::vector;
```
pphaneuf: There's ```util::Executor``` and ```util::Task``` used below that could be here, but they don't really have to, only a few uses, and they're not too obtrusive, your call.

AlCutter: What the hell, why not.

cpp/server/proxy.cc
```diff
@@ -0,0 +1,183 @@
...
+  });
+}
+
+
+}  // namespace
+
+
+// Filters out any headers which should not be proxied on.
+// See http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.10
+void FilterHeaders(UrlFetcher::Headers* headers) {
```
pphaneuf: Why isn't this in the anonymous namespace?

AlCutter: See comment in the `.h` file

cpp/server/proxy.h
```diff
@@ -0,0 +1,62 @@
...
+namespace cert_trans {
+namespace libevent {
+class Base;
+}  // namespace libevent
+
+class JsonOutput;
+
+
+// Visible for testing
+void FilterHeaders(UrlFetcher::Headers* headers);
```
pphaneuf: Ah, there goes my question. :smiley:

pphaneuf:
LGTM!

PR #556, opened 2015-03-05
Fetch sequenced entries in one go.

AlCutter:


cpp/log/tree_signer-inl.h
```diff
@@ -242,32 +242,36 @@ typename TreeSigner<Logged>::UpdateResult TreeSigner<Logged>::UpdateTree() {
...
+  for (; it != sequenced_entries.end(); ++it) {
     // Paranoid much?
-    CHECK(logged.Entry().has_sequence_number());
-    CHECK_EQ(next_seq, logged.Entry().sequence_number());
+    CHECK(it->Entry().has_sequence_number());
+    CHECK_EQ(next_seq, it->Entry().sequence_number());
 
-    if (!Append(logged.Entry())) {
+    if (!Append(it->Entry())) {
       LOG(ERROR) << "Assigning sequence number failed";
```
pphaneuf: Is this message accurate? Isn't this all already sequenced?

pphaneuf:
LGTM, because I think this is strictly better, but I'm thinking that the signer (which is active on all the nodes) shouldn't be touching the consistent store, it should sign from the local database?

I'm thinking of this to minimize the number of RPCs going to etcd (which is probably our "weakest link"), and that only one node should be copying sequenced entries from the consistent to their local database, which is then going to get replicated to the other nodes local databases, and having the signers work on those local databases.

cpp/log/tree_signer-inl.h
AlCutter: Ah, I guess this is a message from the olden-days, when Appending implied sequencing.
I've updated it.

AlCutter:
Right, there's already a `TODO` in `tree_signer-inl.h`about that.
I'll send a PR which implements that shortly, but as you guessed this PR is more about reducing the etcd load for the moment.

PR #562, opened 2015-03-09
Store sequencing information in a single proto in etcd.

AlCutter:
Here's an implementation of the idea I mentioned the other day about storing sequencing info in a single proto.
We can drop the `provisional_sequence_number` stuff because now the whole sequence info update is atomic with CAS, it also speeds the whole sequencing shebang up quite a bit.

### DIFF
--- a/cpp/log/etcd_consistent_store-inl.h
+++ b/cpp/log/etcd_consistent_store-inl.h
@@ -17,10 +17,24 @@ namespace {
 
 // etcd path constants.
 const char kClusterConfigFile[] = "/cluster_config";
-const char kUnsequencedDir[] = "/unsequenced/";
-const char kSequencedDir[] = "/sequenced/";
+const char kEntriesDir[] = "/entries/";
+const char kSequenceFile[] = "/sequence_mapping";
 const char kServingSthFile[] = "/serving_sth";
 const char kNodesDir[] = "/nodes/";
+
+
+void CheckMappingIsOrdered(const ct::SequenceMapping& mapping) {
+  if (mapping.mapping_size() < 2) {
+    return;
+  }
+  int64_t seq(mapping.mapping(0).sequence_number());
+  for (int64_t i = 1; i < mapping.mapping_size(); ++i) {
+    const int64_t entry_seq(mapping.mapping(i).sequence_number());
+    CHECK_EQ(entry_seq, seq + 1);
+    seq = entry_seq;
+  }
+}
+
 
 }  // namespace
 
@@ -71,14 +85,16 @@ EtcdConsistentStore<Logged>::~EtcdConsistentStore() {
 template <class Logged>
 util::StatusOr<int64_t>
 EtcdConsistentStore<Logged>::NextAvailableSequenceNumber() const {
-  std::vector<EntryHandle<Logged>> sequenced_entries;
-  util::Status status(GetSequencedEntries(&sequenced_entries));
+  EntryHandle<ct::SequenceMapping> sequence_mapping;
+  util::Status status(GetSequenceMapping(&sequence_mapping));
   if (!status.ok()) {
     return status;
   }
-  if (!sequenced_entries.empty()) {
-    CHECK(sequenced_entries.back().Entry().has_sequence_number());
-    return sequenced_entries.back().Entry().sequence_number() + 1;
+  if (sequence_mapping.Entry().mapping_size() > 0) {
+    return sequence_mapping.Entry()
+               .mapping(sequence_mapping.Entry().mapping_size() - 1)
+               .sequence_number() +
+           1;
   }
 
   if (!serving_sth_) {
@@ -179,7 +195,7 @@ template <class Logged>
 util::Status EtcdConsistentStore<Logged>::AddPendingEntry(Logged* entry) {
   CHECK_NOTNULL(entry);
   CHECK(!entry->has_sequence_number());
-  const std::string full_path(GetUnsequencedPath(*entry));
+  const std::string full_path(GetEntryPath(*entry));
   EntryHandle<Logged> handle(full_path, *entry);
   util::Status status(CreateEntry(&handle));
   if (status.CanonicalCode() == util::error::FAILED_PRECONDITION) {
@@ -205,7 +221,7 @@ util::Status EtcdConsistentStore<Logged>::AddPendingEntry(Logged* entry) {
 template <class Logged>
 util::Status EtcdConsistentStore<Logged>::GetPendingEntryForHash(
     const std::string& hash, EntryHandle<Logged>* entry) const {
-  util::Status status(GetEntry(GetUnsequencedPath(hash), entry));
+  util::Status status(GetEntry(GetEntryPath(hash), entry));
   if (status.ok()) {
     CHECK(!entry->Entry().has_sequence_number());
   }
@@ -217,8 +233,7 @@ util::Status EtcdConsistentStore<Logged>::GetPendingEntryForHash(
 template <class Logged>
 util::Status EtcdConsistentStore<Logged>::GetPendingEntries(
     std::vector<EntryHandle<Logged>>* entries) const {
-  util::Status status(
-      GetAllEntriesInDir(GetFullPath(kUnsequencedDir), entries));
+  util::Status status(GetAllEntriesInDir(GetFullPath(kEntriesDir), entries));
   if (status.ok()) {
     for (const auto& entry : *entries) {
       CHECK(!entry.Entry().has_sequence_number());
@@ -238,55 +253,25 @@ bool LessBySequenceNumber(const EntryHandle<Logged>& lhs,
 
 
 template <class Logged>
-util::Status EtcdConsistentStore<Logged>::GetSequencedEntries(
-    std::vector<EntryHandle<Logged>>* entries) const {
-  util::Status status(GetAllEntriesInDir(GetFullPath(kSequencedDir), entries));
-  sort(entries->begin(), entries->end(), LessBySequenceNumber<Logged>);
-  if (status.ok()) {
-    for (const auto& entry : *entries) {
-      CHECK(entry.Entry().has_sequence_number());
-    }
-  }
-  return status;
-}
-
-
-template <class Logged>
-util::Status EtcdConsistentStore<Logged>::GetSequencedEntry(
-    const int64_t sequence_number, EntryHandle<Logged>* entry) const {
-  CHECK_GE(sequence_number, 0);
-  util::Status status(GetEntry(GetSequencedPath(sequence_number), entry));
-  if (status.ok()) {
-    CHECK(entry->Entry().has_sequence_number());
-  }
-
-  return status;
-}
-
-
-template <class Logged>
-util::Status EtcdConsistentStore<Logged>::AssignSequenceNumber(
-    const int64_t sequence_number, EntryHandle<Logged>* entry) {
-  CHECK_GE(sequence_number, 0);
-  CHECK(!entry->Entry().has_sequence_number());
-  if (entry->Entry().has_provisional_sequence_number()) {
-    CHECK_EQ(sequence_number, entry->Entry().provisional_sequence_number());
-  } else {
-    entry->MutableEntry()->set_provisional_sequence_number(sequence_number);
-  }
-  // Record provisional sequence number:
-  util::Status status(UpdateEntry(entry));
+util::Status EtcdConsistentStore<Logged>::GetSequenceMapping(
+    EntryHandle<ct::SequenceMapping>* sequence_mapping) const {
+  util::Status status(GetEntry(GetFullPath(kSequenceFile), sequence_mapping));
   if (!status.ok()) {
     return status;
   }
-  // Now finalise the sequence number assignment.
-  // Create a temporary EntryHandle here to avoid stomping over the version
-  // info held by the unsequenced entry handle passed in:
-  EntryHandle<Logged> seq_entry(GetSequencedPath(sequence_number),
-                                entry->Entry());
-  seq_entry.MutableEntry()->clear_provisional_sequence_number();
-  seq_entry.MutableEntry()->set_sequence_number(sequence_number);
-  return CreateEntry(&seq_entry);
+  CheckMappingIsOrdered(sequence_mapping->Entry());
+  CheckMappingIsContiguousWithServingTree(sequence_mapping->Entry());
+  return util::Status::OK;
+}
+
+
+template <class Logged>
+util::Status EtcdConsistentStore<Logged>::UpdateSequenceMapping(
+    EntryHandle<ct::SequenceMapping>* entry) {
+  CHECK(entry->HasHandle());
+  CheckMappingIsOrdered(entry->Entry());
+  CheckMappingIsContiguousWithServingTree(entry->Entry());
+  return UpdateEntry(entry);
 }
 
 
@@ -543,16 +528,16 @@ util::Status EtcdConsistentStore<Logged>::DeleteEntry(EntryHandle<T>* entry) {
 
 
 template <class Logged>
-std::string EtcdConsistentStore<Logged>::GetUnsequencedPath(
-    const Logged& unseq) const {
-  return GetUnsequencedPath(unseq.Hash());
+std::string EtcdConsistentStore<Logged>::GetEntryPath(
+    const Logged& entry) const {
+  return GetEntryPath(entry.Hash());
 }
 
 
 template <class Logged>
-std::string EtcdConsistentStore<Logged>::GetUnsequencedPath(
+std::string EtcdConsistentStore<Logged>::GetEntryPath(
     const std::string& hash) const {
-  return GetFullPath(std::string(kUnsequencedDir) + util::HexString(hash));
+  return GetFullPath(std::string(kEntriesDir) + util::HexString(hash));
 }
 
 
@@ -564,18 +549,22 @@ std::string EtcdConsistentStore<Logged>::GetNodePath(
 
 
 template <class Logged>
-std::string EtcdConsistentStore<Logged>::GetSequencedPath(int64_t seq) const {
-  CHECK_GE(seq, 0);
-  return GetFullPath(std::string(kSequencedDir) + std::to_string(seq));
-}
-
-
-template <class Logged>
 std::string EtcdConsistentStore<Logged>::GetFullPath(
     const std::string& key) const {
   CHECK(key.size() > 0);
   CHECK_EQ('/', key[0]);
   return root_ + key;
+}
+
+
+template <class Logged>
+void EtcdConsistentStore<Logged>::CheckMappingIsContiguousWithServingTree(
+    const ct::SequenceMapping& mapping) const {
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (serving_sth_ && mapping.mapping_size() > 0) {
+    CHECK_LE(mapping.mapping(0).sequence_number(),
+             serving_sth_->Entry().tree_size());
+  }
 }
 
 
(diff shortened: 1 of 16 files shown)


cpp/log/etcd_consistent_store.h
```diff
@@ -106,12 +103,12 @@ class EtcdConsistentStore : public ConsistentStore<Logged> {
 
   std::string GetUnsequencedPath(const std::string& hash) const;
 
-  std::string GetSequencedPath(int64_t seq) const;
-
   std::string GetNodePath(const std::string& node_id) const;
 
   std::string GetFullPath(const std::string& key) const;
 
+  void CheckMappingIsOrdered(const ct::SequenceMapping& mapping) const;
```
pphaneuf: Doesn't really need to be a member, could be a plain function in the anonymous namespace?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -238,55 +240,38 @@ bool LessBySequenceNumber(const EntryHandle<Logged>& lhs,
...
-    CHECK(entry->Entry().has_sequence_number());
+util::Status EtcdConsistentStore<Logged>::GetSequenceMapping(
+    EntryHandle<ct::SequenceMapping>* sequencing) const {
+  util::Status status(GetEntry(GetFullPath(kSequenceFile), sequencing));
+  if (!status.ok()) {
+    return status;
   }
-
-  return status;
+  CheckMappingIsOrdered(sequencing->Entry());
```
pphaneuf: Maybe we should also check that the sequence number of the first entry is less than or equal to the serving STH tree size?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -238,55 +240,38 @@ bool LessBySequenceNumber(const EntryHandle<Logged>& lhs,
...
-  // info held by the unsequenced entry handle passed in:
-  EntryHandle<Logged> seq_entry(GetSequencedPath(sequence_number),
-                                entry->Entry());
-  seq_entry.MutableEntry()->clear_provisional_sequence_number();
-  seq_entry.MutableEntry()->set_sequence_number(sequence_number);
-  return CreateEntry(&seq_entry);
+util::Status EtcdConsistentStore<Logged>::UpdateSequenceMapping(
+    EntryHandle<ct::SequenceMapping>* entry) {
+  CHECK(entry->HasHandle());
+  CheckMappingIsOrdered(entry->Entry());
```
pphaneuf: Same thing here?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -648,32 +626,38 @@ util::Status EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
   }
 
-  for (auto& entry : sequenced_entries) {
-    const uint64_t sequence_number(entry.Entry().sequence_number());
-
-    // return if we've hit the limit
-    if (entry.Entry().sequence_number() > clean_up_to_sequence_number) {
-      return util::Status::OK;
-    }
+  size_t i;
```
pphaneuf: Style guide says to use unsigned integers, *especially* for iteration...

If I'm not mistaken, it's also not used outside of the loop? If so, I'd declare it in the ```for``` itself.

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -18,7 +18,7 @@ namespace {
 // etcd path constants.
 const char kClusterConfigFile[] = "/cluster_config";
 const char kUnsequencedDir[] = "/unsequenced/";
```
pphaneuf: They're not really "unsequenced" there anymore, they're more just... entries, right? Maybe rename this, otherwise it might be a bit confusing down the road?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -18,7 +18,7 @@ namespace {
...
-const char kSequencedDir[] = "/sequenced/";
+const char kSequenceFile[] = "/sequencing";
```
pphaneuf: Not sure I find that name very enticing, but I haven't got much better... "mapping"?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -648,32 +626,38 @@ util::Status EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
-      // Log a warning here, but don't bail on the clean up as we could just be
-      // recovering from a crash halfway through a previous clean up.
-      LOG(WARNING) << "Cleanup couldn't get unsequenced entry ("
-                   << util::ToBase64(hash)
-                   << ") corresponding to sequenced entry #" << sequence_number
-                   << " : " << status;
+      if (status.CanonicalCode() == util::error::NOT_FOUND) {
+        // Log a warning here, but don't bail on the clean up as we could just
+        // be recovering from a crash halfway through a previous clean up.
+        LOG(WARNING) << "Cleanup couldn't get unsequenced entry ("
```
pphaneuf: These aren't really unsequenced entries anymore, right? (hence my comment higher up in this file)

If you change this, also change the log message at line 664.

pphaneuf: And 671.

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -686,18 +670,21 @@ util::Status EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
-      LOG(WARNING) << "Cleanup couldn't delete sequenced entry #"
-                   << sequence_number << " (but corresponding unsequenced "
-                   << "entry was successfully deleted) : " << status;
-      return status;
-    }
-    VLOG(1) << "Cleanup deleted sequenced entry #" << sequence_number;
   }
-  return util::Status::OK;
+  // Now remove mappings for deleted sequence numbers.
+  size_t to, from(i);
```
pphaneuf: Again, use unsigned integers here (and possibly declare them inside the ```for```, if it makes sense).

cpp/log/etcd_consistent_store-inl.h
pphaneuf: I saw the use of ```i``` below, and I'm not sure I like it. :smiley: 

Either it's going to be one past ```clean_up_to_sequence_number```, or we'll have bailed out of this method altogether, right?

pphaneuf: No, uh, wait... Crap. Anyway, you can recalculate it below? ```clean_up_to_sequence_number - sequence_mapping.Entry().mapping(0).sequence_number()```? Possibly off by one, but you know what I mean... :wink:

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -686,18 +670,21 @@ util::Status EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
+  for (to = 0; from < sequence_mapping.Entry().mapping_size(); ++from, ++to) {
+    sequence_mapping.MutableEntry()->mutable_mapping()->SwapElements(from, to);
```
pphaneuf: Add a ```CHECK_GT(to, from);```, to ensure we haven't screwed up?

pphaneuf: Other than the logging (that I'd expect to be a bit noisy, actually, maybe collapse it into a single message with the first and last sequence numbers?), why not ```DeleteSubrange``` directly?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -686,18 +670,21 @@ util::Status EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
+    VLOG(1) << "Cleanup keeping sequence mapping for "
+            << sequence_mapping.Entry().mapping(to).sequence_number();
+  }
+  sequence_mapping.MutableEntry()->mutable_mapping()->DeleteSubrange(to, i);
+  for (auto m : sequence_mapping.Entry().mapping()) {
+    VLOG(1) << "Kept " << m.sequence_number() << " => "
+            << util::ToBase64(m.unsequenced_hash());
+  }
+  status = UpdateSequenceMapping(&sequence_mapping);
+  return status;
```
pphaneuf: I was going to say that you could ```return UpdateSequenceMapping(&sequence_mapping);``` directly, but I think a log message in case of an error would be nice?

It's possible that a lower level already does log errors, in which case, return straight up?

cpp/log/log_lookup_test.cc
```diff
@@ -59,6 +62,13 @@ class LogLookupTest : public ::testing::Test {
                   new MerkleVerifier(new Sha256Hasher())) {
     // Set some noddy STH so that we can call UpdateTree on the Tree Signer.
     store_.SetServingSTH(ct::SignedTreeHead());
+    // Force an empty sequence mapping file:
```
pphaneuf: Not directly related to this pull request, but I'm not entirely clear why this test needs a ```ConsistentStore``` (```LogLookup``` only needs a ```ReadOnlyDatabase```)? Is this testing a lot more than just ```LogLookup```?

cpp/log/strict_consistent_store.h
```diff
@@ -29,8 +29,8 @@ class StrictConsistentStore : public ConsistentStore<Logged> {
 
   util::Status SetServingSTH(const ct::SignedTreeHead& new_sth) override;
 
-  util::Status AssignSequenceNumber(const int64_t sequence_number,
-                                    EntryHandle<Logged>* entry) override;
+  util::Status UpdateSequenceMapping(
+      EntryHandle<ct::SequenceMapping>* entry) override;
```
pphaneuf: A bit late now, but this pull request is a rather complex one, could certainly have done without the noise! :stuck_out_tongue:

cpp/log/tree_signer-inl.h
```diff
@@ -100,44 +77,24 @@ uint64_t TreeSigner<Logged>::LastUpdateTime() const {
...
-          << "number:\n" << presequenced.Entry().DebugString()
-          << "\nclearing provisional_sequence_number to "
-          << "resequence pending entry";
-      // This entry is already sequenced just fine, no need to return it.
-      VLOG(1) << "Entry already sequenced: "
-              << util::ToBase64(it->Entry().Hash()) << " ("
-              << presequenced.Entry().sequence_number()
-              << "), removing from list";
+    if (sequenced_hashes.find(it->Entry().Hash()) != sequenced_hashes.end()) {
+      // This hash is already sequenced.
```
pphaneuf: Check that the sequence number in the entry matches that of the mapping? This is very sensitive, and I'd rather be paranoid here...

cpp/log/tree_signer-inl.h
```diff
@@ -155,8 +112,22 @@ util::Status TreeSigner<Logged>::SequenceNewEntries() {
...
   CHECK_GE(next_sequence_number, 0);
   VLOG(1) << "Next available sequence number: " << next_sequence_number;
+
+  EntryHandle<ct::SequenceMapping> mapping;
+  util::Status status(consistent_store_->GetSequenceMapping(&mapping));
+  if (!status.ok()) {
+    return status;
+  }
+
+  // hashes which are already sequenced.
```
pphaneuf: Capitalize the sentence?

cpp/log/tree_signer-inl.h
```diff
@@ -178,23 +148,41 @@ util::Status TreeSigner<Logged>::SequenceNewEntries() {
...
+      ct::SequenceMapping::Mapping* m(mapping.MutableEntry()->add_mapping());
+      m->set_sequence_number(next_sequence_number);
+      m->set_unsequenced_hash(pending_entry.Entry().Hash());
+      pending_entry.MutableEntry()->set_sequence_number(next_sequence_number);
+      ++num_sequenced;
+      ++next_sequence_number;
+    } else {
+      VLOG(1) << "Previously sequenced " << util::ToBase64(pending_hash)
+              << " = " << next_sequence_number;
+      pending_entry.MutableEntry()->set_sequence_number(seq_it->second);
```
pphaneuf: I feel like we should have some more paranoid checks here, somehow... Okay, so we found it in the list of sequenced hashes, but does the sequence number match what we expect? Is it the same as the one we would have given it if we hadn't found it? What if there's a gap somewhere, for some crazy reason? ```NextAvailableSequenceNumber``` has its own call to ```GetSequenceMapping```, and yes, they *really* should be working with the same information, but... I'd like to catch them if they're not?

cpp/log/etcd_consistent_store-inl.h
AlCutter: "mapping" could be anything, I think it needs to have some reference to "sequence" in there, I'll compromise with "/sequence_mapping" :)

cpp/log/etcd_consistent_store-inl.h
AlCutter: Added.

cpp/log/etcd_consistent_store-inl.h
AlCutter: Here too

cpp/log/etcd_consistent_store-inl.h
AlCutter: Or just, you know, use this one we've already got without the fear of being off-by-one? :)
On the plus side, I've dinged all the swapping stuff, and added a few more comments which has made it a bit clearer.

`size_t` is unsigned, but I suspect you meant *use signed* :)

cpp/log/etcd_consistent_store-inl.h
AlCutter: Yeah, I didn't do it because I thought it might well ripple out, but it's got to be done at some point anyway, may as well be now.
Done.

cpp/log/etcd_consistent_store-inl.h
AlCutter: gone away

cpp/log/etcd_consistent_store-inl.h
AlCutter: gone away

cpp/log/etcd_consistent_store-inl.h
AlCutter: it probably was logging something for debugging and I accidentally left it like that.
The caller didn't log errors, but it does now.

cpp/log/tree_signer-inl.h
pphaneuf: Umm... The updated entry is never written back to etcd? So, in etcd, the ```sequence_number``` field is always unset, it's all about the mapping?

cpp/log/etcd_consistent_store.h
AlCutter: Ah yes, it *was* like that originally, but I got a multiple definition error when linking and (for some weird reason) assuming it was because ${template}, but that's clearly bogus and upon closer inspection it's just because cluster_state_controller-inl.h was directly including this file.  Fixed that, fixed this.

cpp/log/strict_consistent_store.h
AlCutter: what noise?

pphaneuf:
Okay, I'm done with this round of reviewing. I'm trying to make really sure that I understand the new algorithm, and that it makes sense. :wink:

I may have made assumptions when writing some of the earlier comments, like that entries in etcd would have their ```sequence_number``` set when sequenced, so forgive me if I don't always (ever?) make sense?

cpp/log/etcd_consistent_store-inl.h
pphaneuf: Duh, yeah, *signed*, not unsigned! :smiley:

cpp/log/strict_consistent_store.h
pphaneuf: Adding the ```override```s in this file. Could also have remove the signing of entries in etcd in a separate commit, but after all the rest, that was easy. :wink:

pphaneuf:
Just to make sure I understand this correctly... The big improvement here is basically to have just one etcd RPC per round of sequencing?

I had the idea of dropping the mapping itself, and just updating the ```sequence_number``` in-place, but that would be O(N) etcd RPCs (the old way was O(2N), so kind of better, despite what those silly computer scientists would say).

cpp/log/tree_signer-inl.h
AlCutter: This is actually dead code now, I've removed it.
But we couldn't have done that anyway because the entry in etcd doesn't get updated with the sequence number, that info lives in the `sequence_mapping` file.

cpp/log/tree_signer-inl.h
AlCutter: Done

cpp/log/tree_signer-inl.h
pphaneuf: Yeah, I got that later on... I suppose this means that the entries in etcd are still "unsequenced", one could say.

cpp/log/tree_signer-inl.h
AlCutter: Yup (all about the mapping)

cpp/log/etcd_consistent_store-inl.h
AlCutter: Yeah, see my reply to the other comment related to this for details, but I've changed it.

AlCutter:
Right, etcd is slooow and so we're going from O(2N) to O(1) in RPCs, which is significant when N is a couple of thousand and each RPC is taking several hundred ms !

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -238,55 +260,31 @@ bool LessBySequenceNumber(const EntryHandle<Logged>& lhs,
...
-
-template <class Logged>
-util::Status EtcdConsistentStore<Logged>::GetSequencedEntry(
-    const int64_t sequence_number, EntryHandle<Logged>* entry) const {
-  CHECK_GE(sequence_number, 0);
-  util::Status status(GetEntry(GetSequencedPath(sequence_number), entry));
-  if (status.ok()) {
-    CHECK(entry->Entry().has_sequence_number());
+  CheckMappingIsOrdered(sequence_mapping->Entry());
+  if (serving_sth_) {
```
pphaneuf: You need a lock to access this, no?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -238,55 +260,31 @@ bool LessBySequenceNumber(const EntryHandle<Logged>& lhs,
...
-  }
-  // Record provisional sequence number:
-  util::Status status(UpdateEntry(entry));
-  if (!status.ok()) {
-    return status;
+util::Status EtcdConsistentStore<Logged>::UpdateSequenceMapping(
+    EntryHandle<ct::SequenceMapping>* entry) {
+  CHECK(entry->HasHandle());
+  CheckMappingIsOrdered(entry->Entry());
+  if (serving_sth_) {
```
pphaneuf: Same here, need the lock?

cpp/log/tree_signer.h
```diff
@@ -63,6 +63,7 @@ class TreeSigner {
 
  private:
   util::Status HandlePreviouslySequencedEntries(
+      const ct::SequenceMapping& mapping,
```
pphaneuf: You deleted this method's definition, should delete the declaration too.

cpp/log/strict_consistent_store.h
AlCutter: Ah soz, I guess I let it all kinda pile up a bit since you were away :/

cpp/log/etcd_consistent_store-inl.h
AlCutter: Frickin' stupid brain.
done.

cpp/log/etcd_consistent_store-inl.h
AlCutter: I've moved that `CheckMappingIsContiguousWithServingTree()` method into `EtcdConsistentStore` and grab the lock locally now.

cpp/log/tree_signer.h
AlCutter: A keepsake for sentimental reasons?
no? ok - deleted.

cpp/server/ct-server.cc
```diff
@@ -202,11 +202,15 @@ void SequenceEntries(TreeSigner<LoggedCertificate>* tree_signer,
...
-      store->CleanupOldEntries();
-
-      // TODO(alcutter): Probably don't need to blow up here
-      CHECK_EQ(util::Status::OK, tree_signer->SequenceNewEntries())
-          << "Problem sequencing new entries";
+      util::Status status(store->CleanupOldEntries());
+      if (!status.ok()) {
+        LOG(WARNING) << "Problem cleaning up old entries: " << status;
+      } else {
+        status = tree_signer->SequenceNewEntries();
```
pphaneuf: Only if there was no problem cleaning up? I agree these should be done one after the other (because they'll both try to modify the mapping entry), but it should be fine to at least attempt to sequence even if there was a problem cleaning up, I'd say.

AlCutter: I figured that if the clean-up failed it's likely because of some transient problem (or at least, that's the best possible explanation!), so going back and hitting etcd again straight away was likely to also fail and so best to skip the rest of the run entirely.

I don't feel strongly about it though, so feel free to convince me...

pphaneuf: I agree that this is the most likely thing to happen, but I wouldn't want some minor issue in the not-critical cleanup hold up sequencing either. Hitting etcd again straightaway seems like the lesser evil?

If it wasn't for the contention on the mapping key, these two things could be completely separate, so at least we can limit the damage of a possible bug or issue in the cleanup, is my theory.

AlCutter: I was hoping for a bribe of some kind at the very least.
Fair point about hitting etcd vs getting stuck, I'll change it.

pphaneuf:
LGTM

PR #563, opened 2015-03-09
Disable exceptions.

pphaneuf:
As per the [style guide](http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Exceptions), we're already avoiding exceptions, but disabling them altogether lets us see better where any library-generated exceptions are (uncaught exceptions in threads only report the thread entry point, which makes it rather difficult to debug).

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -5,7 +5,7 @@ default: all
 
 -include local.mk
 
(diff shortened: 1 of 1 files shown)


AlCutter:
Lgtm

PR #564, opened 2015-03-09
Fix a few bugs in the HTTP handlers.

pphaneuf:
A use-after-free, and some uses of non-thread-safe objects from the wrong thread.


AlCutter:
Lgtm

PR #565, opened 2015-03-10
Centralise the parsing of etcd nodes.

pphaneuf:
This one is a bit more delicate... I don't think I changed the semantics in any significant way? It now a bit more strict, checking for some things that it might not need, but I checked that the real etcd does provide them in all the cases I changed (I did have to update ```fake_etcd.cc``` to follow suit).

### DIFF
--- a/cpp/util/etcd.cc
+++ b/cpp/util/etcd.cc
@@ -97,6 +97,37 @@ Status StatusFromResponseCode(const int response_code,
 }
 
 
+StatusOr<EtcdClient::Node> ParseNodeFromJson(const JsonObject& json_node) {
+  const JsonInt createdIndex(json_node, "createdIndex");
+  if (!createdIndex.Ok()) {
+    return Status(util::error::FAILED_PRECONDITION,
+                  "Invalid JSON: Couldn't find 'createdIndex'");
+  }
+
+  const JsonInt modifiedIndex(json_node, "modifiedIndex");
+  if (!modifiedIndex.Ok()) {
+    return Status(util::error::FAILED_PRECONDITION,
+                  "Invalid JSON: Couldn't find 'modifiedIndex'");
+  }
+
+  const JsonString key(json_node, "key");
+  if (!key.Ok()) {
+    return Status(util::error::FAILED_PRECONDITION,
+                  "Invalid JSON: Couldn't find 'key'");
+  }
+
+  const JsonString value(json_node, "value");
+  const JsonBoolean isDir(json_node, "dir");
+  // TODO(pphaneuf): The Node class should support directories.
+  const bool is_dir(isDir.Ok() && isDir.Value());
+  const bool deleted(!value.Ok() && !is_dir);
+
+  return EtcdClient::Node(createdIndex.Value(), modifiedIndex.Value(),
+                          key.Value(),
+                          (deleted || is_dir) ? "" : value.Value(), deleted);
+}
+
+
 void GetRequestDone(const string& keyname, EtcdClient::GetResponse* resp,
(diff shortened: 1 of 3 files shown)


cpp/util/etcd.cc
```diff
@@ -108,44 +139,27 @@ void GetRequestDone(const string& keyname, EtcdClient::GetResponse* resp,
...
-    parent_task->Return(Status(util::error::FAILED_PRECONDITION,
-                               "Invalid JSON: Couldn't find 'key'"));
-    return;
-  }
-
-  const JsonString value(node, "value");
-  if (!value.Ok()) {
+  if (node.ValueOrDie().deleted_) {
     parent_task->Return(Status(util::error::FAILED_PRECONDITION,
                                "Invalid JSON: Couldn't find 'value'"));
```
AlCutter: Kinda reads a bit weirdly now with this error message for `if (node.ValueOrDie().deleted_)`, maybe update it too?

cpp/util/etcd.cc
```diff
@@ -191,47 +213,27 @@ void GetAllRequestDone(const string& dir, EtcdClient::GetAllResponse* resp,
...
       return;
     }
 
-    const JsonInt modifiedIndex(entry, "modifiedIndex");
-    if (!modifiedIndex.Ok()) {
-      parent_task->Return(
-          Status(util::error::FAILED_PRECONDITION,
-                 "Invalid JSON: Coulnd't find 'modifiedIndex'"));
+    StatusOr<EtcdClient::Node> entry(ParseNodeFromJson(json_entry));
+    if (!node.status().ok()) {
```
AlCutter: `entry.status().ok()` perhaps?

cpp/util/etcd.cc
```diff
@@ -191,47 +213,27 @@ void GetAllRequestDone(const string& dir, EtcdClient::GetAllResponse* resp,
...
+      parent_task->Return(node.status());
```
AlCutter: and similarly `entry.status()` here too

cpp/util/etcd.cc
```diff
@@ -191,47 +213,27 @@ void GetAllRequestDone(const string& dir, EtcdClient::GetAllResponse* resp,
...
       return;
     }
 
-    const JsonString key(entry, "key");
-    if (!key.Ok()) {
+    if (entry.ValueOrDie().deleted_) {
       parent_task->Return(Status(util::error::FAILED_PRECONDITION,
-                                 "Invalid JSON: Couldn't find 'key'"));
+                                 "Invalid JSON: Couldn't find 'value'"));
```
AlCutter: Same comment as earlier about the status message here

cpp/util/etcd.cc
```diff
@@ -633,16 +582,16 @@ void EtcdClient::WatchRequestDone(WatchState* state, GenericResponse* gen_resp,
...
-    if (!node.Ok()) {
+    const JsonObject json_node(*gen_resp->json_body, "node");
+    if (!json_node.Ok()) {
       LOG(INFO) << "Invalid JSON: Couldn't find 'node'";
       goto fail;
     }
 
     vector<Node> updates;
-    StatusOr<Node> status(UpdateForNode(node));
+    StatusOr<Node> status(ParseNodeFromJson(json_node));
```
AlCutter: /status/node/ ?

cpp/util/etcd.cc
pphaneuf: Oh dear, good catch!

pphaneuf:
Addressed the comments.

PR #566, opened 2015-03-12
Avoid the use of Database<>::TreeSize in TreeSigner.

pphaneuf:
Looking to make ```TreeSize``` work lazily, so since that means it could do disk I/O, trying to remove it from any hot spots. This is just one I found, not all of it!


AlCutter:
The call I'm in the process of adding in `NodeIsStale()` in my proxy PR will be hot...

pphaneuf:
FYI, this is part of #568.

PR #567, opened 2015-03-12
Bit of hackery to allow easy building and testing with TSAN

AlCutter:
I really wanted to be able to have `make TSAN=1 <whatever>` just work, but that'd require determining whether the previous run had it enabled or not and cleaning/fiddling/rebuilding gmock as appropriate and my Make-fu is weak. 
It works well enough for me though like this.

```
$ make -j24 tsan_enabled
$ make -j24 
$ make test
# other makery
# ...
# finished tsanning
$ make tsan_disabled
```

Plenty of things to fix :)

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -5,14 +5,14 @@ default: all
 
 -include local.mk
 
-CXXFLAG = -std=c++11 -fno-exceptions -Wall -Werror -g -O3
+CXXFLAG = -std=c++11 -fno-exceptions -Wall -Werror -g -O3 --stdlib=libstdc++
 LINK.o = $(LINK.cc)
 
 INCLUDE = -I .
 LOCAL_LIBS = log/libcert.a log/libcluster.a log/libdatabase.a \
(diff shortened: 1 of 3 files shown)


pphaneuf:
Yowser. I... can't quite bring myself to say "looks good to me" here (don't misread any such intention in the labels!), but, um, sure, you can push this. :stuck_out_tongue:

Just so you know, the traditional way to do such things with a nice-enough makefile (say, one spewed out by automake is nice *enough*) is to use a separate build directory for each different build options you might have. Something like this:

```shell
mkdir build-wacky
cd build-wacky
# The "configure" script looks where it is to look for the source
# files, and keeps any configuration variables passed to it as
# parameters into the Makefile it generates into the current
# working directory.
../configure CXXFLAGS=-fwacky-option
make
```

AlCutter:
I'll take that comment as a firm commitment that you'll be writing us an `automake` config RSN then!

pphaneuf:
There *is* a ```autosnake``` branch in my fork... :wink:

PR #569, opened 2015-03-12
Lazy-load the TLD list in CheckTldMatches

phad:
We don't want the tld_list.TLDList() to be instantiated during module loading.
Defer instantation until the first call to tld_check.CheckTldMatches.check()

### DIFF
--- a/python/ct/cert_analysis/tld_check_test.py
+++ b/python/ct/cert_analysis/tld_check_test.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+# coding=utf-8
+import mock
+import unittest
+from ct.cert_analysis import tld_check
+
+def gen_dns_name(name):
+    dns_name = mock.Mock()
+    dns_name.value = name
+    return dns_name
+
+EXAMPLE = gen_dns_name("example.com")
+
+class TLDCheckTest(unittest.TestCase):
+    @mock.patch('ct.cert_analysis.tld_list.TLDList')
+    def test_tld_list_not_created_until_check_called(self, tld_list):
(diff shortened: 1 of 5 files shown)


pphaneuf:
Before looking at this, it seems that your "master" is somewhat out of date? Could you pull and rebase on the current one?

phad:
Damn.  I've pull upstream/master and rebased.

python/ct/cert_analysis/tld_check_test.py
```diff
@@ -0,0 +1,21 @@
...
+    dns_name = mock.Mock()
+    dns_name.value = name
+    return dns_name
+
+EXAMPLE = gen_dns_name("example.com")
+
+class TLDCheckTest(unittest.TestCase):
+    def test_tld_list_not_created_until_check_called(self):
+        self.assertIsNone(tld_check.CheckTldMatches.TLD_LIST_)
+        _ = tld_check.CheckTldMatches.check([EXAMPLE], "dNSNames: ")
```
eranmes: Nit: This test will fail when there's no network connectivity. If you really wanted you could mock the tld_list package and have it return a dummy object.

pphaneuf:
There was some spurious Travis failure, but I re-ran it, and now there's what looks like a genuine failure?

phad:
Yes, I'm trying to reproduce it.  It doesn't occur on my local build.

phad:
Travis is green:  good to go, Pierre/Eran?  I'll squash the commits.

pphaneuf:
LGTM

PR #571, opened 2015-03-13
Improve signal handling.

robstradling:
During startup, ignore SIGHUP, SIGINT and SIGTERM signals.

Exit the libevent dispatch loop when a SIGHUP, SIGINT or SIGTERM signal is received, allowing the process to terminate normally.

### DIFF
--- a/cpp/server/ct-server.cc
+++ b/cpp/server/ct-server.cc
@@ -6,6 +6,7 @@
 #include <iostream>
 #include <memory>
 #include <openssl/err.h>
+#include <signal.h>
 #include <string>
 
 #include "log/cert_checker.h"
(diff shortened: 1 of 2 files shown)


eranmes:
LGTM, but maybe @AlCutter or @pphaneuf will have other insights.

pphaneuf:
I didn't like too much the setting of signal handlers in a library rather than in (or very near) the ```main()```, but it would be a bit awkward to do it that way in ```master```, so maybe we should live with this (since it's an helpful change), and make it better in ```super_duper```?

Sorry for the delay, I wasn't paying good attention.

robstradling:
Pierre, thanks for merging this.  :-)

PR #576, opened 2015-03-16
Downgrade from logging.critical to logging.error when exception caught

phad:
during writing report.

### DIFF
--- a/python/ct/client/reporter.py
+++ b/python/ct/client/reporter.py
@@ -135,10 +135,10 @@ def handle_writing(queue, report):
             result = result.get()
         except PoolException as e:
             ex, ex_tb, first, last, bad_one = e.failure
-            logging.critical(ex_tb)
-            logging.critical(ex.args[0])
-            logging.critical("Entry %d in batch <%d, %d> %s" % (bad_one,
-                             first, last, "raised an exception during scan"))
+            logging.error(ex_tb)
+            logging.error(ex.args[0])
+            logging.error("Entry %d in batch <%d, %d> %s" % (bad_one,
+                          first, last, "raised an exception during scan"))
         else:
             report._batch_scanned_callback(result)
         queue.task_done()


python/ct/client/reporter.py
```diff
@@ -135,10 +135,10 @@ def handle_writing(queue, report):
...
         except PoolException as e:
             ex, ex_tb, first, last, bad_one = e.failure
-            logging.critical(ex_tb)
-            logging.critical(ex.args[0])
-            logging.critical("Entry %d in batch <%d, %d> %s" % (bad_one,
-                             first, last, "raised an exception during scan"))
+            logging.error(ex_tb)
+            logging.error(ex.args[0])
+            logging.error("Entry %d in batch <%d, %d> %s" % (bad_one,
+                          first, last, "raised an exception during scan"))
```
pphaneuf: I think the intention was to crash here, since the internal state is probably undefined at that point? Add something after the logging to accomplish that?

pphaneuf:
I have some vague memories of things getting wedged (no progress, no error message) when an unhandled exception came to this level, but at least you've got a log message now, so that should be all right (and at least let you debug the problem).

LGTM

phad:
Right, I don't know what the state is like at this point so would at least like to find out to start with.
It might be that we have no sensible option but to terminate.  I don't know right now if that's the case.

eranmes:
If the intention was to terminate, it was done rather poorly - the first logging.critical message would terminate the process...
When this exception is caught the implication is that _scan_der_cert in this file threw an unknown exception (probably because a truly odd error was encountered while parsing the certificate). All it means is that a batch of certificates was not fully processed, that should not justify exiting.

pphaneuf:
I was under the (incorrect) impression that ```logging.critical``` terminated (which, as you point out, would be a bit dubious, if you have more than one!), but it doesn't, it's just a higher log level. So this change is *really* fine!

PR #577, opened 2015-03-16
Test EtcdClient at the UrlFetcher level.

pphaneuf:
This is a straightforward conversion, but this allows testing of the parsing layer as well, which was previously mocked out inside ```EtcdClient::Generic```.

### DIFF
--- a/cpp/util/etcd_test.cc
+++ b/cpp/util/etcd_test.cc
@@ -2,11 +2,10 @@
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
-#include <map>
 #include <memory>
 #include <string>
 
-#include "base/notification.h"
+#include "net/mock_url_fetcher.h"
 #include "util/json_wrapper.h"
 #include "util/sync_task.h"
 #include "util/testing.h"
@@ -17,22 +16,18 @@ using std::bind;
 using std::chrono::seconds;
 using std::make_pair;
 using std::make_shared;
-using std::map;
-using std::pair;
 using std::placeholders::_1;
 using std::placeholders::_2;
 using std::placeholders::_3;
-using std::placeholders::_4;
-using std::placeholders::_5;
 using std::shared_ptr;
 using std::string;
-using std::vector;
-using testing::AllOf;
-using testing::Contains;
+using std::to_string;
 using testing::ElementsAre;
-using testing::Field;
 using testing::Invoke;
+using testing::IsEmpty;
 using testing::Pair;
+using testing::StrCaseEq;
+using testing::StrictMock;
 using testing::_;
 using util::Status;
 using util::SyncTask;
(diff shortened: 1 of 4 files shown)


cpp/util/etcd_test.cc
```diff
@@ -184,53 +169,79 @@ class EtcdTest : public ::testing::Test {
...
+  EtcdClient client_;
 };
 
-void GenericReturn(Status status, const shared_ptr<JsonObject>& json_body,
-                   int64_t etcd_index, EtcdClient::GenericResponse* resp,
-                   Task* task) {
-  resp->etcd_index = etcd_index;
-  resp->json_body = json_body;
+string GetEtcdUrl(const string& key) {
+  return "http://" + string(kEtcdHost) + ":" + to_string(kEtcdPort) +
```
AlCutter: Worth a `CHECK(!key.empty() && key[0] == '/')`here just to avoid potential typo-induced head-scratchers?

cpp/util/etcd_test.cc
```diff
@@ -245,78 +256,106 @@ TEST_F(EtcdTest, TestGetAll) {
...
+      url_fetcher_,
+      Fetch(IsUrlFetchRequest(
+                UrlFetcher::Verb::PUT, URL(GetEtcdUrl(kEntryKey)),
+                ElementsAre(Pair(StrCaseEq("content-type"),
+                                 "application/x-www-form-urlencoded")),
+                "consistent=true&prevExist=false&quorum=true&value=123"),
+            _, _))
+      .WillOnce(
+          Invoke(bind(HandleFetch, Status::OK, 200,
+                      UrlFetcher::Headers{make_pair("x-etcd-index", "1")},
```
AlCutter: This is kinda weird in that `X-Etcd-Index` will be less than the `modifiedIndex` of whatever's returned, is that deliberate to ensure we choose the higher of the two?

pphaneuf: I actually ported that straight through (note the 4th parameter of the ```bind()``` that was deleted)...

```EtcdClient``` is kind of inconsistent (at the moment) about what it puts in the ```etcd_index``` field of its ```Response```: sometimes it's the ```X-Etcd-Index``` header, sometimes it's from the Node that it received... I'm being very careful when changing this at the moment, to avoid breaking anything, but my intent is for the ```Response::etcd_index``` to always be the ```X-Etcd-Index```, and if you care about one of the node indexes, just look at them yourself, since we now provide them all (we didn't use to).

pphaneuf: I was implying that as I sort this out, and make the ```Response::etcd_index``` more rigorous, I'll tighten up the test (many of the tests aren't even checking the value of that field!).

I've seen actual etcd replies that had a lower ```X-Etcd-Index``` than the index in their JSON payload, so I guess we shouldn't be too strict on this... :wink:

cpp/util/etcd_test.cc
pphaneuf: Done.

PR #579, opened 2015-03-17
Python: Fix CheckValidityNotBeforeFuture

linshunghuang:
Python test test_not_before_regular was failing on my Mac because `CheckValidityNotBeforeFuture()` was comparing `time.gmtime()` to `datetime.datetime.utcnow().timetuple()` which uses different `dst` flags.

```
PYTHONPATH=:. ./ct/cert_analysis/validity_test.py
F....
FAIL: test_not_before_regular (__main__.ValidityTest)
  File "./ct/cert_analysis/validity_test.py", line 15, in test_not_before_regular
    self.assertIsNone(result)
AssertionError: [NotBeforeInFuture('Validity: not before is in future', None, time.struct_time(tm_year=2015, tm_mon=3, tm_mday=16, tm_hour=23, tm_min=3, tm_sec=58, tm_wday=0, tm_yday=75, tm_isdst=0))] is not None
```

### DIFF
--- a/python/ct/cert_analysis/validity.py
+++ b/python/ct/cert_analysis/validity.py
@@ -47,7 +47,7 @@ class CheckValidityNotBeforeFuture(object):
             now = datetime.datetime.utcnow()
             # time.mktime assumes localtime, but because both dates are in utc
             # comparision will be valid
-            if time.mktime(not_before) - time.mktime(now.timetuple()) > 0:
+            if time.mktime(not_before) - time.mktime(now.utctimetuple()) > 0:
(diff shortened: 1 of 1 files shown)


pphaneuf:
I see you signed the CLA now? You can keep working with this pull request, @googlebot should notice that you've signed it (I think you have to comment to cause a re-verification), and we'll be on our way!

linshunghuang:
Oops, at the moment I'd like to close this request since I'm still checking with my employer about contributing to open source. Sorry about that. I'll reopen once I get cleared and if this hasn't been fixed by then.

PR #582, opened 2015-03-17
Fix http threadpool size rather than based on #CPUs, and up the internal pool size a bit.

AlCutter:


pphaneuf:
Ben might not appreciate this too much, but...

LGTM. :wink:

PR #583, opened 2015-03-18
Follower adds entries to DB via non-libevent thread.

AlCutter:


cpp/fetcher/continuous_fetcher.cc
```diff
@@ -123,23 +125,18 @@ void ContinuousFetcherImpl::FetchDone(Task* task) {
...
 
   if (restart_fetch_) {
-    base_->Add(bind(&ContinuousFetcherImpl::FetchDelayDone, this, nullptr));
+    executor_->Add(bind(&ContinuousFetcherImpl::FetchDelayDone, this));
   } else {
-    base_->Delay(seconds(FLAGS_delay_between_fetches_seconds),
-                 new Task(bind(&ContinuousFetcherImpl::FetchDelayDone, this,
-                               _1),
-                          base_));
+    executor_->Add([this]() {
```
pphaneuf: I'd rather keep ```base_``` for the delay, this kind of sleeping in a thread pool callback is the kind of thing that's getting us in trouble... :wink:

pphaneuf:
I sat down at my computer meaning to do *this*, only to find your PR. :smiley:

cpp/fetcher/continuous_fetcher.cc
```diff
@@ -107,7 +109,7 @@ void ContinuousFetcherImpl::StartFetch(const unique_lock<mutex>& lock) {
   }
 
   fetch_task_.reset(
-      new Task(bind(&ContinuousFetcherImpl::FetchDone, this, _1), base_));
+      new Task(bind(&ContinuousFetcherImpl::FetchDone, this, _1), executor_));
```
pphaneuf: The executor for this task is the one that is used to do the disk writes, no need to change anything else.

AlCutter: Sorry, not thinking, done.

AlCutter: oke doke.

cpp/fetcher/continuous_fetcher.cc
AlCutter: Sorry, not thinking, done.

cpp/fetcher/continuous_fetcher.cc
```diff
@@ -2,6 +2,7 @@
 
 #include <gflags/gflags.h>
 #include <glog/logging.h>
+#include <thread>
```
pphaneuf: I think this header is a leftover?

pphaneuf:
LGTM, other than the leftover header. Thanks!

cpp/fetcher/continuous_fetcher.cc
AlCutter: dinged.

PR #585, opened 2015-03-18
Add a noddy script to get a freshly turned up etcd initialised.

AlCutter:


pphaneuf:
LGTM (and not like this is likely to break Travis!).

PR #588, opened 2015-03-18
Use a Request struct for EtcdClient::Get.

pphaneuf:
Also add support for recursive get, through that ```Request``` struct.

### DIFF
--- a/cpp/util/etcd.h
+++ b/cpp/util/etcd.h
@@ -46,6 +46,14 @@ class EtcdClient {
     bool deleted_;
   };
 
+  struct Request {
+    Request(const std::string& thekey) : key(thekey), recursive(false) {
+    }
+
+    std::string key;
+    bool recursive;
+  };
+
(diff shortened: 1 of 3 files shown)


cpp/util/etcd.h
```diff
@@ -46,6 +46,14 @@ class EtcdClient {
     bool deleted_;
   };
 
+  struct Request {
+    Request(const std::string& thekey) : key(thekey), recursive(false) {
+    }
+
+    std::string key;
+    bool recursive;
```
pphaneuf: This is where ```FakeEtcdClient``` subtly becomes broken: if you sent it a request with this... Hmm... something would happen, but probably not what you want. :smiley: 

But no one is sending such requests to it, so "it's okay"? I've worked out a few other changes past this commit, but the next thing I'm on is to fix up ```FakeEtcdClient``` to keep up... So this situation shouldn't be too long.

PR #590, opened 2015-03-19
Run the etcd populating requests on an actual etcd node.

AlCutter:


pphaneuf:
This only changes where the command is run? Makes much of a difference?

LGTM, in any case...

AlCutter:
It does indeed: when you're creating the log the etcd nodes are spun up early on in the process, and so some nodes don't yet have a local registry running. The cmd on line 36 can then fail if you happen to be running on one of these virgin nodes.  Doing it this way means that you should be on a host which is running a local image registry fo' sho'.

PR #592, opened 2015-03-19
Allow fetching of SCTs along with log entries.

AlCutter:
Currently, the follower code just downloads certs from other nodes in the log, but if a certificate is presented by a user to add-chain, unless that request just happens to hit the same node the cert was originally presented to, the local DB will only contain the cert - crucially, it will not contain the SCT and so a 'null' SCT is returned.

This PR addresses that.

### DIFF
--- a/cpp/client/async_log_client.cc
+++ b/cpp/client/async_log_client.cc
@@ -21,8 +21,8 @@ using ct::SignedCertificateTimestamp;
 using ct::SignedTreeHead;
 using std::back_inserter;
 using std::bind;
-using std::copy;
 using std::make_shared;
+using std::move;
 using std::placeholders::_1;
 using std::shared_ptr;
 using std::string;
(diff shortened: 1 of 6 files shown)


cpp/client/async_log_client.cc
```diff
@@ -163,39 +163,57 @@ void DoneGetEntries(UrlFetcher::Response* resp,
...
       Deserializer::DeserializePrecertChainEntry(
           extra_data.FromBase64(), log_entry.entry.mutable_precert_entry());
-    else
+    } else {
       LOG(FATAL) << "Don't understand entry type: "
                  << log_entry.leaf.timestamped_entry().entry_type();
+    }
 
-    new_entries.push_back(log_entry);
+    new_entries.emplace_back(std::move(log_entry));
```
pphaneuf: We're not in one of those nasty ```-inl.h``` files, you can add a ```using``` statement. :wink:

cpp/client/async_log_client.cc
```diff
@@ -163,39 +163,57 @@ void DoneGetEntries(UrlFetcher::Response* resp,
...
   }
 
   entries->reserve(entries->size() + new_entries.size());
-  copy(new_entries.begin(), new_entries.end(), back_inserter(*entries));
+  move(new_entries.begin(), new_entries.end(), back_inserter(*entries));
```
pphaneuf: Somehow, this one doesn't need ```std::``` qualification?! One of those wacky resolution rules kicking in, I guess...

pphaneuf:
LGTM

cpp/client/async_log_client.cc
AlCutter: done :)

cpp/client/async_log_client.cc
AlCutter: it has one anyway now :)

pphaneuf:
This won't pass Travis if you're not at least at 8f5dbfc3af16046b21b1ff405fe4381c18fa4502.

PR #594, opened 2015-03-19
Add more counters for latency monitoring

AlCutter:


pphaneuf:
This won't pass Travis if you're not at least at 8f5dbfc3af16046b21b1ff405fe4381c18fa4502.

cpp/monitoring/latency.h
```diff
@@ -0,0 +1,137 @@
...
+//
+//   // Here's an example where there's no one scope so can't easily use the
+//   // ScopedLatency helper.
+//   void FinishedDoingStuffCallback(const string& name,
+//                                   const system_clock::time_point started_at)
+//                                   {
+//      latency_by_name.RecordLatency(name, system_clock::now() - started_at);
+//   }
+template <class TimeUnit, class... LabelTypes>
+class Latency {
```
pphaneuf: How about calling this ```EventMetric```? Can apply to all sorts of things that can have a different value, for example the size of a certificate?

pphaneuf: Actually, ignore this, there's some template type ramifications, I think, that would make this kind of weird... Best leave it as-is for now and revisit later, as needed.

cpp/monitoring/latency.h
```diff
@@ -0,0 +1,137 @@
...
+  DISALLOW_COPY_AND_ASSIGN(ScopedLatency);
+};
+
+
+template <class TimeUnit, class... LabelTypes>
+Latency<TimeUnit, LabelTypes...>::Latency(
+    const std::string& base_name,
+    const typename NameType<LabelTypes>::name&... label_names,
+    const std::string& help)
+    : latencies_(Counter<LabelTypes...>::New(base_name, label_names..., help)),
```
pphaneuf: Wonder if it wouldn't be a good idea to also have a suffix here (say, "_overall_sum", for old time's sake), so that a computed variable built on that could have that name? Not sure how much sense that makes with Prometheus, though...

cpp/server/ct-server.cc
```diff
@@ -132,6 +137,22 @@ Gauge<>* latest_local_tree_size_gauge =
...
+                                               "Total runtime of sequencer");
+Latency<milliseconds> sequencer_delete_latency_ms(
+    "sequencer_delete_latency_ms",
+    "Total time spent deleting entries by sequencer");
+Latency<milliseconds> sequencer_sequence_latency_ms(
+    "sequencer_sequence_latency_ms",
+    "Total time spent sequencing entries by sequencer");
+
+Counter<>* signer_total_runs =
+    Counter<>::New("signer_total_runs", "Total number of signer runs.");
```
pphaneuf: Isn't that included as part of ```signer_run_latency_ms```? But looking below, this is really the number of *successful* runs, could just update the docstring.

cpp/log/sqlite_db-inl.h
```diff
@@ -21,7 +23,14 @@ DEFINE_bool(sqlite_synchronous_off, false,
 namespace {
 
 
+static cert_trans::Latency<std::chrono::milliseconds, std::string>
+    latency_by_op_ms("database_latency_by_operation_ms", "operation",
```
pphaneuf: Maybe have "sqlite" somewhere in there, to reflect it's in that file? There might be similar counters for FileDB?

pphaneuf: By which I mean, there *are*, and it's a bit confusing having two variables with the same name (I'm not even 100% on whether that's allowed, or if it should even be).

pphaneuf:
Done with this PR for now.

cpp/monitoring/latency.h
AlCutter: Hmm, well this knows about `TimeUnit`s and `durations`s etc. which is quite handy, but I *could* factor out the common *"keep a count and a total"* stuff into an EventMetric class and compose that into here instead.

cpp/monitoring/latency.h
```diff
@@ -0,0 +1,137 @@
...
+
+// Helper class to automatically calculate and record latency.
+// Measures the duration between its construction and destruction times, and
+// automatically registers that with the Latency<> class which created it.
+class ScopedLatency {
+ public:
+  ScopedLatency(ScopedLatency&& other) = default;
+
+  ~ScopedLatency() {
+    record_latency_(std::chrono::system_clock::now() - start_);
```
pphaneuf: I had forgotten one comment: wouldn't this whole thing be better off using ```std::chrono::steady_clock``` instead?

cpp/monitoring/latency.h
pphaneuf: Yeah, that's what I meant, it remembers what unit the variable is in for you.

I'd say leave it like this for now, when we need more, we can refactor it. If it had been already exactly the way an hypothetical ```EventMetric``` would have been, I'd have pushed to "just rename it", but it isn't. :smiley:

cpp/monitoring/latency.h
AlCutter: Not sure either, but doesn't cost much to add and seems like a good idea.
Done.

cpp/monitoring/latency.h
AlCutter: I've already done it, so... :)

cpp/monitoring/latency.h
pphaneuf: Did you, really? :smiley:

cpp/log/sqlite_db-inl.h
AlCutter: Yeah, that was kinda deliberate; I figured that the server is only going to use one of these, and the operations are the same for both - this would save you from having to change all your monitoring if you switched between them.

Really this should be done at the `database.cc` level and rename all the virtual methods, but for now I'll just rename the metric names and we'll do the `database.cc` thing as/when.

pphaneuf: And the two variables won't fight with each other when it comes the time to generate the protobuf? If so, SGTM.

cpp/monitoring/latency.h
AlCutter: Yes.

AlCutter:
PTAL

pphaneuf:
LGTM

PR #595, opened 2015-03-20
Change the etcd watch support to use EtcdClient::Get.

pphaneuf:
This is instead of using the internal ```EtcdClient::Generic``` method.

### DIFF
--- a/cpp/util/etcd.cc
+++ b/cpp/util/etcd.cc
@@ -462,12 +462,12 @@ void EtcdClient::WatchInitialGetDone(WatchState* state, GetResponse* resp,
 }
 
 
-void EtcdClient::WatchRequestDone(WatchState* state, GenericResponse* gen_resp,
+void EtcdClient::WatchRequestDone(WatchState* state, GetResponse* get_resp,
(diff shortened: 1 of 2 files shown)


cpp/util/etcd.cc
```diff
@@ -506,31 +506,16 @@ void EtcdClient::WatchRequestDone(WatchState* state, GenericResponse* gen_resp,
...
 
-    if (!node.ValueOrDie().deleted_) {
-      state->known_keys_[node.ValueOrDie().key_] =
-          node.ValueOrDie().modified_index_;
+    if (!get_resp->node.deleted_) {
+      state->known_keys_[get_resp->node.key_] = get_resp->node.modified_index_;
     } else {
-      LOG(INFO) << "erased key: " << node.ValueOrDie().key_;
-      state->known_keys_.erase(node.ValueOrDie().key_);
+      LOG(INFO) << "erased key: " << get_resp->node.key_;
```
AlCutter: Should probably demote this to a `VLOG` since you're here :)

PR #601, opened 2015-03-25
Switch to vanilla GCE.

AlCutter:
So that we have more control over the instance configs, in particular persistent disk configs.


pphaneuf:
LGTM, albeit a bit sad that our nice tools are letting us down...

PR #602, opened 2015-03-25
It's better if instances actually know which host they're on.

AlCutter:
Rather than all believing they're running on the same container.


pphaneuf:
LGTM

PR #603, opened 2015-03-25
Speed up SQLite.

AlCutter:
This speeds up SQLite operations considerably.

### DIFF
--- a/cpp/log/sqlite_db-inl.h
+++ b/cpp/log/sqlite_db-inl.h
@@ -13,12 +13,32 @@
 #include "monitoring/latency.h"
 #include "util/util.h"
 
+// Several of these flags pass their value directly through to SQLite PRAGMA
+// statements, see the SQLite documentation
+// (https://www.sqlite.org/pragma.html) for a description of the various
+// values available and the implications they have.
+
 // TODO(pphaneuf): For now, just a flag, but ideally, when adding a
 // new node, it would do an initial load of its local database with
 // "synchronous" set to OFF, then put it back before starting normal
 // operation.
-DEFINE_bool(sqlite_synchronous_off, false,
-            "whether to set SQLite's \"synchronous\" to OFF");
+DEFINE_string(sqlite_synchronous_mode, "FULL",
+              "Which SQLite synchronous option to use, see SQLite pragma "
+              "documentation for details.");
+
+DEFINE_string(sqlite_journal_mode, "WAL",
+              "Which SQLite journal_mode option to use, see SQLite pragma "
+              "documentation for defails.");
+
+DEFINE_int32(sqlite_cache_size, 100000,
+             "Number of 1KB btree pages to keep in memory.");
+
+DEFINE_bool(sqlite_batch_into_transactions, true,
+            "Whether to batch operations into transactions behind the "
+            "scenes.");
+DEFINE_int32(sqlite_transaction_batch_size, 400,
+             "Max number of operations to batch into one transaction.");
+
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


cpp/log/sqlite_db-inl.h
```diff
@@ -17,8 +17,26 @@
...
+DEFINE_string(sqlite_temp_store, "DISK",
+              "Which SQLite temp_store option to use.");
+
+DEFINE_int32(sqlite_cache_size, 100000,
+             "Number of 1KB btree pages to keep in memory.");
+
+DEFINE_bool(sqlite_batch_into_transactions, true,
+            "Whether to batch operations into transactions behind the scenes. "
+            "Transactions will always be committed when new STHs are written. "
+            "It's recommended to run with the FULL synchronous mode when "
```
pphaneuf: I'm not sure this recommendation makes much sense? I think the synchronization mode only changes what happens when a transaction is completed, so it doesn't really matter what it does, it'll just be happening less often.

cpp/util/testing.cc
```diff
@@ -5,6 +5,8 @@
 #include <glog/logging.h>
 #include <gtest/gtest.h>
 
+#include "log/logged_certificate.h"
```
pphaneuf: Huh?

cpp/log/sqlite_db-inl.h
```diff
@@ -73,13 +91,50 @@ sqlite3* SQLiteOpen(const std::string& dbfile) {
...
+    if (FLAGS_sqlite_batch_into_transactions) {
+      LOG(WARNING) << "SQLite running with batched transactions, you should "
+                   << "set sqlite_synchronous_mode = FULL !";
+    }
   }
 
+  {
+    std::ostringstream oss;
+    oss << "PRAGMA journal_mode = " << FLAGS_sqlite_journal_mode;
+    sqlite::Statement statement(db_, oss.str().c_str());
```
pphaneuf: Ugh, ```sqlite::Statement``` should take a ```const string&``` for the statement, so you can build it easily, this is rather silly... For later, though.

cpp/log/sqlite_db-inl.h
```diff
@@ -73,13 +91,50 @@ sqlite3* SQLiteOpen(const std::string& dbfile) {
...
+    CHECK_EQ(SQLITE_ROW, statement.Step());
+    std::string mode;
+    statement.GetBlob(0, &mode);
+    CHECK_STRCASEEQ(mode.c_str(), FLAGS_sqlite_journal_mode.c_str());
```
pphaneuf: You need ```.c_str()``` with this? I've never used it, just wondering...

cpp/log/sqlite_db-inl.h
```diff
@@ -17,8 +17,26 @@
...
 // operation.
-DEFINE_bool(sqlite_synchronous_off, false,
-            "whether to set SQLite's \"synchronous\" to OFF");
+DEFINE_string(sqlite_synchronous_mode, "FULL",
+              "While SQLite synchronous option to use");
+
+DEFINE_string(sqlite_journal_mode, "WAL",
+              "Which SQLite journal_mode option to use.");
+
+DEFINE_string(sqlite_temp_store, "DISK",
```
pphaneuf: If that one is turning out to be pointless, might as well remove it?

pphaneuf:
Done with this round of review.

cpp/log/sqlite_db-inl.h
AlCutter: I suppose it's up to the operator how lucky they're feeling.
What I was trying to get at here was that the transaction batching amortises the cost of FULL synchronisation, and so there's less/no reason not to back away from it.
But maybe this isn't place; I'll remove this from here, and we should have some docs which discuss these flags.

cpp/util/testing.cc
AlCutter: gnash gnash, that shouldn't be here. Removed.

cpp/log/sqlite_db-inl.h
AlCutter: Yup, it's just a wrapper around `strcasecmp`

cpp/log/sqlite_db-inl.h
AlCutter: Removed

AlCutter: Removed

cpp/util/testing.cc
```diff
@@ -5,6 +5,7 @@
 #include <glog/logging.h>
 #include <gtest/gtest.h>
 
+
```
pphaneuf: Still a stray extra line?

cpp/server/ct-server.cc
```diff
@@ -382,9 +382,10 @@ int main(int argc, char* argv[]) {
   CHECK(checker.LoadTrustedCertificates(FLAGS_trusted_cert_file))
       << "Could not load CA certs from " << FLAGS_trusted_cert_file;
 
-  if (FLAGS_sqlite_db == "")
+  if (FLAGS_sqlite_db == "") {
```
pphaneuf: While you're there, make this ```FLAGS_sqlite_db.empty()```? :wink:

cpp/server/ct-server.cc
```diff
@@ -395,13 +396,14 @@ int main(int argc, char* argv[]) {
 
   Database<LoggedCertificate>* db;
 
-  if (FLAGS_sqlite_db != "")
+  if (FLAGS_sqlite_db != "") {
```
pphaneuf: Similar here, ```!FLAGS_sqlite_db.empty()```?

pphaneuf:
LGTM, leaving those last things to your discretion, no need to review them separately.

cpp/log/sqlite_db-inl.h
```diff
@@ -17,8 +17,21 @@
 // new node, it would do an initial load of its local database with
 // "synchronous" set to OFF, then put it back before starting normal
 // operation.
-DEFINE_bool(sqlite_synchronous_off, false,
-            "whether to set SQLite's \"synchronous\" to OFF");
+DEFINE_string(sqlite_synchronous_mode, "FULL",
+              "While SQLite synchronous option to use");
```
benlaurie: s/While/Which/?

benlaurie: Also, what are the options?

cpp/log/sqlite_db-inl.h
```diff
@@ -17,8 +17,21 @@
...
+
+DEFINE_string(sqlite_journal_mode, "WAL",
+              "Which SQLite journal_mode option to use.");
```
benlaurie: Likewise, what are the options?

cpp/log/sqlite_db-inl.h
AlCutter: Me fail English? That's unpossible!

Options are documented here: https://www.sqlite.org/pragma.html

I've added a comment and a hint in the flags' help strings

cpp/log/sqlite_db-inl.h
AlCutter: same here.

PR #604, opened 2015-03-26
Add a helper function to delete etcd keys concurrently.

pphaneuf:


cpp/util/etcd_delete.cc
```diff
@@ -0,0 +1,127 @@
...
+ private:
+  void RequestDone(Task* child_task);
+  void StartNextRequest(unique_lock<mutex>&& lock);
+
+  EtcdClient* const client_;
+  Task* const task_;
+  mutex mutex_;
+  int outstanding_;
+  int next_;
+  const vector<pair<string, int64_t> > keys_;
```
AlCutter: no need for the old school extra space in `> >` anymore :)

cpp/util/etcd_delete.cc
```diff
@@ -0,0 +1,127 @@
...
+    // In case the task uses an inline executor.
+    lock.unlock();
+    task_->Return(Status::CANCELLED);
+    return;
+  }
+
+  while (outstanding_ < FLAGS_etcd_delete_concurrency &&
+         next_ < keys_.size()) {
+    CHECK(lock.owns_lock());
+    const pair<string, int64_t>& key(keys_[next_]);
```
AlCutter: You could make `next_` a `vector<pair<string, int64_t>>::const_iterator` and then not need this line. (more for clarity if you think it'd help, I realise this is O(1) anyway.)

cpp/util/etcd_delete.cc
```diff
@@ -0,0 +1,127 @@
...
+    ++next_;
+    ++outstanding_;
+
+    // In case the task uses an inline executor.
+    lock.unlock();
```
AlCutter: case for a `ScopedUnlocker` ? :)

cpp/util/etcd_delete.cc
pphaneuf: Done.

cpp/util/etcd_delete.cc
pphaneuf: Can't really get away from it, but I replaced ```next_``` with ```it_```.

cpp/util/etcd_delete.cc
pphaneuf: A weak one... I'll pass, if you'll allow? :smiley:

PR #607, opened 2015-03-27
Move free functions in fake_etcd.cc to the anonymous namespace.

pphaneuf:


pphaneuf:
Rationale for this, lest people think I'm a bit odd: I'm doing big refactoring in there, and I'd like the compiler to tell me when something isn't used anymore.

PR #608, opened 2015-03-27
Update bench_etcd.

pphaneuf:
Avoid use of ```CreateInQueue``` (removing it), freshen up in general.

### DIFF
--- a/cpp/util/bench_etcd.cc
+++ b/cpp/util/bench_etcd.cc
@@ -2,7 +2,6 @@
 #include <functional>
 #include <gflags/gflags.h>
 #include <glog/logging.h>
-#include <map>
 #include <memory>
 #include <string>
 #include <thread>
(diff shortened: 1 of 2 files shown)


cpp/util/bench_etcd.cc
```diff
@@ -93,14 +110,13 @@ int main(int argc, char* argv[]) {
   CHECK_GE(FLAGS_bytes_per_request, 0);
   CHECK_GT(FLAGS_num_threads, 0);
 
-  vector<thread*> threads;
-  for (int i = FLAGS_num_threads; i > 0; --i)
-    threads.push_back(new thread(&test_etcd));
+  vector<thread> threads;
+  for (int i = 0; i < FLAGS_num_threads; ++i)
```
AlCutter: `{...}`

PR #609, opened 2015-03-27
etcd watch tweaks

pphaneuf:
Remove some pointless bouncing around executor (when we're already on the right one), and it looks like the only case we used ```libevent::Base::Delay``` was when handling a timeout, we can just retry immediately, which is much simpler.


pphaneuf:
Note that this is in two commits, for your reviewing convenience. :wink:

PR #611, opened 2015-03-27
Send the etcd index with Get as much as we can.

pphaneuf:


pphaneuf:
Now also with a test to catch the problem! I validated the test (it fails without the fix), of course.

Issue #605, opened 2015-03-26
cleaning should be concurrent to sequencing


pphaneuf:
In ```super_duper```, the sequencing and cleaning of replicated entries is done one after the other (in a loop). It's very important to keep etcd as "light" as possible, and not have too many entries there, so both sequencing and cleaning need to happen more aggressively.

The main reason for having them done serially is because they both update the sequence number mapping key, but if the sequencer took care of maintaining the mapping, the cleaning could just proceed as quickly as possible, at the same time.

pphaneuf:
@AlCutter, didn't you do this?

AlCutter:
Yes!

PR #619, opened 2015-03-30
Cleanup etcd entries in parallel.

AlCutter:


cpp/server/ct-server.cc
```diff
@@ -224,34 +225,50 @@ static const bool sign_dummy =
...
                      const MasterElection* election) {
+  CHECK_NOTNULL(tree_signer);
+  CHECK_NOTNULL(election);
   const steady_clock::duration period(
       (seconds(FLAGS_sequencing_frequency_seconds)));
   steady_clock::time_point target_run_time(steady_clock::now());
 
   while (true) {
     if (election->IsMaster()) {
-      const ScopedLatency sequence_run_latency(
```
pphaneuf: Why lose this latency metric?

pphaneuf:
LGTM

cpp/server/ct-server.cc
AlCutter: Just because it'd now be measuring the same thing as `sequencer_delete_latency_ms` give or take a couple of µs

AlCutter:
Fixes #605

PR #631, opened 2015-04-01
Fix #627 - Use utctimetuple() instead of timetuple() in validity analysis

### LINKED ISSUE #627: validity_test.py is failing
Error message:
```
PYTHONPATH=:. ./ct/cert_analysis/validity_test.py
F....
FAIL: test_not_before_regular (__main__.ValidityTest)
  File "./ct/cert_analysis/validity_test.py", line 15, in test_not_before_regular
    self.assertIsNone(result)
AssertionError: [NotBeforeInFuture('Validity: not before is in future', None, time.struct_time(tm_year=2015, tm_mon=3, tm_mday=16, tm_hour=23, tm_min=3, tm_sec=58, tm_wday=0, tm_yday=75, tm_isdst=0))] is not None
```
Looks like Line 50 of validity.py (https://github.com/google/certificate-transparency/blob/c99d0b75cb33156cb922b6544fdf2e6e216aa597/python/ct/cert_analysis/validity.py#L50) compares dates with different `DST` flags. Should use `utctimetuple()` instead of `timetuple()`.

phad:
where timestamps with different DST flags could be compared.  This caused test failures, possibly related to the DST transition or because of the TZ of the computer running the test.

### DIFF
--- a/python/ct/cert_analysis/validity.py
+++ b/python/ct/cert_analysis/validity.py
@@ -47,7 +47,7 @@ class CheckValidityNotBeforeFuture(object):
             now = datetime.datetime.utcnow()
             # time.mktime assumes localtime, but because both dates are in utc
             # comparision will be valid
-            if time.mktime(not_before) - time.mktime(now.timetuple()) > 0:
+            if time.mktime(not_before) - time.mktime(now.utctimetuple()) > 0:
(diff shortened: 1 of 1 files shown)


pphaneuf:
LGTM

PR #638, opened 2015-04-08
Add a test for EtcdDeleteKeys.

pphaneuf:
Also fixed a bug found by the new test, and clarified one of the error cases (giving it a null task). Verified that the bug fixed by #633 would have been exposed by the test.

### DIFF
--- a/cpp/util/mock_etcd.h
+++ b/cpp/util/mock_etcd.h
@@ -1,0 +1,51 @@
+#ifndef CERT_TRANS_UTIL_MOCK_ETCD_H_
+#define CERT_TRANS_UTIL_MOCK_ETCD_H_
+
+#include <gmock/gmock.h>
+
+#include "util/etcd.h"
+
+namespace cert_trans {
+
+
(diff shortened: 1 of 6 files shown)


cpp/util/mock_etcd.h
```diff
@@ -0,0 +1,51 @@
...
+                    const std::chrono::seconds& ttl, Response* resp,
+                    util::Task* task));
+  MOCK_METHOD3(Delete, void(const std::string& key,
+                            const int64_t current_index, util::Task* task));
+  MOCK_METHOD3(Watch, void(const std::string& key, const WatchCallback& cb,
+                           util::Task* task));
+
+  // TODO(pphaneuf): Shouldn't be used, and shouldn't be there, but
+  // has to be for now.
+  MOCK_METHOD5(Generic, void(const std::string& key,
```
pphaneuf: Didn't think of it before I sent this PR, but I could make this one private? It needs to be there so that the vtable gets emitted...

PR #640, opened 2015-04-08
Wait for prometheus node(s) to be up before updating config.

AlCutter:


cloud/google/util.sh
```diff
@@ -20,19 +19,12 @@ function WaitForStatus() {
...
 
-function WaitForPod() {
-  set +e
-  NODE_LABEL="name=$1"
-  echo "Waiting for $1"
-
-  until [ "${STATE}" != "" ]; do
+function WaitMachineUp() {
+  echo "Waiting for ${1}"
+  until ${GCLOUD} compute ssh ${1} -c "logout"; do
```
pphaneuf: Doesn't this spin quite fast until the machine is up? Is there a ```usleep```?

pphaneuf:
LGTM, although possibly moderating that busy-loop might make sense?

PR #641, opened 2015-04-08
Fake is not real enough

pphaneuf:
Split into a few separate commits, to ease reviewing.


pphaneuf:
With these commits, ```fake_etcd_test``` passes both normally and with the real etcd (using the ```--etcd``` flag). It's rather slow with the real etcd, because ```UrlFetcher``` doesn't pro-actively support cancellation, but it's still correct.

pphaneuf:
I addressed these two comments.

pphaneuf:
This is part of #645, by the way.

PR #646, opened 2015-04-09
Use util::Status in frontend.cc and frontend_signer.cc

AlCutter:
... and don't CHECK fail when we can't add an entry to the consistent store, just return an error instead.

### DIFF
--- a/cpp/log/frontend.cc
+++ b/cpp/log/frontend.cc
@@ -6,7 +6,9 @@
 #include "log/cert.h"
 #include "log/cert_submission_handler.h"
 #include "log/frontend_signer.h"
+#include "monitoring/event_metric.h"
 #include "proto/ct.pb.h"
+#include "util/status.h"
 
 using cert_trans::CertChain;
 using cert_trans::PreCertChain;
@@ -15,9 +17,19 @@ using ct::SignedCertificateTimestamp;
 using std::string;
 using std::lock_guard;
 using std::mutex;
+using util::Status;
+
+namespace {
+
+static cert_trans::EventMetric<std::string, std::string>
+    submission_status_metric(
+        "submission_status", "entry_type", "status",
+        "Submission status totals broken down by entry type and status code.");
+
+}  // namespace
 
 Frontend::Frontend(CertSubmissionHandler* handler, FrontendSigner* signer)
-    : handler_(handler), signer_(signer), stats_() {
+    : handler_(handler), signer_(signer) {
 }
 
 Frontend::~Frontend() {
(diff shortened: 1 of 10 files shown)


cpp/log/frontend_signer.cc
```diff
@@ -47,7 +48,8 @@ FrontendSigner::SubmitResult FrontendSigner::QueueEntry(
     if (sct != nullptr) {
       *sct = logged.sct();
     }
-    return DUPLICATE;
+    return Status(util::error::ALREADY_EXISTS,
+                  "Entry already exists in Database.");
```
pphaneuf: Those shouldn't be sentences, IMHO (no starting capital, no period). But I'm not going to be harsh on this, if you feel strongly about it... ;-)

cpp/log/frontend_signer.cc
```diff
@@ -70,13 +72,7 @@ FrontendSigner::SubmitResult FrontendSigner::QueueEntry(
     *sct = new_logged.sct();
   }
 
-  if (status.CanonicalCode() == util::error::ALREADY_EXISTS) {
-    return DUPLICATE;
-  } else {
-    // TODO(alcutter): really need to propagate failures back up!
-    CHECK(status.ok());
```
pphaneuf: This is the key change, right?

cpp/log/frontend.cc
```diff
@@ -111,95 +105,40 @@ std::string Frontend::SubmitResultString(SubmitResult result) {
...
-      ++stats_.x509_verify_errors;
-      break;
-    case INTERNAL_ERROR:
-      ++stats_.internal_errors;
-      break;
-    case PRECERT_CHAIN_NOT_WELL_FORMED:
-      LOG(FATAL) << "invalid PRECERT_CHAIN_NOT_WELL_FORMED on an X509 certificate";
-    default:
-      LOG(FATAL) << "unknown SubmitResult enum value: " << result;
+void Frontend::UpdateStats(ct::LogEntryType type, const Status& status) {
```
pphaneuf: Just a suggestion: make this return the status, so you could use it inline above, like this:

    return UpdateStats(entry.type(), GetSubmitError(pre_result));

cpp/log/frontend_signer.cc
pphaneuf: And, more importantly, this just gets sent back to the client, rather than setting things on fire? :smiley:

pphaneuf:
Done reviewing.

pphaneuf:
If all you change is the sentence in the ```util::Status```, LGTM, feel free to merge. If you do the "inline ```UpdateStats```" thing, there's a bit more involved, might be worth another look?

cpp/log/frontend.cc
```diff
@@ -111,95 +105,40 @@ std::string Frontend::SubmitResultString(SubmitResult result) {
...
-      break;
+      return util::Status(util::error::INVALID_ARGUMENT,
+                          "Invalid PEM encoded chain.");
     case CertSubmissionHandler::SUBMISSION_TOO_LONG:
-      submit_result = SUBMISSION_TOO_LONG;
-      break;
+      return util::Status(util::error::INVALID_ARGUMENT,
+                          "Submission too long.");
     case CertSubmissionHandler::INVALID_CERTIFICATE_CHAIN:
+      return util::Status(util::error::FAILED_PRECONDITION,
```
pphaneuf: That might be more ```INVALID_ARGUMENT```? Going by this:

https://github.com/google/certificate-transparency/blob/super_duper/cpp/util/status.h#L75

cpp/log/frontend.cc
```diff
@@ -111,95 +105,40 @@ std::string Frontend::SubmitResultString(SubmitResult result) {
...
+                          "Invalid certificate chain.");
     case CertSubmissionHandler::UNKNOWN_ROOT:
-      submit_result = CERTIFICATE_VERIFY_ERROR;
-      break;
+      return util::Status(util::error::FAILED_PRECONDITION, "Unknown root.");
     case CertSubmissionHandler::PRECERT_CHAIN_NOT_WELL_FORMED:
-      submit_result = PRECERT_CHAIN_NOT_WELL_FORMED;
-      break;
+      return util::Status(util::error::FAILED_PRECONDITION,
```
pphaneuf: Same here?

For ```UNKNOWN_ROOT```, I'd guess that "adding the root" is something that could be done (the request might otherwise be valid, but not acceptable in the current state of the server), so that's ok.

cpp/log/frontend_signer.cc
AlCutter: Done.
Sorry:
done

cpp/log/frontend_signer.cc
AlCutter: Right!

cpp/log/frontend.cc
AlCutter: Nice idea, thanks.

cpp/log/frontend.cc
AlCutter: I was going on the the principle that having a valid chain was a precondition to being able to add it to a log, but it does seem to fit better into the `INVALID_ARGUMENT` description.

cpp/log/frontend.cc
AlCutter: Ditto

pphaneuf:
LGTM

PR #650, opened 2015-04-10
Don't accidentally clip root certs, or blow up when there's no chain.

AlCutter:
This worked previously because the inclusion of the root is optional in the chain.


pphaneuf:
Ah, classic... LGTM. :smiley:

PR #651, opened 2015-04-10
Clean up old entries more aggressively.

pphaneuf:
Do the cleaning at least every 10 seconds, instead of starting on the next 10 seconds interval.


pphaneuf:
Never mind, this is actually not aggressive enough (in our test, it was taking about 6 seconds to run, so this commit makes just about no difference).

Issue #645, opened 2015-04-09
make code using the etcd client more testable


pphaneuf:
We have some very sensitive code (like the election algorithm) based on the etcd client, we should make sure that this is all as reliable as possible. For one, test the etcd client itself with ```MockUrlFetcher``` thoroughly (we've got a start on this, but should be improved), and then test the code using the etcd client with ```MockEtcdClient``` thoroughly (as I write this, only ```EtcdDeleteKeys```has gone through this, it was rather simple, but doing it did find one or two issues!).

pphaneuf:
Pull request #641 is serving to make sure that the ```FakeEtcdClient``` we use in a number of tests tracks the behaviour of the real etcd implementation, by running its test against the real thing. This ensures that we're not having a test that checks we maintaining a bug properly, for example (I did find some differences in behaviour while doing this pull request).

Another thing I've started on is exposing the options needed to implement ```EtcdClient::Watch``` in ```EtcdClient::Get```. The goal of this is that the watch logic (which is very important for things like the election algorithm) could then be pulled out into its own module, which could then be tested in detail. It would also avoid having a wholly separate implementation of the watch logic in ```FakeEtcdClient::Watch``` (as is the case now). The current blocker to this is that ```FakeEtcdClient::Get``` does not support the "wait" option needed.

PR #652, opened 2015-04-10
Finish replacing the guts of FakeEtcdClient.

pphaneuf:
Still a few things unimplemented, but the mocking/faking is all done at the top level now, which is easier for tests.


pphaneuf:
This is part of #645.

PR #658, opened 2015-04-10
Make FakeEtcdClient prune TTL'd entries properly.

pphaneuf:


cpp/util/fake_etcd.h
```diff
@@ -15,9 +17,9 @@ namespace cert_trans {
 
 class FakeEtcdClient : public EtcdClient {
  public:
-  FakeEtcdClient();
+  FakeEtcdClient(libevent::Base* base);
```
AlCutter: `explicit`

AlCutter:
LGTM w/ one nit.

pphaneuf:
Updated and squashed it (only diff with previous is the "explicit").

PR #668, opened 2015-04-14
Add a GetStoreStats() method to EtcdClient.

AlCutter:
Exposes the json formatted stats under etcd's "/v2/stats/store" key.

### DIFF
--- a/cpp/util/etcd.cc
+++ b/cpp/util/etcd.cc
@@ -47,6 +47,19 @@ namespace cert_trans {
 
 namespace {
 
+const char* kStoreStats[] = {"setsFail", "getsSuccess", "watchers",
+                             "expireCount", "createFail", "setsSuccess",
+                             "compareAndDeleteFail", "createSuccess",
+                             "deleteFail", "compareAndSwapSuccess",
+                             "compareAndSwapFail", "compareAndDeleteSuccess",
+                             "updateFail", "deleteSuccess", "updateSuccess",
+                             "getsFail"};
+
+const char kKeysSpace[] = "/v2/keys";
+const char kStatsSpace[] = "/v2/stats";
+
+const char kStoreStatsKey[] = "/store";
+
 
 string MessageFromJsonStatus(const shared_ptr<JsonObject>& json) {
   string message;
@@ -180,6 +193,42 @@ void GetRequestDone(const string& keyname, EtcdClient::GetResponse* resp,
 }
 
 
+void CopyStat(const string& key, const JsonObject& from,
+              map<string, int64_t>* to) {
+  CHECK_NOTNULL(to);
+  const JsonInt stat(from, key.c_str());
+  if (!stat.Ok()) {
+    LOG(WARNING) << "Failed to find stat " << key;
+    return;
+  }
+  (*to)[key] = stat.Value();
+}
+
+
+void GetStoreStatsRequestDone(EtcdClient::StatsResponse* resp,
+                              Task* parent_task,
+                              EtcdClient::GenericResponse* gen_resp,
+                              Task* task) {
+  *resp = EtcdClient::StatsResponse();
+  resp->etcd_index = gen_resp->etcd_index;
+  if (!task->status().ok()) {
+    parent_task->Return(task->status());
+    return;
+  }
+
+  if (!gen_resp->json_body->Ok()) {
+    parent_task->Return(Status(util::error::FAILED_PRECONDITION,
+                               "Invalid JSON: json_body not Ok."));
+    return;
+  }
+
+  for (const auto& stat : kStoreStats) {
+    CopyStat(stat, *gen_resp->json_body, &resp->stats);
+  }
+  parent_task->Return();
+}
+
+
 void CreateRequestDone(EtcdClient::Response* resp, Task* parent_task,
                        EtcdClient::GenericResponse* gen_resp, Task* task) {
   if (!task->status().ok()) {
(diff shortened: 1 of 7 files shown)


cpp/util/etcd.cc
```diff
@@ -756,7 +804,18 @@ void EtcdClient::Delete(const string& key, const int64_t current_index,
...
+  Generic(key, kKeysSpace, params, UrlFetcher::Verb::DELETE, gen_resp, task);
+}
+
+
+void EtcdClient::GetStoreStats(StatsResponse* resp, Task* task) {
+  map<string, string> params;
+  GenericResponse* const gen_resp(new GenericResponse);
+  task->DeleteWhenDone(gen_resp);
+
+  Generic("/store", kStatsSpace, params, UrlFetcher::Verb::GET, gen_resp,
```
pphaneuf: Make a constant for that "key"?

cpp/util/etcd.cc
```diff
@@ -180,6 +191,42 @@ void GetRequestDone(const string& keyname, EtcdClient::GetResponse* resp,
...
 
 
+void CopyStat(const string& key, const JsonObject& from,
+              map<string, int64_t>* to) {
+  CHECK_NOTNULL(to);
+  const JsonInt stat(from, key.c_str());
+  if (!stat.Ok()) {
+    LOG(WARNING) << "Failed to find stat " << key;
+  }
+  (*to)[key] = stat.Value();
```
pphaneuf: Should you set it if it doesn't exist? I'm not sure what this does with a null...

cpp/util/etcd.cc
```diff
@@ -180,6 +191,42 @@ void GetRequestDone(const string& keyname, EtcdClient::GetResponse* resp,
...
+
+
+void GetStoreStatsRequestDone(EtcdClient::StatsResponse* resp,
+                              Task* parent_task,
+                              EtcdClient::GenericResponse* gen_resp,
+                              Task* task) {
+  *resp = EtcdClient::StatsResponse();
+  resp->etcd_index = gen_resp->etcd_index;
+  if (!task->status().ok()) {
+    parent_task->Return(Status(task->status().CanonicalCode(),
```
pphaneuf: If you're not modifying the status, you should just return it directly:

    parent_task->Return(task->status());

cpp/util/fake_etcd.cc
```diff
@@ -178,26 +181,37 @@ void FakeEtcdClient::InternalPut(const string& key, const string& value,
...
+}
+
+
 void FakeEtcdClient::Create(const string& key, const string& value,
                             Response* resp, Task* task) {
-  return InternalPut(key, value, system_clock::time_point::max(), true, -1,
-                     resp, task);
+  InternalPut(key, value, system_clock::time_point::max(), true, -1, resp,
+              task);
+  UpdateOperationStats("create", task);
```
pphaneuf: This isn't safe, the ```task``` could have already been deleted... You could register it as a cleanup callback on the task, though, those get run after ```Return``` has been called, but before the task is gone, so you could check the status on it. Plus, you ensure it getting called, no matter what crazy code path is taken.

cpp/util/etcd.cc
AlCutter: done

cpp/util/fake_etcd.cc
```diff
@@ -178,26 +181,37 @@ void FakeEtcdClient::InternalPut(const string& key, const string& value,
 }
 
 
+void FakeEtcdClient::UpdateOperationStats(const string& op, const Task* task) {
+  CHECK_NOTNULL(task);
+  if (!task->IsActive()) {
```
pphaneuf: I'd fold the ```CHECK_NOTNULL``` unto this line, personally. Your call...

cpp/util/etcd.cc
AlCutter: whoops, missing `return` there, thanks.

cpp/util/etcd.cc
AlCutter: Heh, this was copied from `GetRequestDone`, and then later modified.
Fixed.

pphaneuf:
Done for now, and your first update looks good, just the fake to fix up...

cpp/util/fake_etcd.cc
AlCutter: Nice, thanks, done.

cpp/util/fake_etcd.cc
AlCutter: I kinda prefer them separated out; scans a bit nicer I think.

cpp/util/fake_etcd.cc
```diff
@@ -178,26 +181,40 @@ void FakeEtcdClient::InternalPut(const string& key, const string& value,
...
+  }
+}
+
+
 void FakeEtcdClient::Create(const string& key, const string& value,
                             Response* resp, Task* task) {
-  return InternalPut(key, value, system_clock::time_point::max(), true, -1,
-                     resp, task);
+  task->CleanupWhenDone(
+      std::bind(&FakeEtcdClient::UpdateOperationStats, this, "create", task));
```
pphaneuf: Take off these ```std::```, this is not an -inl.h file, there are rules! :wink:

pphaneuf:
Other than those spurious ```std::```, LGTM.

cpp/util/fake_etcd.cc
AlCutter: Rules, I think you'll find, are there to be broken.

PR #669, opened 2015-04-14
Add support for recursive get in FakeEtcdClient.

pphaneuf:


pphaneuf:
Ugh, that failure is an actual bug. Stand by...

pphaneuf:
Okay, added a commit to fix this (I'll squash it before merging, this is just so you can see what I did).

cpp/util/fake_etcd.cc
```diff
@@ -118,39 +156,86 @@ void FakeEtcdClient::Get(const Request& req, GetResponse* resp, Task* task) {
...
+      parent_key.append("/" + path);
+      // Either this inserts a directory, or it fails, and it gets us
+      // the entry. If we insert a directory, then success is
+      // guaranteed below, so we're sure to update index_
+      // appropriately.
+      if (!entries_.insert(make_pair(parent_key,
+                                     Node(new_index, new_index, parent_key,
+                                          true, "", {}, false)))
+               .first->second.is_dir_) {
+        task->Return(Status(util::error::ABORTED, "Not a directory"));
```
AlCutter: Don't capitalise the message.

cpp/util/fake_etcd.cc
```diff
@@ -25,6 +25,35 @@ using util::StatusOr;
 using util::Task;
 
 namespace cert_trans {
+namespace {
+
+
+string NormalizeKey(const string& input) {
```
AlCutter: This doesn't really seem to normalise per se, just collapse repeated slashes (doesn't handle `..` for example. What does etcd do with things like that?)

pphaneuf: Good catch. It does handle ```.``` and ```..``` properly, it looks like! I'll revise this accordingly...

cpp/util/fake_etcd.cc
pphaneuf: I was "cloning" the behaviour of ```EtcdClient``` which uses the message from the server (which is exactly this). It doesn't really matter much, and I suppose code that would depend on this would be wrong to do so anyway... Keep it? Change it?

AlCutter: Ah, your call :)

pphaneuf:
Ok, fixed the ```NormalizeKey``` function to handle ```.``` and ```..```.

PR #670, opened 2015-04-15
Fix fake_etcd_test to still work with a real etcd.

pphaneuf:
Adding the stats somewhat mangled the ability to run ```fake_etcd_test``` against a real etcd, to validate that the fake is close enough. Sure enough, there's a few updates in the fake stats to behave more like the real thing. :wink:


cpp/util/fake_etcd.cc
```diff
@@ -112,6 +112,9 @@ void FakeEtcdClient::Get(const Request& req, GetResponse* resp, Task* task) {
   CHECK(!req.recursive) << "not implemented";
   CHECK_LE(req.wait_index, 0) << "not implemented";
 
+  task->CleanupWhenDone(
```
pphaneuf: This was correct, but I figured this was a safer pattern, considering other changes coming here...

PR #673, opened 2015-04-16
Add support for recursive get in FakeEtcdClient.

pphaneuf:


cpp/util/fake_etcd.cc
```diff
@@ -157,7 +190,7 @@ void FakeEtcdClient::Get(const Request& req, GetResponse* resp, Task* task) {
   VLOG(1) << "GET " << req.key;
   const string key(NormalizeKey(req.key));
 
-  CHECK_LE(req.wait_index, 0) << "not implemented";
+  CHECK_NE(key, "/") << "not implemented";
```
pphaneuf: This is a weird edge case, and since nothing uses it, this will do for now. :stuck_out_tongue:

cpp/util/fake_etcd.cc
```diff
@@ -375,8 +424,9 @@ void FakeEtcdClient::Delete(const string& key, const int64_t current_index,
                             to_string(entry->second.modified_index_)));
     return;
   }
+  entry->second.modified_index_ = ++index_;
+  entry->second.value_.clear();
```
pphaneuf: This is some "be consistent with the real thing" stuff that I sneaked in, covering the watch logic (which isn't covered in ```etcd_test```!). Remember that ```masterelection_test``` uses the ```FakeEtcdClient```, so acting weird here could percolate into false negative (not failing when it should) in the election testing, which makes me nervous...

AlCutter: This fake_etcd is really starting to outstay its welcome...

Issue #606, opened 2015-03-26
lame duck adding entries if we are overloaded


pphaneuf:
If there are too many entries in etcd, we could get into a lot of trouble that might be very hard to recover from (if etcd can't load its Raft log without crashing, say). So we would like to provide a safety valve, where we would not accept new entries if we are getting overloaded.

PR #675, opened 2015-04-16
Reject calls to AddPendingEntry() if etcd has too many entries.

AlCutter:


cpp/log/etcd_consistent_store-inl.h
```diff
@@ -789,6 +879,79 @@ util::StatusOr<int64_t> EtcdConsistentStore<Logged>::CleanupOldEntries() {
 }
 
 
+template <class Logged>
+void EtcdConsistentStore<Logged>::StartEtcdStatsFetch() {
+  if (etcd_stats_task_.task()->CancelRequested()) {
+    etcd_stats_task_.task()->Return(
+        util::Status(util::error::CANCELLED, "Cancelled."));
```
pphaneuf: A nicer message is kind of nice, but I'd stick to just ```util::Status::CANCELLED```, this is a pattern that happens a lot, best to keep it low-profile.

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -789,6 +879,79 @@ util::StatusOr<int64_t> EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
+      bind(&EtcdConsistentStore<Logged>::EtcdStatsFetchDone, this, response,
+           std::placeholders::_1)));
+  client_->GetStoreStats(response, stats_task);
+}
+
+
+template <class Logged>
+void EtcdConsistentStore<Logged>::EtcdStatsFetchDone(
+    EtcdClient::StatsResponse* response, util::Task* task) {
+  CHECK_NOTNULL(response);
```
pphaneuf: Could inline that one in the deleter?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -789,6 +879,79 @@ util::StatusOr<int64_t> EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
+    if (num_entries.ok()) {
+      {
+        std::lock_guard<std::mutex> lock(mutex_);
+        num_etcd_entries_ = num_entries.ValueOrDie();
+      }
+      etcd_total_entries->Set("all", num_etcd_entries_);
+    } else {
+      VLOG(1) << num_entries.status();
+    }
+  }
```
pphaneuf: Maybe log something if the task wasn't OK? Shouldn't happen much, right?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -789,6 +879,79 @@ util::StatusOr<int64_t> EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
+    const util::StatusOr<int64_t> num_entries(
+        CalculateNumEtcdEntries(response->stats));
+    if (num_entries.ok()) {
+      {
+        std::lock_guard<std::mutex> lock(mutex_);
+        num_etcd_entries_ = num_entries.ValueOrDie();
+      }
+      etcd_total_entries->Set("all", num_etcd_entries_);
+    } else {
+      VLOG(1) << num_entries.status();
```
pphaneuf: That's going to be a bit bare, if it happens (maybe as little as "UNKNOWN: "!).

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -789,6 +879,79 @@ util::StatusOr<int64_t> EtcdConsistentStore<Logged>::CleanupOldEntries() {
...
+// This method attempts to modulate the incoming traffic in response to the
+// number of entries currently in etcd.
+//
+// Once the number of entries is above reject_threshold, we will start
+// returning a RESOURCE_EXHAUSTED status, which should result in a 503 being
+// sent to the client.
+template <class Logged>
+util::Status EtcdConsistentStore<Logged>::MaybeReject(
+    const std::string& type) const {
+  if (!cluster_config_) {
```
pphaneuf: Doesn't that need to be protected by the lock?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -84,8 +148,13 @@ template <class Logged>
 EtcdConsistentStore<Logged>::~EtcdConsistentStore() {
   VLOG(1) << "Cancelling watch tasks.";
   serving_sth_watch_task_.Cancel();
+  cluster_config_watch_task_.Cancel();
   VLOG(1) << "Waiting for watch tasks to return.";
   serving_sth_watch_task_.Wait();
+  cluster_config_watch_task_.Wait();
+  VLOG(1) << "Cancelling stats task.";
+  etcd_stats_task_.Cancel();
```
pphaneuf: Would be slightly better to cancel everything, then wait on everything?

This has a slight smell that maybe we should just have one SyncTask, and use child tasks for all these things, but don't do this in this PR, just putting the idea in your head... :wink:

pphaneuf:
Done for this round.

cpp/log/etcd_consistent_store-inl.h
AlCutter: I forgot about the static `CANCELLED` instance, thanks.

cpp/log/etcd_consistent_store-inl.h
AlCutter: It causes the line to wrap so I'm not sure there's much to be gained...

cpp/log/etcd_consistent_store-inl.h
AlCutter: Done.

cpp/log/etcd_consistent_store-inl.h
AlCutter: Added verbiage.

cpp/log/etcd_consistent_store-inl.h
AlCutter: Whoops, yes.

cpp/log/etcd_consistent_store-inl.h
AlCutter: Yeah, good call.
Another day, another PR.

cpp/log/etcd_consistent_store-inl.h
pphaneuf: Okay, fine like this, then.

pphaneuf:
LGTM

AlCutter:
Fixes #606

PR #677, opened 2015-04-17
Relocate http stats for now.

AlCutter:
There was a race between the http handler and updating the `total_http_server_requests` and `total_http_server_response_codes` stats, for now just relocate them to remove the race.


pphaneuf:
LGTM

PR #680, opened 2015-04-20
Use autoconf/automake.

pphaneuf:
This gets us many things.

 - much easier control of variables controlling compilation (for example, overriding CXXFLAGS, and adding -I/-L flags to CPPFLAGS/LDFLAGS)
 - gtest/gmock is now compiled with compilation flags consistent with the rest of our build (as per their recommendations)
 - faster automatic dependencies
 - concurrent execution of tests

Still missing is validation that all required tools and libraries are found and working, but it should be possible to point at working ones without modifying any Makefile or scripts.

This should also fix #253, as a side-effect.


pphaneuf:
I forgot to mention: this *should* also allow out-of-tree compilation, which makes it easier to do things like TSan/ASan builds (which, themselves, should be a simple matter of passing the appropriate ```CXXFLAGS``` to the ```configure``` script).

pphaneuf:
Also, a few tips for the ```make``` user: the target we called ```test``` is called ```check``` in GNU-land, and there's a new ```recheck``` target that only runs failing tests.

AlCutter:
er, lgtm?

PR #682, opened 2015-04-20
Fix a crashing bug in the fetcher.

pphaneuf:
That lock inside the Range was poorly used. The state_ member can play an equivalent role already, without the dangers of unlocking deleted mutexes.

### DIFF
--- a/cpp/fetcher/fetcher.cc
+++ b/cpp/fetcher/fetcher.cc
@@ -15,6 +15,7 @@ using std::lock_guard;
 using std::move;
 using std::mutex;
 using std::placeholders::_1;
+using std::unique_lock;
 using std::unique_ptr;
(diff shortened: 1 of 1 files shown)


cpp/fetcher/fetcher.cc
```diff
@@ -178,7 +177,8 @@ void FetchState::WalkEntries() {
 }
 
 
-void FetchState::FetchRange(Range* current, int64_t index, Task* range_task) {
+void FetchState::FetchRange(const unique_lock<mutex>& lock, Range* current,
+                            int64_t index, Task* range_task) {
   const int64_t end_index(index + current->size_ - 1);
```
AlCutter: Want a `CHECK(lock.owns_lock())` here?

PR #684, opened 2015-04-21
Proposed protobuf changes to make the distinguished name fields of X509Description a bit more strongly typed.

phad:
I'd like to deprecate the existing ct.X509Description.{subject|issuer|subject_alternative_names|root_issuer} fields and introduce replacements that are more strongly typed.

The rationale for this is that the components of a Distinguished Name that most apps care about are predictable.  The current repeated DNAttribute fields are basically a string dictionary.  This doesn't lend itself to DB indexing, eg. if one wanted to build an index of subject CN fields.

### DIFF
--- a/python/ct/client/db/log_db_test.py
+++ b/python/ct/client/db/log_db_test.py
@@ -15,7 +15,7 @@ class LogDBTest(object):
     # Set up a default fake test log server and STH.
     default_log = client_pb2.CtLogMetadata()
     default_log.log_server = "test"
-    default_log.log_id = "somekeyid"
+    default_log.log_id = "c29tZWtleWlk"  # b64("somekeyid")
     default_log.public_key_info.type = client_pb2.KeyInfo.ECDSA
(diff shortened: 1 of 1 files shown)


pphaneuf:
LGTM

eranmes:
LGTM, it seems that the proto fields can be entirely removed rather than deprecated.

phad:
OK, if you guys are happy for me to remove the existing protos straight away I'll make a bunch of commits and squash the eventual PR.

pphaneuf:
You can delete them, but normally you'd leave a comment to say the the field identifier isn't available (to protect against stored protobufs being "re-interpreted" in the future). The ```deprecated``` tag kind of fill in this role, as well avoiding breaking existing code (which might still be referring to those fields.

If you know there's nothing using those fields, you could just remove them, yes (and maybe leave a comment to indicate which field IDs are no longer available).

pphaneuf:
Abandoning this? Or closed by mistake?

Issue #683, opened 2015-04-21
compilation problems with gcc


pphaneuf:
We've been using Clang for a while, because it gave us more detailed warnings and diagnostics, but it seems one or two small problems creeped in that only show with GCC.

```
cpp/fetcher/fetcher.cc: In member function ‘void {anonymous}::FetchState::WriteToDatabase(int64_t, {anonymous}::Range*, const std::vector<cert_trans::AsyncLogClient::Entry>*, util::Task*, util::Task*)’:
cpp/fetcher/fetcher.cc:253:32: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   if (processed < retval->size()) {
                                ^
In file included from ./cpp/log/etcd_consistent_store-inl.h:13:0,
                 from cpp/log/etcd_consistent_store_cert.cc:1:
./cpp/monitoring/latency.h:66:58: error: declaration of ‘cert_trans::ScopedLatency cert_trans::Latency<TimeUnit, LabelTypes>::ScopedLatency(const LabelTypes& ...)’ [-fpermissive]
   ScopedLatency ScopedLatency(const LabelTypes&... labels);
                                                          ^
./cpp/monitoring/latency.h:13:7: error: changes meaning of ‘ScopedLatency’ from ‘class cert_trans::ScopedLatency’ [-fpermissive]
 class ScopedLatency;
       ^
In file included from cpp/log/etcd_consistent_store_cert.cc:1:0:
./cpp/log/etcd_consistent_store-inl.h: In function ‘bool cert_trans::LeafEntriesMatch(const Logged&, const Logged&) [with Logged = cert_trans::LoggedCertificate]’:
./cpp/log/etcd_consistent_store-inl.h:212:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
cpp/log/file_storage.cc: In member function ‘std::string cert_trans::FileStorage::StoragePathBasename(const string&) const’:
cpp/log/file_storage.cc:94:23: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   if (hex.length() <= storage_depth_)
                       ^
cpp/log/file_storage.cc: In member function ‘std::string cert_trans::FileStorage::StoragePathComponent(const string&, int) const’:
cpp/log/file_storage.cc:103:23: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   if (n >= hex.length())
                       ^
[...]

PR #686, opened 2015-04-21
Fix warnings found by gcc.

pphaneuf:
A few things that we didn't catch with Clang, oddly enough.


pphaneuf:
This should fix #683.

PR #691, opened 2015-04-22
Match style of generated header to the latest Chromium one.

eranmes:


pphaneuf:
LGTM as well. :smiley:

PR #695, opened 2015-04-23
Include (or at least, fwd-declare) what you use!

phad:
Let's forward-declare classes we use in our .h files eh?


AlCutter:
Looks wonderful to me, squash'n'merge.

PR #697, opened 2015-04-23
Remove CR characters from the plain-text dump in some PEM files.

pphaneuf:
Am I correct in thinking that only the Base64 chunks in those files really matter, and those plain-text bits are just for convenience?


eranmes:
You're correct.

PR #705, opened 2015-04-23
Add explicit checks for GTest/GMock/libevent in configure.

pphaneuf:


pphaneuf:
I added libevent to this PR as well (as a separate commit).

PR #709, opened 2015-04-24
Faster cleaner

pphaneuf:
Since JSON parsing the big `GetPendingEntries()` response, and decoding all the base64 in it is quite intensive, don't do it so much. :smiley:

### DIFF
--- a/cpp/log/etcd_consistent_store_test.cc
+++ b/cpp/log/etcd_consistent_store_test.cc
@@ -728,7 +728,6 @@ TEST_F(EtcdConsistentStoreTest, TestCleansUpToNewSTH) {
   EXPECT_EQ(orig_seq_mapping.Entry().DebugString(),
             seq_mapping.Entry().DebugString());
 
-
   // Now update ServingSTH so that all sequenced entries should be cleaned up:
   sth.set_timestamp(sth.timestamp() + 1);
   sth.set_tree_size(105);
@@ -736,7 +735,7 @@ TEST_F(EtcdConsistentStoreTest, TestCleansUpToNewSTH) {
   {
     const StatusOr<int64_t> num_cleaned(CleanupOldEntries());
     ASSERT_OK(num_cleaned.status());
-    EXPECT_EQ(2, num_cleaned.ValueOrDie());
+    EXPECT_EQ(5, num_cleaned.ValueOrDie());
   }
 
 
--- a/cpp/util/etcd_delete.cc
+++ b/cpp/util/etcd_delete.cc
@@ -8,7 +8,6 @@
 using std::bind;
 using std::move;
 using std::mutex;
-using std::pair;
 using std::placeholders::_1;
 using std::string;
 using std::unique_lock;
@@ -26,8 +25,7 @@ namespace {
 
 class DeleteState {
  public:
-  DeleteState(EtcdClient* client, vector<pair<string, int64_t>>&& keys,
-              Task* task)
+  DeleteState(EtcdClient* client, vector<string>&& keys, Task* task)
       : client_(CHECK_NOTNULL(client)),
         task_(CHECK_NOTNULL(task)),
         outstanding_(0),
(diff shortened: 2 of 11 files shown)


pphaneuf:
Hmm, I don't get that failure here, not sure what's going on, any idea?

cpp/util/etcd_delete.h
```diff
@@ -11,6 +11,10 @@
 namespace cert_trans {
 
 
+// Delete keys in batches (implemented using concurrent requests). The
+// "keys" argument are pairs of key and modified index. If the index
+// for an entry is zero, EtcdClient::ForceDelete will be used instead
+// of EtcdClient::Delete.
```
AlCutter: Not super keen on the special casing here, how about a separate method `EtcdForceDeleteKeys()` ?

cpp/util/etcd_delete.cc
```diff
@@ -64,9 +64,11 @@ void DeleteState::RequestDone(Task* child_task) {
   unique_lock<mutex> lock(mutex_);
   --outstanding_;
 
-  // If a child task has an error, return that error, and do not start
-  // any more requests.
-  if (!child_task->status().ok()) {
+  // If a child task has an error (except for not found, this is close
+  // enough to success), return that error, and do not start any more
+  // requests.
+  if (!child_task->status().ok() &&
```
AlCutter: Might be nice to still treat `NOT_FOUND` as a legitimate error when it's from a regular (i.e. with handle) delete; failing to delete something you really expected to be there could be "interesting" info.

cpp/log/etcd_consistent_store_test.cc
```diff
@@ -736,7 +735,7 @@ TEST_F(EtcdConsistentStoreTest, TestCleansUpToNewSTH) {
   {
     const StatusOr<int64_t> num_cleaned(CleanupOldEntries());
     ASSERT_OK(num_cleaned.status());
-    EXPECT_EQ(2, num_cleaned.ValueOrDie());
+    EXPECT_EQ(5, num_cleaned.ValueOrDie());
```
AlCutter: image

We can't keep track of how many were deleted `OK`?

pphaneuf: The faster cleaner doesn't do a `Get` first, and `EtcdDeleteKeys` doesn't report the number of keys deleted, indeed. It'll now repeatedly try to delete keys that are still in the mapping, and it's the signer that will be cleaning the stale mapping entries.

cpp/util/etcd_delete_test.cc
```diff
@@ -221,6 +224,22 @@ TEST_F(EtcdDeleteTest, ErrorHandling) {
 }
 
 
+TEST_F(EtcdDeleteTest, ForceDelete) {
+  vector<pair<string, int64_t> > keys{make_pair("/one", 1),
+                                      make_pair("/two", 0)};
+  SyncTask sync(&pool_);
+
+  EXPECT_CALL(client_, Delete("/one", 1, _))
```
AlCutter: I think this is failing because `Task::Return()` returns a `bool`, gmock noticed that and is trying to use that returned bool as the return value for the mocked function which is `void`

You probably want something like:
```c++
EXPECT_CALL(client_, Delete("/one", 1, _))
    .WillOnce(IgnoreResult(Invoke(bind(&Task::Return, _3, Status::OK))));
```

pphaneuf: Right on, the compiler on my machine was fine with it, the one on Travis wasn't but this did fix it.

cpp/util/etcd_delete.h
pphaneuf: One of the issues is that I was trying to reuse the internals and avoid changing them too much...

It's doable, but I'm pondering changing the signature altogether to be only "force", since that's the only spot we're using it. What do you think?

AlCutter: Could do that, yup.

pphaneuf:
Okay. fixed and rebased this.

PR #710, opened 2015-04-24
make SQLite<>::TreeSize() lazily evaluated

### LINKED ISSUE #568: SQLite is quite slow to start
Turns out we scan the database three times at startup: once to find how big is the contiguous tree size, once to initialize ```TreeSigner```, and once more to initialize ```LogLookup```.

Issue #428 deals with combining the last two, but we should be able to initialize the contiguous tree size lazily: make ```tree_size_``` mean "the *last known* contiguous tree size", have it start at 0, and every time an entry is read or written with an index number equal to ```tree_size_```, it means we can advance ```tree_size_``` by one.

Calling ```TreeSize()``` should confirm that this is correct, but if it's done lazily, it should normally be quick. This does mean that it will become effectively blocking, though, so call sites will have to be reviewed.

pphaneuf:
Starts with a few commits to try to avoid calling `TreeSize()` from the libevent thread.

This fixes #568.

### DIFF
--- a/cpp/log/sqlite_db-inl.h
+++ b/cpp/log/sqlite_db-inl.h
@@ -134,7 +134,6 @@ SQLiteDB<Logged>::SQLiteDB(const std::string& dbfile)
   }
 
   BeginTransaction(lock);
-  UpdateTreeSize();
 }
 
 
@@ -177,6 +176,10 @@ typename Database<Logged>::WriteResult SQLiteDB<Logged>::CreateSequencedEntry_(
       std::string existing_hash;
       s2.GetBlob(1, &existing_hash);
 
+      if (logged.sequence_number() == tree_size_) {
+        ++tree_size_;
+      }
+
       if (hash == existing_hash) {
         return this->OK;
       }
@@ -185,7 +188,9 @@ typename Database<Logged>::WriteResult SQLiteDB<Logged>::CreateSequencedEntry_(
   }
   CHECK_EQ(SQLITE_DONE, ret);
 
-  UpdateTreeSize();
+  if (logged.sequence_number() == tree_size_) {
+    ++tree_size_;
+  }
 
   return this->OK;
 }
@@ -220,6 +225,9 @@ typename Database<Logged>::LookupResult SQLiteDB<Logged>::LookupByHash(
     result->clear_sequence_number();
   } else {
     result->set_sequence_number(statement.GetUInt64(1));
+    if (result->sequence_number() == tree_size_) {
+      ++tree_size_;
+    }
   }
 
   return this->LOOKUP_OK;
@@ -252,6 +260,9 @@ typename Database<Logged>::LookupResult SQLiteDB<Logged>::LookupByIndex(
   CHECK_EQ(result->Hash(), hash);
 
   result->set_sequence_number(sequence_number);
+  if (result->sequence_number() == tree_size_) {
+    ++tree_size_;
+  }
 
   return this->LOOKUP_OK;
 }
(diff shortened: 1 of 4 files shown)


cpp/log/sqlite_db-inl.h
```diff
@@ -317,9 +328,27 @@ template <class Logged>
 int64_t SQLiteDB<Logged>::TreeSize() const {
   cert_trans::ScopedLatency latency(
       latency_by_op_ms.GetScopedLatency("tree_size"));
-  std::lock_guard<std::mutex> lock(lock_);
+  std::unique_lock<std::mutex> lock(lock_);
 
   CHECK_GE(tree_size_, 0);
+  sqlite::Statement statement(
```
AlCutter: Whoa!
`TreeSize()` is called from a few places with the expectation that it's *very* cheap, e.g. `TreeSigner::SequenceNewEntries()`, `ClusterStateController::NodeIsStale()` (which in turn is called for every single HTTP request).

cpp/log/sqlite_db-inl.h
```diff
@@ -317,9 +328,27 @@ template <class Logged>
...
+      db_,
+      "SELECT sequence FROM leaves WHERE sequence >= ? ORDER BY sequence;");
```
AlCutter: Don't need the trailing semicolon on the SQL.

cpp/log/sqlite_db-inl.h
pphaneuf: Yeah, but in truth, due to taking the lock (which is held during other arbitrarily long, blocking operations), this was already blocking and potentially very much *not* cheap, despite wishing it not to be. :wink: 

The previous commits remove some of the calls, or makes sure they're not on the libevent thread.

`ClusterStateController::NodeIsStale()`, for example, is called from `HttpHandler::ProxyInterceptor()`, but specifically bounced to the thread pool. `TreeSigner::SequenceNewEntries()` is called from its own thread. They both also only check it once, not in a loop or anything like that.

I think the staleness check could be improved to be cached a bit (check a raw bool that is updated regularly, say?), but I'd leave this for later, considering how `TreeSize()` was already "secretly blocking"...

cpp/log/sqlite_db-inl.h
pphaneuf: Fixed.

cpp/log/sqlite_db-inl.h
AlCutter: Yeah, but "not cheap" because blocking and "not cheap" because blocking and extending the lock while burning cycles doing SQL on every query are quite different.  Maybe it's ok if you can cache the staleness state as you suggest, but it'd be nice to have some benchmarks showing what effect this has (if any) on our current state of affairs, doesn't need to be over the full 6M certs.

pphaneuf: I'll measure the time to start and report back. I think the only place that is actually affected is `HttpHandler::ProxyInterceptor` (via `NodeIsStale`), so if there's any win to be had (and I'm hoping there is), I'll cache the staleness state in a separate PR first.

AlCutter: Cool, thanks!

pphaneuf: Hacked up a little test program that only signs a new tree head, with 100K entries in the database. On master, 20 to 45 seconds, with this PR, 12 to 25 seconds. I used `sync && sudo sh -c 'echo 3 >/proc/sys/vm/drop_caches'` to minimize caching effects.

I'm guessing that this lazy evaluation isn't *twice* as fast because there's some cache warming for the tree signing from having walked the database once, but even then, it's a good fraction of 2x faster.

pphaneuf: I just send #731 to cache the staleness status in `HttpHandler`. Once that's merged in, I'd say that declaring `Database<>::TreeSize()` to be blocking would be okay.

AlCutter: Cool :+1:

PR #711, opened 2015-04-27
Use read base64'd logIDs in the log_db_test

phad:
The malformed log_id 'somekeyid' caused a derived database test to fail.
Also, the new_log used in a later test had no log_id at all, also causing the derived database test to fail.


pphaneuf:
LGTM

PR #712, opened 2015-04-27
When processing subject name, don't downcase strings unless processing dot-separated DNS names.

phad:
Examining the function, the only time it's important to downcase is where the subject string is interpreted as a DNS name.  In other cases, the string is human-readable and shouldn't be downcased.

### DIFF
--- a/python/ct/client/db/cert_db_test.py
+++ b/python/ct/client/db/cert_db_test.py
@@ -108,7 +108,7 @@ class CertDBTest(object):
         for i in range(4):
             cert = certificate_pb2.X509Description()
             cert.der = "hello-%d" % i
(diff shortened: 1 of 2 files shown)


eranmes:
Code LGTM, but the description is a bit odd: To me it reads that strings will be downcased if  dot-separated DNS names are processed and won't be downcased otherwise. From the code it seems that the labels will always be downcased while the subject never will be?

phad:
Sorry, it's a bit of a double-negative.
It would read better as: "When processing subject name, only downcase strings if they are dot-separated DNS names".
Is that better?  I don't know if I can change the commit message at this point.

phad:
New commit description added to clarify.

PR #720, opened 2015-04-28
Election lock files in the right place.

AlCutter:


cpp/server/ct-server.cc
```diff
@@ -86,6 +86,7 @@ DEFINE_double(guard_window_seconds, 60,
               "number of seconds will not be sequenced.");
 DEFINE_string(etcd_host, "", "Hostname of the etcd server");
 DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
+DEFINE_string(etcd_root, "/root", "Root of cluster entries in etcd.");
```
pphaneuf: Maybe name this root with something related to `ct-server`?

pphaneuf:
LGTM, with possible small tweak (I'm having sharing of an etcd instance with other services in mind).

cpp/server/ct-server.cc
AlCutter: I'll save that for another PR because it'll also involve changing some of the GCE scripts etc.

Issue #644, opened 2015-04-09
wait for the election proposal to disappear


pphaneuf:
We currently crash on startup if an election proposal with our node ID is already there, maybe we should wait a bit instead of CHECK-failing. Maybe get the TTL, and sleep until just a bit after, and if it's still there after that, *then* crash.

PR #721, opened 2015-04-28
Don't check when failing to create election proposal.

AlCutter:
This addresses #644.


pphaneuf:
Is this related to #644? Update the commit message to say something about fixing it, if so?

pphaneuf:
LGTM

PR #729, opened 2015-04-30
Hermetic build, works on FreeBSD for C++ only currently, previous version works on MacOS.

benlaurie:
Definitely a work in progress.

Probably only works on MacOS.

Probably omits needed packages.

gclient will only run a single command: sometimes its a shell script, sometimes its make. Haven't decided which I prefer yet.

gclient seems to run all the hooks all the time. Not investigated why yet.


pphaneuf:
Hey, @maruel, have you got tips for us? Or are we just dooooomed, as usual? :wink:

maruel:
Hermetic build != deterministic build != reproducible build.
Chromium has reproducible build (we sync everything at known commit hash), I'm working on deterministic build and hermetic build requires using a full image, which is a tad intense.

So you meant reproducible build. :)

pphaneuf:
Thanks @maruel!

One thing I was finding a bit sketchy was in `build/rebuild`, there's a call to `configure`... That would be every time we run `gclient sync`? Isn't that a bit... aggressive? Considering that `configure` takes some time to run, and that the makefile has dependencies set up for it (so it'll run it for us if needed)...

maruel:
Exact, that's why I do not recommend to use the hooks functionality inside the DEPS file here.

pphaneuf:
I looked at Chromium, and instantly got a headache. :wink: 

I see that they're not using the hooks in the `DEPS` file, but I don't remember enough of their build system to see where dependencies get configured and build... Surely, there's at least one of these dependencies that uses autoconf/automake and could be used as an example, right? :smiley:

pphaneuf:
I chatted with @maruel a bit, and I think the overall suggestion would be having a separate Makefile for gclient (can have some wacky name, and invoke it with "make -f build.gclient" or something), that would make sure everything is as it should be, and invoke just that one with a hook, at the end. Being clever in that one, we'd compile the dependencies, and configure ourselves, but leave "make" to the user, in the "classic style".

benlaurie:
On 30 April 2015 at 14:12, M-A <notifications@github.com> wrote:

> Hermetic build != deterministic build != reproducible build.
> Chromium has reproducible build (we sync everything at known commit hash),
>
I'm interested that you make this claim. I haven't yet found anyone
prepared to vouch that if git says the hash is X (or, strictly, is prepared
to retrieve content whose hash is allegedly X), the actual content of your
checkout is what you expect. That is, if the upstream repo is malicious,
no-one is so far prepared to say you're defended.

What's your view?

pphaneuf:
I think he means more that an upstream repo might not have the best maintainers, and randomly move/delete tags?

maruel:
In the case of Chromium, we host a copy on Google's server, and our infrastructure use this copy. They are hosted on https://chromium.googlesource.com/, look for all the ones prefixed with external/.

We control this via the gclient variables in https://chromium.googlesource.com/chromium/src.git/+/master/DEPS#29

When a dev forgets to do the variable replacement, we usually find it relatively soon: a git server on the internet cries out in terror then is silenced, as it is DDoS'ed by thousands of chromium infrastructure bots trying to git clone from it simultaneously.

python/Makefile
```diff
@@ -2,7 +2,7 @@ all: ct/proto/client_pb2.py ct/proto/ct_pb2.py ct/proto/tls_options_pb2.py \
 	ct/proto/test_message_pb2.py ct/proto/certificate_pb2.py
 
 ct/proto/%_pb2.py: ct/proto/%.proto
-	protoc $^ -I/usr/include/ -I/usr/local/include -I. --python_out=.
+	protoc $^ -I/usr/include/ -I/usr/local/include -I`pwd`/../../install/include -I. --python_out=.
```
pphaneuf: Surely, you jest?

pphaneuf:
dc53cafb51bd853d6320fa31985ee6e717318ad8 has been merged into master, if you rebase this, it'll disappear. :smiley:

PR #736, opened 2015-05-07
Another downcasing fix for extraction of values from Distinguished Names.

phad:
Single-word names (eg. State, City etc.) were being down-cased
incorrectly because they fell into the 'valid [DNS name]' block.
Tighten the 'valid DNS name' check by requiring >1 dot-separated
term.


pphaneuf:
LGTM

PR #737, opened 2015-05-07
Fix a few lint errors from 558cab0

phad:
Some too-long lines and wrong types of quotes.
The file produces a ton of indentation lint errors which I'm not fixing here.


pphaneuf:
LGTM

PR #740, opened 2015-05-08
Try a bit harder to detect leveldb.

pphaneuf:
Some distributions only have static libraries packaged, which requires a bit more effort (explicitly listing libraries it depends on).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -76,6 +76,15 @@ AS_IF([test "x$have_glog" = "xno"],
 
 save_LIBS="$LIBS"
 AS_UNSET([LIBS])
+AC_SEARCH_LIBS([snappy_compress], [snappy],,, [$save_LIBS])
+AC_SEARCH_LIBS([leveldb_open], [leveldb],, [missing_leveldb=1], [$save_LIBS])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -76,6 +76,15 @@ AS_IF([test "x$have_glog" = "xno"],
 
 save_LIBS="$LIBS"
 AS_UNSET([LIBS])
+AC_SEARCH_LIBS([snappy_compress], [snappy],,, [$save_LIBS])
```
pphaneuf: This is the tricksy bit: if you have `libsnappy-dev` installed, it will detect it, and `leveldb_LIBS` will have `-lleveldb -lsnappy`. If you have the older `libleveldb-dev` from Ubuntu Precise that only has static libraries, this is what you have to do. If you have the newer one that has shared libraries (where `libleveldb.so` is linked with `libsnappy.so`, so you don't have to), then it's spurious, but shouldn't harm anything.

I figure this is an acceptable compromise, keeping things working, in exchange for having an extra direct dependency that we have anyway, indirectly.

PR #741, opened 2015-05-11
Tweak the LevelDB options a bit.

pphaneuf:
Nothing too fancy...

### DIFF
--- a/cpp/log/leveldb_db-inl.h
+++ b/cpp/log/leveldb_db-inl.h
@@ -14,6 +14,10 @@
 #include "monitoring/latency.h"
 #include "util/util.h"
 
+DEFINE_int32(leveldb_max_open_files, 0,
+             "number of open files that can be used by leveldb");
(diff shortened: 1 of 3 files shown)


pphaneuf:
The version of LevelDB packaged with Ubuntu Precise does not have bloom filter support. I revised this pull request to detect that, and only compile it in when it is there.

cpp/log/leveldb_db-inl.h
```diff
@@ -47,11 +65,24 @@ const size_t LevelDB<Logged>::kTimestampBytesIndexed = 6;
 
 template <class Logged>
 LevelDB<Logged>::LevelDB(const std::string& dbfile)
-    : contiguous_size_(0), latest_tree_timestamp_(0) {
+    : lock_(),
```
AlCutter: Should I laugh or cry? :)

pphaneuf: I felt pretty bad, so I took it out. Formatting is clang-format's own.

PR #742, opened 2015-05-11
Add a Delay() method to Executor.

AlCutter:
Arguably an `AddAt()` might also be useful, but I thought I'd just mimic the `libevent::Base::Delay()` for now.

### DIFF
--- a/cpp/util/thread_pool.cc
+++ b/cpp/util/thread_pool.cc
@@ -1,4 +1,5 @@
 #include "util/thread_pool.h"
+#include "util/task.h"
 
 #include <condition_variable>
 #include <glog/logging.h>
@@ -7,16 +8,36 @@
 #include <thread>
 #include <vector>
 
+using std::chrono::duration;
+using std::chrono::duration_cast;
+using std::chrono::seconds;
+using std::chrono::steady_clock;
 using std::condition_variable;
 using std::function;
+using std::get;
 using std::lock_guard;
 using std::mutex;
-using std::queue;
+using std::priority_queue;
 using std::thread;
+using std::tuple;
 using std::unique_lock;
 using std::vector;
 
 namespace cert_trans {
+namespace {
+
+typedef tuple<steady_clock::time_point, function<void()>, util::Task*>
+    QueueEntry;
+
+
+struct QueueOrdering {
+  bool operator()(const QueueEntry& lhs, const QueueEntry& rhs) const {
+    return get<0>(lhs) > get<0>(rhs);
+  }
+};
+
+
+}  // namespace
 
 
 class ThreadPool::Impl {
@@ -31,7 +52,7 @@ class ThreadPool::Impl {
 
   mutex queue_lock_;
   condition_variable queue_cond_var_;
-  queue<function<void()> > queue_;
+  priority_queue<QueueEntry, vector<QueueEntry>, QueueOrdering> queue_;
 };
 
 
@@ -41,7 +62,8 @@ ThreadPool::Impl::~Impl() {
   {
     lock_guard<mutex> lock(queue_lock_);
     for (int i = threads_.size(); i > 0; --i)
-      queue_.push(function<void()>());
+      queue_.emplace(
+          make_tuple(steady_clock::time_point(), function<void()>(), nullptr));
   }
   // Notify all the threads *after* adding all the empty closures, to
   // avoid any races.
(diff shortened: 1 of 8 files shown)


pphaneuf:
As discussed in person, keeping a FIFO ordering of the `Add()`ed functions would be nice (to avoid starvation), and busy-waiting when there's delayed tasks in the future, that's not cool. :wink:

AlCutter:
Have another look.

cpp/util/thread_pool.cc
```diff
@@ -7,16 +8,36 @@
...
+using std::priority_queue;
 using std::thread;
+using std::tuple;
 using std::unique_lock;
 using std::vector;
 
 namespace cert_trans {
+namespace {
+
+typedef tuple<system_clock::time_point, function<void()>, util::Task*>
```
pphaneuf: Using `system_clock`, instead of `steady_clock`?

cpp/util/thread_pool.cc
```diff
@@ -114,7 +146,19 @@ void ThreadPool::Add(const function<void()>& closure) {
...
+  impl_->queue_cond_var_.notify_one();
+}
+
+
+void ThreadPool::Delay(const duration<double>& delay, util::Task* task) {
+  CHECK_NOTNULL(task);
+  {
+    lock_guard<mutex> lock(impl_->queue_lock_);
+    impl_->queue_.emplace(make_tuple(
+        system_clock::now() + duration_cast<std::chrono::microseconds>(delay),
```
pphaneuf: There was the need for a `duration_cast` here? `system_clock` uses an integer duration or something like that?

cpp/util/thread_pool.cc
AlCutter: `steady_clock` would be better, thanks - I'll switch it over.

cpp/util/thread_pool.cc
AlCutter: Yup.

cpp/util/thread_pool.cc
```diff
@@ -56,30 +78,40 @@ ThreadPool::Impl::~Impl() {
...
-      if (!queue_.front())
-        break;
-
-      closure = queue_.front();
+      entry = queue_.top();
       queue_.pop();
+
+      // If we received an empty entry, exit cleanly.
+      if (!get<1>(entry)) {
+        // Anything left in the queue must be either other exit sentintels, or
```
pphaneuf: "sentinels"

AlCutter: Sentintels are better than plain old sentinels; they do the same job but come with extra tea.

cpp/util/thread_pool.cc
```diff
@@ -56,30 +78,40 @@ ThreadPool::Impl::~Impl() {
...
+        // future (delayed) tasks, so we'll cancel anything we find, up until
+        // either the next exit sentinel, or the end of the queue.
+        VLOG(1) << "Cancelling delayed tasks...";
+        int cancelled(0);
+        while (!queue_.empty() && get<2>(queue_.top())) {
+          get<2>(queue_.top())->Return(util::Status::CANCELLED);
```
pphaneuf: Should never happen, of course, but just in case, sticking a `CHECK_NOTNULL` in there will make the error message a lot more readable.

pphaneuf: And in the "should never happen" scenario, think of this: someone uses a thread pool for a delay, using a `util::Task` that uses that thread pool as its executor. :wink:

AlCutter: `CHECK_NOTNULL` added

AlCutter: Ah, you mean like in `ThreadPoolTest::Delay` ? :)

pphaneuf: Oh, I had missed the check for `get<2>(queue_.top())`, that'll avoid the null-dereference, at least, but I think I'd like a `CHECK` if there is something we don't expect.

AlCutter: I've added a `CHECK(queue_.empty())` to the d'tor too.

pphaneuf: Nice thought, with `ThreadPoolTest::Delay`, but I mean what if you get here, and one of the `util::Task` you `Return()` here use this thread pool? The call to `Return()` will call `Add()`, and now you have something *after* the sentinels. Or, if your clock has low enough precision, and you're "lucky" enough, it might slip in the middle. :wink:

pphaneuf: Something like this:

```C++
unique_ptr<ThreadPool> my_pool_of_one(new ThreadPool(1));
SyncTask task(my_pool_of_one.get());
my_pool_of_one->Delay(milliseconds(200), task.task());
my_pool_of_one.reset();
task.Wait();
```

AlCutter: Ah, gotcha.
Well, then arguably the task's callback should check that the return value isn't `CANCELLED` before doing more stuff with the pool.
With the new CHECK I added you'll at least know about it if it went after the sentinels.  If it snuck in before last sentinel in the queue then it should be fine; it too will be `CANCELLED`.
If THAT then adds another task and so on, then the caller should be shot.

pphaneuf: Yeah, I wasn't thinking of the `util::Task` callback itself doing more stuff with the pool (although it's not strictly wrong either), but `util::Task` itself has to queue the user's callback.

In practice, this should be all okay, because we never shut down, but this could turn up as weird crashes in tests, though (the delayed task never *getting* the cancellation status, or an outright CHECK-fail in the thread pool destructor)...

pphaneuf:
LGETM.

cpp/util/thread_pool.cc
AlCutter: Actually, what'll happen at the moment is that the `Add()` will block forever because the cancelling is done under `queue_lock`, I can sort that and then you'd hit the CHECK in the d'tor.

pphaneuf: Ah, yes, deadlock, good, good... :smiley:

PR #743, opened 2015-05-12
Use key format for entries in LevelDB that keeps them in order.

pphaneuf:


cpp/log/leveldb_db-inl.h
```diff
@@ -47,18 +47,37 @@ std::unique_ptr<const leveldb::FilterPolicy> BuildFilterPolicy() {
 #endif
 
 
-std::string FormatSequenceNumber(const int64_t seq) {
-  return std::to_string(seq);
+std::string IndexToKey(int64_t index) {
+  CHECK_EQ(8, sizeof(index));
```
AlCutter: AFAIK `int64_t` is required to be exactly 64bits, are you worried about people trying to run this on a PDP-8? :)

cpp/log/leveldb_db-inl.h
```diff
@@ -47,18 +47,37 @@ std::unique_ptr<const leveldb::FilterPolicy> BuildFilterPolicy() {
...
+  const char nibble[] = "0123456789abcdef";
+  char index_str[sizeof(index) * 2];
```
AlCutter: Could avoid the string c'tor copy below with by having a string here and `reserve()`ing the length, and just add the chars in the loop below.

cpp/log/leveldb_db-inl.h
pphaneuf: I'm slightly paranoid, sometimes. :wink:

pphaneuf: Also, this is `index` I'm looking at, I'm more worried someone will change the type of the parameter, which would invalidate existing databases...

cpp/log/leveldb_db-inl.h
pphaneuf: I'm going right to left, and `std::string` doesn't let you easily overwrite its data... My first cut had an `std::string`, and did `index_str = index_str + nibble[index & 0xf]`, which has an implicit temporary, and a copy, and... ugh. :stuck_out_tongue: 

I could do something fancier here, stand by.

pphaneuf:
Ok, added a commit, PTAL.

cpp/log/leveldb_db-inl.h
AlCutter: If they're nuts enough to do that then they'll likely just change the `CHECK` too - I'd suggest replacing the `CHECK` with a comment that changing this type will break your stuff.

cpp/log/leveldb_db-inl.h
AlCutter: doesn't `[...] = ...` let you overwrite the data? 
http://en.cppreference.com/w/cpp/string/basic_string/operator_at

AlCutter: Ah - looks like you figured that out too, then :)

cpp/log/leveldb_db-inl.h
pphaneuf: Done, PTAL.

PR #745, opened 2015-05-15
Fix CompactMerkleTree when cloning an empty source tree.

AlCutter:


cpp/merkletree/merkle_tree_test.cc
```diff
@@ -577,6 +577,12 @@ TEST_F(MerkleTreeTest, AddLeafHash) {
   EXPECT_EQ(kHashValue, tree.LeafHash(index));
 }
 
+TEST_F(CompactMerkleTreeTest, TestCloneEmptyTreeProducesWorkingTree) {
+  MerkleTree tree(new Sha256Hasher);
+  CompactMerkleTree compact(tree, new Sha256Hasher);
+  LOG(INFO) << compact.CurrentRoot();
```
pphaneuf: Hmm, this should be some sort of `EXPECT_EQ`, no?

AlCutter: It could be, but what I'm checking here is that the whole damn thing doesn't just `abort()` anymore.

pphaneuf:
Done reviewing, the change to the code itself looks good, but the test needs a tweak?

cpp/merkletree/merkle_tree_test.cc
pphaneuf: Okay, that's a start, yeah, but what about something like this:

```C++
EXPECT_EQ(tree.CurrentRoot(), compact.CurrentRoot());
```

PR #746, opened 2015-05-15
Add a mirroring server.

AlCutter:


cpp/server/ct-mirror.cc
```diff
@@ -0,0 +1,386 @@
...
+
+  thread sth_updater(&STHUpdater, db, server.cluster_state_controller(),
+                     &queue_mutex, &queue,
+                     fetcher_task.task()->AddChild([](Task* task) {
+                       LOG(INFO) << "STHUpdater exited.";
+                     }));
+
+  server.Run();
+
+  fetcher_task.Cancel();
```
pphaneuf: You never give the actual `fetcher_task` to anything, you only make childs with it, so there's no one to "notice" that you cancelled it, so the `Wait()` below is going to be a loooong wait! :wink:

Make this `fetcher_task.Return();`, and you'll be on your way. :smiley:

pphaneuf:
Just a one-liner fix, LGTM otherwise.

cpp/server/ct-mirror.cc
AlCutter: Done!

PR #750, opened 2015-05-18
Always participate in elections.

AlCutter:
Since mastership only controls whether a node should be sequencing [which is done purely on the contents of etcd], it doesn't actually matter what a node's local replication status is, so it shouldn't join and leave the election as local replication waxes and wanes.


pphaneuf:
More simple, more better, more LGTM!

PR #755, opened 2015-05-18
Use full hostnames for the Prometheus targets.

pphaneuf:
With the latest `prom/prometheus` Docker image, this seems to be necessary.


AlCutter:
Feels sketchy as hell, but if it gets it working for now...

PR #756, opened 2015-05-19
Make RemotePeer periodically refresh its STH

AlCutter:


cpp/fetcher/remote_peer.cc
```diff
@@ -16,29 +19,46 @@ using util::TaskHold;
...
 namespace cert_trans {
 
+DEFINE_int32(remote_peer_sth_refresh_interval_seconds, 10,
+             "Number of seconds between checks for updated STHs from the "
+             "remote peer.");
+
 
 struct RemotePeer::Impl {
-  Impl(unique_ptr<LogVerifier>&& verifier,
+  Impl(unique_ptr<LogVerifier>&& verifier, AsyncLogClient* client,
```
pphaneuf: I'm not sure what I was thinking when I wrote this class, sorry about that.

cpp/fetcher/remote_peer.cc
```diff
@@ -62,10 +82,17 @@ void RemotePeer::Impl::DoneGetSTH(
     CHECK_GE(new_sth->tree_size(), 0);
   }
 
-  sth_ = new_sth;
-  if (on_new_sth_) {
-    on_new_sth_(*sth_);
+  if (!sth_ || new_sth->timestamp() > sth_->timestamp()) {
```
pphaneuf: Nothing about the size of the tree?

cpp/fetcher/remote_peer.cc
```diff
@@ -16,29 +19,46 @@ using util::TaskHold;
...
 
+  void FetchSTH(util::Task* task);
   void DoneGetSTH(const std::shared_ptr<ct::SignedTreeHead>& on_new_sth,
                   AsyncLogClient::Status status);
 };
 
 
+void RemotePeer::Impl::FetchSTH(util::Task* task) {
+  if (CHECK_NOTNULL(task)->CancelRequested()) {
+    task->Return(util::Status::CANCELLED);
```
pphaneuf: Wrong task here, for this condition and this `Return()`, you want to use `task_`, I think. `CancelRequested()` is a bit odd (since `task` is an already completed task!), but might work. The `Return()` has to be on `task_`, for sure, though! :smiley:

cpp/fetcher/remote_peer.cc
```diff
@@ -73,12 +100,12 @@ RemotePeer::RemotePeer(
...
     util::Task* task)
-    : Peer(move(client)), impl_(new Impl(move(verifier), on_new_sth, task)) {
+    : Peer(move(client)),
+      impl_(new Impl(move(verifier), client_.get(), on_new_sth, task)) {
   TaskHold hold(task);
   task->DeleteWhenDone(impl_);
 
-  shared_ptr<SignedTreeHead> new_sth(make_shared<SignedTreeHead>());
-  client_->GetSTH(new_sth.get(), bind(&Impl::DoneGetSTH, impl_, new_sth, _1));
+  impl_->FetchSTH(task);
```
pphaneuf: This also looked weird to me, as it feels like `FetchSTH` would be liable to either try to get a status out of `task` (which wouldn't work, as it's not a done task), or "take it" to do the work and eventually call `Return()` on it (which would cause everything to fall apart!)...

If you apply my previous suggestion of just using `task_` instead of `task` in `FetchSTH`, you could just pass in `nullptr` here without problem.

cpp/fetcher/remote_peer.cc
AlCutter: I forgive you.

cpp/fetcher/remote_peer.cc
AlCutter: Nope, we already know it grew (or is > 0) from the CHECKs above, and the code which gets to hear about these new STHs will want to know about same-size-but-newer STHs (so that they can potentially become the mirror's serving STH.)

cpp/fetcher/remote_peer.cc
AlCutter: Whoops, good catch!

cpp/fetcher/remote_peer.cc
AlCutter: Done.

AlCutter:
PTAL

cpp/fetcher/remote_peer.cc
pphaneuf: Collapsed diff strikes again!

Totally agree that we're interested in same-size-but-newer, for sure, I was just worried about shrinking trees (but the `CHECK_GE` at line 80 has this covered).

pphaneuf:
LGTM

PR #757, opened 2015-05-20
Adding a small Python utility to submit certificate chains.

eranmes:


python/utilities/submit_chain/submit_chain.py
```diff
@@ -0,0 +1,116 @@
...
+from ct.crypto import pem
+from ct.crypto import verify
+from ct.proto import client_pb2
+from ct.serialization import tls_message
+
+FLAGS = gflags.FLAGS
+
+gflags.DEFINE_string("log_list", None, "List of logs to submit to.")
+gflags.DEFINE_string("chain", None, "Certificate chain to submit (PEM).")
+gflags.DEFINE_string("log_schema", "http", "Log schema (http/https)")
```
phad: s/schema/scheme/ since scheme is the term used in HTTP.

python/utilities/submit_chain/submit_chain.py
```diff
@@ -0,0 +1,116 @@
...
+gflags.MarkFlagAsRequired("chain")
+gflags.MarkFlagAsRequired("output")
+
+def _read_ct_log_list(log_list_file):
+    """Parses the log list JSON, returns a log url to key map."""
+    try:
+        log_list_json = json.loads(log_list_file)
+        log_url_to_key = {}
+        for log_info in log_list_json['logs']:
+            log_url_to_key[FLAGS.log_schema + '://' + log_info['url']] = (
```
phad: You might just require that the file already contains full URLs including scheme and avoid doing this.
Then you avoid the possibility of ending up with http://http://my.log.com when the user didn't realise their file was not supposed to include the scheme...

python/utilities/submit_chain/submit_chain.py
```diff
@@ -0,0 +1,116 @@
...
+        key_info.pem_key = pem.to_pem(log_key, 'PUBLIC KEY')
+        log_id_to_verifier[key_id] = verify.LogVerifier(key_info)
+    return log_id_to_verifier
+
+def _submit_to_all_logs(log_list, certs_chain):
+    """Submits the chain to all logs in log_list and validates SCTs."""
+    log_id_to_verifier = _map_log_id_to_verifier(log_list)
+
+    chain_der = [c.to_der() for c in certs_chain]
+    logging.info("chain der: %s", chain_der)
```
phad: are you keeping this log info ?

python/utilities/submit_chain/submit_chain.py
```diff
@@ -0,0 +1,116 @@
...
+        if res:
+            raw_scts_for_cert.append(res)
+        else:
+            logging.info("No SCT from log %s", log_url)
+
+    validated_scts = []
+    for raw_sct in raw_scts_for_cert:
+        key_id = raw_sct.id.key_id
+        try:
+            log_id_to_verifier[key_id].verify_sct(raw_sct, certs_chain[0])
```
phad: If things go spectacularly wrong then you could be calling verify_sct on None ...

python/utilities/submit_chain/submit_chain.py
```diff
@@ -0,0 +1,116 @@
...
+    scts_for_cert = [tls_message.encode(proto_sct)
+            for proto_sct in validated_scts
+            if proto_sct]
+    sct_list = client_pb2.SignedCertificateTimestampList()
+    sct_list.sct_list.extend(scts_for_cert)
+    return tls_message.encode(sct_list)
+
+def run():
+    """Submits the chain specified in the flags to all logs."""
+    logging.getLogger().setLevel(logging.DEBUG)
```
phad: Are you keeping this log level?

python/utilities/submit_chain/submit_chain.py
```diff
@@ -0,0 +1,116 @@
...
+from ct.crypto import cert
+from ct.crypto import error
+from ct.crypto import pem
+from ct.crypto import verify
+from ct.proto import client_pb2
+from ct.serialization import tls_message
+
+FLAGS = gflags.FLAGS
+
+gflags.DEFINE_string("log_list", None, "List of logs to submit to.")
```
pphaneuf: Should mention what they are separated with here.

pphaneuf: Or, even better, use `DEFINE_multistring`.

pphaneuf: And... reading the code further down, I find out that this is really a file. I'm no longer sure what this should be, I'm only sure at least the doc string should be improved. :wink:

python/utilities/submit_chain/submit_chain.py
pphaneuf: I would +1 that, especially if it's allowed to host a log on any arbitrary URL (like http://example.com/submariner/, say, where getting the STH would be http://example.com/submariner/ct/v1/get-sth).

python/utilities/submit_chain/submit_chain.py
pphaneuf: As per the comment below, I'd just take it out entirely.

pphaneuf:
Also, BTW, it looks like you haven't pulled `master` recently, might want to do so and rebase?

python/utilities/submit_chain/submit_chain.py
eranmes: Acknowledged, but that's the file format we currently publish on www.certificate-transparency.org/known-logs - without the scheme.

python/utilities/submit_chain/submit_chain.py
eranmes: changed schema -> scheme

python/utilities/submit_chain/submit_chain.py
eranmes: Nope, removed.

python/utilities/submit_chain/submit_chain.py
eranmes: In line 73 I'm checking if res is not None and only then adding it to raw_scts_for_cert.

python/utilities/submit_chain/submit_chain.py
eranmes: Changed to INFO.

python/utilities/submit_chain/submit_chain.py
eranmes: Done.

eranmes:
@phad , @pphaneuf PTAL - addressed your comments in a follow-up commit (which will be squashed prior to merging).

pphaneuf:
LGTM

PR #759, opened 2015-05-21
Use libevhtp for outgoing HTTP requests.

AlCutter:


README.md
```diff
@@ -54,6 +54,9 @@ installing. If you can install the missing header manually, it should
 work.
 
  - [libevent](http://libevent.org/) (tested with 2.0.21-stable)
+ - [libevhtp](https://github.com/ellzey/libevhtp) (tested with 1.2.10)
+ If building libevhtp from source, you may need to disable the regex support
+ with the following cmake flag: -DEVHTP_DISABLE_REGEX:STRING=ON
```
pphaneuf: Put some backticks around the flag?

configure.ac
```diff
@@ -61,6 +63,7 @@ AS_IF([test -n "$missing_gmock"],
 # Checks for libraries.
 AC_SEARCH_LIBS([__b64_ntop], [resolv])
 AC_SEARCH_LIBS([pthread_create], [pthread])
+AC_SEARCH_LIBS([evhtp_ssl_init], [evhtp])
```
pphaneuf: So it turns out you're not using this function, despite its name, might be something internal? If it's an internal function, that might disappear in the future, and it'd be generally better if it was something *we* used.

How about `evhtp_make_request`?

cpp/util/libevent_wrapper.h
```diff
@@ -3,9 +3,12 @@
 
 #include <atomic>
 #include <chrono>
+
```
pphaneuf: I'd keep them together, as they're the "outside" headers.

cpp/net/connection_pool.h
```diff
@@ -16,27 +16,40 @@ namespace cert_trans {
...
+struct evhtp_connection_deleter {
+  void operator()(evhtp_connection_t* con) const {
+    evhtp_connection_free(con);
   }
 };
 
 
-typedef std::unique_ptr<evhttp_connection, evhttp_connection_deleter>
-    evhttp_connection_unique_ptr;
+typedef std::unique_ptr<evhtp_connection_t, evhtp_connection_deleter>
```
pphaneuf: Since this is no longer re-used, specifying it just once inline would be fine? Also, how about doing without `unique_ptr` altogether (for this), just make it a const raw pointer in `Connection`, and free it in `~Connection`?

cpp/net/connection_pool.cc
```diff
@@ -16,45 +17,57 @@ DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
 namespace cert_trans {
 namespace internal {
 
+ConnectionPool::Connection::Connection(evhtp_connection_t* conn,
+                                       HostPortPair&& other_end)
+    : conn_(CHECK_NOTNULL(conn)), other_end_(other_end) {
+}
+
+
+evhtp_connection_t* ConnectionPool::Connection::connection() const {
```
pphaneuf: Having this one in the header would let it inline where it's used, maybe a good idea?

README.md
AlCutter: ticked

configure.ac
AlCutter: That's a public API function, but it's used for writing SSL *servers* instead.
Changed to `evhtp_make_request`.

cpp/util/libevent_wrapper.h
AlCutter: Spurious, removed.

cpp/net/connection_pool.h
AlCutter: Dinged the `typedef`, but I'll keep using the `unique_ptr` rather than the manual memory management unless there's a really good reason not to.

cpp/net/connection_pool.cc
AlCutter: Oke doke, moved.

AlCutter:
PTAL

configure.ac
```diff
@@ -91,9 +94,13 @@ AC_SEARCH_LIBS([event_base_dispatch], [event],, [missing_libevent=1],
                [$save_LIBS])
 AC_SEARCH_LIBS([evthread_use_pthreads], [event_pthreads],, [missing_libevent=1],
                [$save_LIBS])
+AC_SEARCH_LIBS([bufferevent_openssl_socket_new], [event_openssl],, [missing_libevent_openssl=1],
+               [$save_LIBS])
```
pphaneuf: I'll let it slide because I know that's coming right next, but ideally, you wouldn't have added this SSL stuff that goes completely unused in this pull request (same story in `Makefile.am`). :wink:

cpp/net/connection_pool.cc
```diff
@@ -16,45 +17,98 @@ DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
...
 
-evhttp_connection_unique_ptr ConnectionPool::Get(const URL& url) {
+namespace {
+
+
+evhtp_res ConnectionClosedHook(evhtp_connection_t* conn, void* arg) {
+  CHECK_NOTNULL(conn);
+  CHECK_NOTNULL(arg);
+  ConnectionPool::Connection* const c(
+      reinterpret_cast<ConnectionPool::Connection*>(arg));
```
pphaneuf: Can't this be a `static_cast`?

cpp/net/connection_pool.cc
```diff
@@ -16,45 +17,98 @@ DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
...
     : base_(CHECK_NOTNULL(base)), cleanup_scheduled_(false) {
 }
 
 
-evhttp_connection_unique_ptr ConnectionPool::Get(const URL& url) {
+namespace {
+
+
+evhtp_res ConnectionClosedHook(evhtp_connection_t* conn, void* arg) {
+  CHECK_NOTNULL(conn);
```
pphaneuf: Maybe more `CHECK_EQ(conn, conn_.get())`?

pphaneuf: Never mind, I thought this was in `Connection` for a moment...

pphaneuf: Might be an idea to roll it in with `ReleaseConnection`, making this a static member of `Connection`, it would have access to `conn_` to check this, and could just do the `conn_.release()` itself?

cpp/net/connection_pool.cc
```diff
@@ -16,45 +17,98 @@ DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
 namespace cert_trans {
 namespace internal {
 
+ConnectionPool::Connection::Connection(evhtp_connection_t* conn,
+                                       HostPortPair&& other_end)
+    : conn_(CHECK_NOTNULL(conn)), other_end_(other_end) {
```
pphaneuf: Don't you need it to be `other_end_(move(other_end))` if you want to avoid the copy?

cpp/net/connection_pool.cc
AlCutter: Not entirely sure what you mean by rolling it in with `ReleaseConnection`, but I've made it a static member which allows the `CHECK_EQ` and also has the nice side effect of allowing me to make `ReleaseConnection()` private too.

cpp/net/connection_pool.cc
AlCutter: Done

cpp/net/connection_pool.cc
pphaneuf: By "rolling it in", I meant that you wouldn't need a helper method to do `conn_.reset()`, it could just *do it*.

AlCutter:
PTAL

cpp/net/connection_pool.cc
AlCutter: Ah, gotcha - I've, er, "rolled that in" to the Comments commit :)

cpp/net/connection_pool.cc
```diff
@@ -16,45 +17,92 @@ DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
...
 }
 
 
-evhttp_connection_unique_ptr ConnectionPool::Get(const URL& url) {
+// static
+evhtp_res ConnectionPool::Connection::ConnectionClosedHook(
+    evhtp_connection_t* conn, void* arg) {
+  CHECK_NOTNULL(conn);
+  CHECK_NOTNULL(arg);
+  VLOG(1) << "Releasing connection " << conn;
```
pphaneuf: Is there anywhere else where this address will have been printed? As in, something that could be correlated with another log message? Maybe logging `c` would be more useful? I didn't fully check, not sure that gets logged either... If not, maybe logging the host:port (using `c`) could be *actually* useful? :wink:

cpp/net/connection_pool.cc
```diff
@@ -73,14 +121,13 @@ void ConnectionPool::Cleanup() {
   lock_guard<mutex> lock(lock_);
   cleanup_scheduled_ = false;
 
-  // std::map<HostPortPair, std::deque<evhttp_connection_unique_ptr>> conns_;
+  // std::map<HostPortPair, std::deque<unique_ptr<Connection>>> conns_;
```
pphaneuf: This cleanup function should probably clean up any dead connections first?

cpp/net/connection_pool.cc
```diff
@@ -16,45 +17,92 @@ DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
...
-  const HostPortPair key(url.Host(), url.Port() != 0 ? url.Port() : 80);
+  HostPortPair key(url.Host(), url.Port() != 0 ? url.Port() : 80);
   lock_guard<mutex> lock(lock_);
 
   auto it(conns_.find(key));
+
+  if (it != conns_.end() && !it->second.empty()) {
+    // Do a sweep and remove any dead connections
+    for (auto deque_it(it->second.begin()); deque_it != it->second.end();) {
+      CHECK(*deque_it);
```
pphaneuf: I think you're trying to do `CHECK_NOTNULL(deque_it->get())` here, right? That might give you a nicer log message, if it ever failed, I think...

cpp/net/connection_pool.cc
AlCutter: Ah, there used to be other locations, but they went away.
Changed to print out the host:port.

cpp/net/url_fetcher.cc
```diff
@@ -162,7 +177,7 @@ void State::RequestDone(evhttp_request* req) {
     return;
```
pphaneuf: Does evhtp ever give us those silly null requests? I hope not, they're the first thing I'd have changed if I made a new library to replace evhttp... :stuck_out_tongue:

cpp/net/url_fetcher.cc
```diff
@@ -174,18 +189,19 @@ void State::RequestDone(evhttp_request* req) {
   }
```
pphaneuf: Leaking `req` when returning inside this block? Maybe use a `unique_ptr` as a deleter?

cpp/net/url_fetcher.cc
```diff
@@ -162,7 +177,7 @@ void State::RequestDone(evhttp_request* req) {
     return;
   }
 
-  response_->status_code = evhttp_request_get_response_code(req);
+  response_->status_code = req->status;
   if (response_->status_code < 100) {
     // TODO(pphaneuf): According to my reading of libevent, this is
     // most likely to be a connection refused?
```
pphaneuf: This might not happen anymore, or mean something else?

pphaneuf:
Okay, mostly done, I think.

cpp/net/connection_pool.cc
AlCutter: Yep, I'm checking the same thing.
Should never happen :)

cpp/net/connection_pool.cc
AlCutter: Factored out and called from here too.

cpp/net/url_fetcher.cc
AlCutter: I don't actually know, would be lovely to be able to remove all that stuff... but that's a story for another PR I think.

cpp/net/url_fetcher.cc
AlCutter: Good catch, deleter'd.

cpp/net/url_fetcher.cc
AlCutter: Punted for now.

pphaneuf:
LGTM

PR #760, opened 2015-05-22
wait_for() and seconds::max() don't place nice.

AlCutter:
`wait_for(..., seconds::max())` returns immediately, this fixes that and stops the workers spinning.


pphaneuf:
LGTM

PR #764, opened 2015-05-26
Use TCMalloc

AlCutter:


configure.ac
```diff
@@ -96,6 +96,16 @@ AS_IF([test -n "$missing_libevent"],
...
 
+# TCMalloc gubbins
+AC_ARG_WITH([tcmalloc],
+            [AS_HELP_STRING([--without-tcmalloc],
+                            [disable tcmalloc for memory allocations])],
+            [],
+            [with_tcmalloc=yes])
+AS_IF([test "x$with_tcmalloc" != xno],
+      [AC_CHECK_LIB([tcmalloc], [malloc],,
+                    [AC_MSG_FAILURE([no tcmalloc found (use --without-tcmalloc to disable)])])])
```
pphaneuf: I used to, but I'm not really in favour of these `--with-foobar` flags that much anymore, especially for a link-flags only thing (there's no `#ifdef` in the code to enable it, for example). I'd drop this, and just add `LIBS=-ltcmalloc` to the `configure` in `.travis.yml`.

pphaneuf: Or the `env` line, where the `CPPFLAGS` and others get set?

AlCutter: If we think that it's a good idea to run with tcmalloc, then it seems like we should encourage it - leaving it up to someone to remember to set `LIBS` to correctly in order to link in something which we recommend but which they won't be able to easily tell when/if they forget to do so doesn't seem like a good idea.

Note that this is a `--without-foobar` flag for the "exceptional" cases where it's a PITA/doesn't-work-on-my-platform scenarios.

Do you have a good example of why you're not in favour of these types of things?

pphaneuf: It's just because you end up getting a flurry of these knobs, which do something that's specific to the package you're looking at (by which I mean, in this case, "certificate-transparency", not tcmalloc), when these sorts of add-on libraries will usually have instructions on how to enable them with Autotools build systems (with the generic `LDFLAGS`/`LIBS` flags).

These knobs are also kind of hard to do right (which is part of my hatred for them). For example, I think `--without-tcmalloc` won't have any effect, the way you've done it. :wink: 

If you want to be *really* nice, you'd also do them take `--with-tcmalloc=<PATH>`, to get it from that place, which is yet more hassle. So I just escape the whole thing by saying "screw it, if you want a thing in `LIBS`, just put it in `LIBS`".

pphaneuf: On the other hand, when I'm not maintaining them *and they work correctly*, I love them. :wink:

pphaneuf: I think it might work after all. I can't bloody tell. :stuck_out_tongue:

pphaneuf:
LGTM

PR #766, opened 2015-05-29
Add SSL support to UrlFetcher.

AlCutter:
In an attempt to avoid blindly falling into the many OpenSSL gotchas, most of the hostname checking stuff is glommed from https://wiki.openssl.org/index.php/Hostname_validation and https://github.com/iSECPartners/ssl-conservatory/tree/master/openssl, and re-uses the cURL hostname checking routines referenced there.

By way of simple sanity checking, ct-mirror happily works over SSL with this against Pilot (etc.) and verifies the *.googleapis.com cert ok, it also correctly rejects the cert served at https://tv.eurosport.com (since that doesn't contain a matching CN/SAN.)

### DIFF
--- a/cpp/net/connection_pool.cc
+++ b/cpp/net/connection_pool.cc
@@ -3,24 +3,140 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 
+#include "util/openssl_util.h"
+
+extern "C" {
+#include "third_party/curl/hostcheck.h"
+#include "third_party/isec_partners/openssl_hostname_validation.h"
+}  // extern "C"
+
+
 using std::bind;
 using std::lock_guard;
+using std::make_pair;
+using std::map;
 using std::move;
 using std::mutex;
 using std::pair;
+using std::placeholders::_1;
+using std::placeholders::_2;
 using std::string;
 using std::unique_lock;
 using std::unique_ptr;
+using util::ClearOpenSSLErrors;
+using util::DumpOpenSSLErrorStack;
 
 DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
              "maximum number of URL fetcher connections per host:port");
+DEFINE_string(trusted_root_certs, "/etc/ssl/certs/ca-certificates.crt",
+              "Location of trusted CA root certs for outgoing SSL "
+              "connections.");
+
 
 namespace cert_trans {
 namespace internal {
 
+
+// static
+mutex ConnectionPool::Connection::ssl_mutex_;
+// static
+map<const SSL*, const ConnectionPool::Connection*>
+    ConnectionPool::Connection::connections_by_ssl_;
+
+
+// static
+int ConnectionPool::Connection::SSLVerifyCallback(const int preverify_ok,
+                                                  X509_STORE_CTX* x509_ctx) {
+  CHECK_NOTNULL(x509_ctx);
+  X509* const server_cert(
+      CHECK_NOTNULL(X509_STORE_CTX_get_current_cert(x509_ctx)));
+
+  if (preverify_ok == 0) {
+    const int err(X509_STORE_CTX_get_error(x509_ctx));
+    char buf[256];
+    X509_NAME_oneline(X509_get_subject_name(server_cert), buf, 256);
+
+    LOG(WARNING) << "OpenSSL failed to verify cert for " << buf << ": "
+                 << X509_verify_cert_error_string(err);
+    return preverify_ok;
+  }
+
+  // Only do extra checks (i.e. hostname matching) for the end-entity cert.
+  const int depth(X509_STORE_CTX_get_error_depth(x509_ctx));
+  if (depth > 0) {
+    return preverify_ok;
+  }
+
+  const SSL* const ssl(static_cast<SSL*>(CHECK_NOTNULL(
+      X509_STORE_CTX_get_ex_data(x509_ctx,
+                                 SSL_get_ex_data_X509_STORE_CTX_idx()))));
+  const ConnectionPool::Connection* conn;
+  {
+    lock_guard<mutex> lock(ssl_mutex_);
+    auto it(connections_by_ssl_.find(ssl));
+    CHECK(it != connections_by_ssl_.end());
+    conn = it->second;
+  }
+  CHECK_NOTNULL(conn);
+
+  const HostnameValidationResult hostname_valid(
+      validate_hostname(conn->other_end().first.c_str(), server_cert));
+  if (hostname_valid != MatchFound) {
+    string error;
+    switch (hostname_valid) {
+      case MatchFound:
+        LOG(FATAL) << "Shouldn't get here.";
+        break;
+      case MatchNotFound:
+        error = "certificate doesn't match hostname";
+        break;
+      case NoSANPresent:
+        // I don't think we should ever see this, should be handled inside
+        // validate_hostname()
+        error = "no SAN present";
+        break;
+      case MalformedCertificate:
+        error = "certificate is malformed";
+        break;
+      case Error:
+        error = "unknown error";
+        break;
+    }
+    conn->connection()->request->status = kSSLErrorStatus;
+    LOG_EVERY_N(WARNING, 100)
+        << "Failed to validate SSL certificate: " << error << " : "
+        << DumpOpenSSLErrorStack();
+    ClearOpenSSLErrors();
+    return 0;
+  }
+  return 1;
+}
+
+
 ConnectionPool::Connection::Connection(evhtp_connection_t* conn,
                                        HostPortPair&& other_end)
-    : conn_(CHECK_NOTNULL(conn)), other_end_(move(other_end)) {
+    : conn_(CHECK_NOTNULL(conn)),
+      other_end_(move(other_end)),
+      ssl_(conn_->ssl) {
+  if (ssl_) {
+    {
+      lock_guard<mutex> lock(ssl_mutex_);
+      CHECK(connections_by_ssl_.insert(make_pair(ssl_, this)).second);
+    }
+    SSL_set_tlsext_host_name(ssl_, other_end_.first.c_str());
+  }
+}
+
+
+ConnectionPool::Connection::~Connection() {
+  // ssl_ is likely pointing to freed memory at this point, so while we use the
+  // value of the pointer itself here as a key don't try to dereference it!
+  if (ssl_) {
+    {
+      lock_guard<mutex> lock(ssl_mutex_);
+      CHECK_EQ(1, connections_by_ssl_.erase(ssl_));
+    }
+  }
 }
 
 
@@ -29,22 +145,60 @@ const HostPortPair& ConnectionPool::Connection::other_end() const {
 }
 
 
+void ConnectionPool::Connection::ReleaseConnection() {
+  conn_.release();
+  // Do not null out ssl_, despite the fact that it's a dangling pointer now we
+  // need it in the d'tor since the value of the pointer itself is a key into a
+  // map which needs to be updated.
+}
+
+
 ConnectionPool::ConnectionPool(libevent::Base* base)
-    : base_(CHECK_NOTNULL(base)), cleanup_scheduled_(false) {
+    : base_(CHECK_NOTNULL(base)),
+      cleanup_scheduled_(false),
+      ssl_ctx_(CHECK_NOTNULL(SSL_CTX_new(TLSv1_client_method())),
+               SSL_CTX_free) {
+  // Try to load trusted root certificates.
+  // TODO(alcutter): This is probably Linux specific, we'll need other sections
+  // for OSX etc.
+  if (SSL_CTX_load_verify_locations(ssl_ctx_.get(),
+                                    FLAGS_trusted_root_certs.c_str(),
+                                    nullptr) != 1) {
+    DumpOpenSSLErrorStack();
+    LOG(FATAL) << "Couldn't load trusted root certificates.";
+  }
+
+  SSL_CTX_set_verify(ssl_ctx_.get(), SSL_VERIFY_PEER,
+                     Connection::SSLVerifyCallback);
 }
 
 
 // static
-evhtp_res ConnectionPool::Connection::ConnectionClosedHook(
-    evhtp_connection_t* conn, void* arg) {
+evhtp_res ConnectionPool::Connection::ConnectionErrorHook(
+    evhtp_connection_t* conn, evhtp_error_flags errtype, void* arg) {
   CHECK_NOTNULL(conn);
   CHECK_NOTNULL(arg);
+  CHECK(libevent::Base::OnEventThread());
   ConnectionPool::Connection* const c(
       static_cast<ConnectionPool::Connection*>(arg));
-  VLOG(1) << "Releasing connection to " << c->other_end().first << ":"
+  VLOG(1) << "Releasing errored connection to " << c->other_end().first << ":"
           << c->other_end().second;
+
   CHECK_EQ(conn, c->connection());
-  c->conn_.release();
+
+  // Need to let the client know their request has failed, seems evhtp doesn't
+  // do that by default so we'll call the request done callback here.
+  if (conn->request) {
+    // If someone hasn't already modified the default status, set it to a
+    // generic "something went wrong" value here:
+    if (conn->request->status == 200) {
+      conn->request->status = kUnknownErrorStatus;
+    }
+    conn->request->cb(conn->request, conn->request->cbarg);
+    conn->request = nullptr;
+  }
+
+  c->ReleaseConnection();
   return EVHTP_RES_OK;
 }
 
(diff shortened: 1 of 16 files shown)


cpp/util/fake_etcd_test.cc
```diff
@@ -951,5 +951,14 @@ TEST_F(FakeEtcdTest, GetWaitRecursive) {
 
 int main(int argc, char** argv) {
   cert_trans::test::InitTesting(argv[0], &argc, &argv, true);
+  // Set-up OpenSSL for multithreaded use:
+  evhtp_ssl_use_threads();
```
pphaneuf: Maybe add at least `evhtp_ssl_use_threads()` to `cpp/util/testing.cc`?

AlCutter: Done.

cpp/net/connection_pool.cc
```diff
@@ -90,12 +229,14 @@ unique_ptr<ConnectionPool::Connection> ConnectionPool::Get(const URL& url) {
 
   if (it == conns_.end() || it->second.empty()) {
     VLOG(1) << "new evhtp_connection for " << key.first << ":" << key.second;
-    unique_ptr<ConnectionPool::Connection> conn(
-        new Connection(base_->HttpConnectionNew(key.first, key.second),
-                       move(key)));
-    evhtp_set_hook(&conn->connection()->hooks, evhtp_hook_on_connection_fini,
+    unique_ptr<ConnectionPool::Connection> conn(new Connection(
+        url.Protocol() == "https"
+            ? base_->HttpsConnectionNew(key.first, key.second, ssl_ctx_.get())
```
pphaneuf: This here is blocking at the moment, which could be fixed in the future, but for now, calls to this method are on the libevent thread, where blocking is Very Bad (tm).

Keeping it for a follow up pull request, as discussed (my head most definitely has a bounded capacity!), but we'll have to work around this, otherwise untold horrors will get us later.

cpp/util/libevent_wrapper.cc
```diff
@@ -82,6 +85,10 @@ Base::Base()
       wake_closures_(event_new(base_.get(), -1, 0, &Base::RunClosures, this),
                      &event_free) {
   evthread_make_base_notifiable(base_.get());
+
+  // So much stuff breaks if there's not a Dns server around to keep the
```
pphaneuf: s/server/client/

test/create_url_fetcher_test_certs.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+cd ${DIR}
+. generate_certs.sh
```
pphaneuf: Why not `. ${DIR}/generate_certs.sh`? Might be missing a few `${DIR}` below...

I have this horrible feeling that this will break building out of tree, but I'll just fix it up later, I suppose.

cpp/net/url_fetcher_test.cc
```diff
@@ -0,0 +1,162 @@
...
+
+namespace {
+
+
+pid_t localhost_pid(0);
+pid_t nonlocalhost_pid(0);
+
+
+void KillAllOpenSSLServers() {
+  LOG(WARNING) << "aTEXIT";
```
pphaneuf: That was for debugging?

cpp/net/connection_pool.h
```diff
@@ -13,6 +14,12 @@
 #include "util/libevent_wrapper.h"
 
 namespace cert_trans {
+
+
+const int kUnknownErrorStatus = 0;
+const int kSSLErrorStatus = 1;
```
pphaneuf: Comment those two? Even though their existence somewhat frighten me...

cpp/util/libevent_wrapper.cc
```diff
@@ -206,6 +213,66 @@ evhtp_connection_t* Base::HttpConnectionNew(const string& host,
...
+  hints.ai_family = AF_INET;
+  hints.ai_socktype = SOCK_STREAM;
+  const int resolved(getaddrinfo(host.c_str(), AF_UNSPEC, &hints, &info));
+  if (resolved != 0) {
+    LOG(WARNING) << "Failed to resolve HTTPS hostname " << host << ": "
+                 << gai_strerror(resolved);
+    return nullptr;
+  }
+
+  char addr_str[INET6_ADDRSTRLEN];
```
pphaneuf: Could you move this one down to where it's used?

cpp/util/libevent_wrapper.cc
```diff
@@ -82,6 +85,10 @@ Base::Base()
       wake_closures_(event_new(base_.get(), -1, 0, &Base::RunClosures, this),
                      &event_free) {
   evthread_make_base_notifiable(base_.get());
+
+  // So much stuff breaks if there's not a Dns server around to keep the
+  // event loop doing stuff that we may as well just have one from the get go.
+  GetDns();
```
pphaneuf: Maybe delete the one in the event pump thread? :sob:

cpp/net/connection_pool.cc
```diff
@@ -30,20 +132,57 @@ const HostPortPair& ConnectionPool::Connection::other_end() const {
 
 
 ConnectionPool::ConnectionPool(libevent::Base* base)
-    : base_(CHECK_NOTNULL(base)), cleanup_scheduled_(false) {
+    : base_(CHECK_NOTNULL(base)),
+      cleanup_scheduled_(false),
+      ssl_ctx_(CHECK_NOTNULL(SSL_CTX_new(TLSv1_client_method())),
+               SSL_CTX_free) {
+  // Try to load trusted root certificates.
+  // TODO(alcutter): This is probably Linux specific, we'll need other sections
```
pphaneuf: Doesn't that comment belong more with the flag definition?

cpp/net/connection_pool.cc
```diff
@@ -30,20 +132,57 @@ const HostPortPair& ConnectionPool::Connection::other_end() const {
...
       static_cast<ConnectionPool::Connection*>(arg));
   VLOG(1) << "Releasing connection to " << c->other_end().first << ":"
           << c->other_end().second;
+
   CHECK_EQ(conn, c->connection());
+
+  if (conn->ssl) {
+    {
+      lock_guard<mutex> lock(ssl_mutex_);
+      CHECK_EQ(1, connections_by_ssl_.erase(conn->ssl));
```
pphaneuf: This is the only spot this is called, and it's in a method called "error hook", so I'm wondering who cleans this up when there's no error? :wink: 

Cleaning this up in the `ConnectionPool::Connection` destructor would be logical, as this is added in its constructor.

pphaneuf:
I think I'll go have a nap now. Or a quiet cry.

cpp/net/connection_pool.cc
AlCutter: Agreed, that's why the big Easy Spot/Remove `----8<----` section is in there (or, rather, in the actual blocking method), along with a reference to the libevt PR which will Do It Right - I'll ping that guy and see if he's had any luck on that front.

cpp/util/libevent_wrapper.cc
AlCutter: It's all just a question of perspective. :)
Done.

test/create_url_fetcher_test_certs.sh
AlCutter: Because the cert/ca scripts in that directory assume that they're run *from* that directory, and I didn't want to roll fixing that up into this PR.
I've added a `TODO` and sprinkled a few more `${DIR}` around in anticipation.

How are all the diodes down your left side doing today, BTW?

cpp/net/url_fetcher_test.cc
AlCutter: Totally wouldn't work without that line.

*cough*

Removed.

cpp/net/connection_pool.h
AlCutter: Commented.

Alternative solutions welcome on a postcard to the usual address.

cpp/util/libevent_wrapper.cc
AlCutter: Moved.

cpp/util/libevent_wrapper.cc
AlCutter: The more the merrier?
Deleted.

cpp/net/connection_pool.cc
AlCutter: Not really; I could see the flag being set to `System` (or whatever) on OS X/Pyramid OS/...  (or another `FLAG_use_system_ca_roots`  - bearing in mind that you may wish to NOT use the system roots and instead load them from a file.)
But the code for what `System` would do is not going to be this code (unless you're running Debian) and *that's* what this comment is about and why it's here and not on the flag.

cpp/net/url_fetcher_test.cc
```diff
@@ -0,0 +1,162 @@
...
+TEST_F(UrlFetcherTest, TestNotListening) {
+  UrlFetcher::Request req(URL("https://localhost:63544"));
+  UrlFetcher::Response resp;
+
+  SyncTask task(&pool_);
+  fetcher_->Fetch(req, &resp, task.task());
+  task.Wait();
+  EXPECT_THAT(task.status(), StatusIs(util::error::FAILED_PRECONDITION));
+  EXPECT_EQ(kUnknownErrorStatus, resp.status_code);
+}
```
benlaurie: How about tests for wildcards? Subdomains? That kind of stuff...

cpp/net/connection_pool.cc
AlCutter: As I mentioned when we discussed, this used to be a "connection finished" hook, which is why it's the only place doing it, but moving it to the d'tor would be nice *except* that the `evhtp_connection_t` is going to get free'd when the hook returns so the d'tor can't dereference that to get the `*ssl`.

But I guess I could store the pointer in `Connection`, knowing full well that it'll be dangling when I come to "use" it, and comment the hell out of it.

cpp/net/url_fetcher_test.cc
AlCutter: Those are hard without messing with the system's DNS/hosts file.
I did try using things like` something.localhst.co.uk` (which resolves to 127.0.0.1 and causes the request to hit a local openssl s_server instance with a * cert, but some resolvers seem to dislike that and error out rather than returning the address.)

I'll see if I can come up with something; stub out the name resolution inside `libevent::Base` maybe.

benlaurie: On 12 June 2015 at 11:42, Al Cutter <notifications@github.com> wrote:

> In cpp/net/url_fetcher_test.cc
> <https://github.com/google/certificate-transparency/pull/766#discussion_r32306155>
> :
>
> > +  task.Wait();
> > +  EXPECT_THAT(task.status(), StatusIs(util::error::FAILED_PRECONDITION));
> > +  EXPECT_EQ(kSSLErrorStatus, resp.status_code);
> > +}
> > +
> > +
> > +TEST_F(UrlFetcherTest, TestNotListening) {
> > +  UrlFetcher::Request req(URL("https://localhost:63544"));
> > +  UrlFetcher::Response resp;
> > +
> > +  SyncTask task(&pool_);
> > +  fetcher_->Fetch(req, &resp, task.task());
> > +  task.Wait();
> > +  EXPECT_THAT(task.status(), StatusIs(util::error::FAILED_PRECONDITION));
> > +  EXPECT_EQ(kUnknownErrorStatus, resp.status_code);
> > +}
>
> Those are hard without messing with the system's DNS/hosts file.
> I did try using things likesomething.localhst.co.uk (which resolves to
> 127.0.0.1 and causes the request to hit a local openssl s_server instance
> with a * cert, but some resolvers seem to dislike that and error out rather
> than returning the address.)
>
> I'll see if I can come up with something; stub out the name resolution
> inside libevent::Base maybe.
>

Yeah. Or maybe some kind of standalone test that shows the matching works
without having to make a connection (not as good, but might be good enough,
combined with an inline test to ensure matching is called at all!).

cpp/net/connection_pool.cc
pphaneuf: I was kind of assuming here that the "Linux specific" bit (in the context that we're using OpenSSL, and not some other possibly platform-specific library) was the filename of where we got the trusted root certs.

cpp/net/connection_pool.cc
pphaneuf: Ah, funky ownership strikes again, where `ConnectionPool::Connection` doesn't really own the `evhtp_connection_t`... I noticed the `c->conn_.release();` below just now.

Maybe duplicating this bit is all right? You need to clean it up here, sure, but if by the time you're in the `ConnectionPool::Connection` destructor, `conn_` is non-null (and has an `ssl`), remove it?

cpp/net/url_fetcher_test.cc
AlCutter: Ok - I added a bunch more tests (even crazy IP Address subject ones), let me know if there are more you'd like to see.

cpp/net/connection_pool.cc
pphaneuf: That would be the best, if a new version of libevhtp would have the fix, but I'm thinking we're likely to need a workaround before then (as in, ASAP).

cpp/net/connection_pool.cc
```diff
@@ -29,21 +145,59 @@ const HostPortPair& ConnectionPool::Connection::other_end() const {
...
+    // If someone hasn't already modified the default status, set it to a
+    // generic "something went wrong" value here:
+    if (conn->request->status == 200) {
+      conn->request->status = kUnknownErrorStatus;
+    }
+    conn->request->cb(conn->request, conn->request->cbarg);
+    conn->request = nullptr;
+  }
+
   c->conn_.release();
```
pphaneuf: Shouldn't you be calling `c->ReleaseConnection()` here?

cpp/net/connection_pool.cc
```diff
@@ -29,21 +145,59 @@ const HostPortPair& ConnectionPool::Connection::other_end() const {
 }
 
 
+void ConnectionPool::Connection::ReleaseConnection() {
+  conn_.release();
+  // Do not null out ssl_, despite the fact that it's a dangling pointer now we
+  // need it in the d'tor since the value of the pointer itself is a key into a
+  // map which needs to be updated.
```
pphaneuf: With a method like this to wrap the `unique_ptr` a bit (maybe `conn_` could be made private, with an accessor, to avoid mistakes?), why not do the removal from `connections_by_ssl_` here (before releasing)?

cpp/net/connection_pool.cc
AlCutter: Yes, I thought I had. Squashed that into the Comments commit.

cpp/net/connection_pool.cc
AlCutter: `conn_` *is* private, and do you mean an accessor like the existing `connection() const`?
`ReleaseConnection()` is private too, but the 'Pool has `friend` access.

What about a `Connection` which doesn't error and is just destroyed by the user? Would still need to clean up the map in the d'tor, so better to just do it in one place I think.

pphaneuf: I wasn't thinking straight... I meant to have the removal from the map centralized somewhere, like there, and possibly calling that from the destructor, but if it's released, then it won't be freed, and... ugh, you know what I mean. :stuck_out_tongue:

AlCutter: But then you can't do the `CHECK` on the erase of course, which is why it is like it is :)

pphaneuf: Well, you'd only erase if you have a `conn_`. And if you do, then you should `CHECK`. But the whole centralizing business might not work out so well.

cpp/util/libevent_wrapper.cc
```diff
@@ -76,12 +80,71 @@ struct HttpServer::Handler {
...
+
+    char addr_str[INET6_ADDRSTRLEN];
+    inet_ntop(res->ai_family, addr, addr_str, INET6_ADDRSTRLEN);
+    freeaddrinfo(info);
+    return string(addr_str);
+  }
+};
+
+
+Base::Base() : Base(unique_ptr<Resolver>(new ResolverImpl)) {
```
pphaneuf: You don't need the `unique_ptr` bit here, no? Just `new ResolverImpl` ought to do it?

Also, this `unique_ptr<Resolver>` looks suspiciously like an `std::function<std::string(const std::string& host)>`... :wink:

AlCutter: Yeah, turns out you do need it - I thought I might not originally too.

pphaneuf:
For the testing that @benlaurie was mentioning, I was taking this to be some testing of `validate_hostname` (and maybe `Curl_cert_hostcheck`?)?

cpp/util/libevent_wrapper.cc
pphaneuf: Oh, because you're passing it to the other constructor here, I thought you were initializing `resolver_` directly (but this is much better!).

AlCutter:
Right, but doing it this way also checks that the validation code is actually being called by our code too.

pphaneuf:
Looks good (enough) to me.

Would definitely need a follow up PR to punt the blocking DNS resolution to another thread, unless the evhtp PR to fix that is merged and released quickly enough (which seems questionable).

PR #774, opened 2015-06-09
Capping of entries in the EV whitelist generator

eranmes:
This changes allows re-generating the EV whitelist, excluding
expired certificates and ones that were added after the previous
whitelist was generated.

There's also a small bugfix in case the chain submitted consisted
of a single certificate (so the extra_data is empty). That happens
when a root was submitted to the log.

### DIFF
--- a/python/utilities/ev_whitelist/create_ev_whitelist_zip.py
+++ b/python/utilities/ev_whitelist/create_ev_whitelist_zip.py
@@ -21,13 +21,19 @@ CRX_URL = (
 TRUNCATED_HASH_LENGTH = 8
 GOLOMB_M_PARAMETER = 2 ** 47
 NUM_FETCHING_PROCESSES = 40
-LOGS_LIST = [
-    "https://ct.googleapis.com/pilot",
-    "https://ct.googleapis.com/aviator",
-    "https://ct1.digicert-ct.com/log",
-    "https://log.certly.io",
-    "https://ct.izenpe.com",
-]
+# Map between a log's URL and the index of the cut-off point after which
+# no new EV certificates will be whitelisted.
+# These entries are basically the index of the last EV cert that was found on
+# January 1st, when the EV whitelist was finalized.
+# Ultimately this allows re-generation of the whitelist, excluding certificates
+# that have recently expired, thus reducing its size.
+LOGS_LIST = {
+    "https://ct.googleapis.com/pilot":6082932,
+    "https://ct.googleapis.com/aviator":5360225,
+    "https://ct1.digicert-ct.com/log":1056,
+    "https://log.certly.io": 4729,
+    "https://ct.izenpe.com": 1,
+}
 
 FLAGS = gflags.FLAGS
 
@@ -113,7 +119,7 @@ def generate_new_ev_hashes_whitelist(logs_list):
     """Scans the provided logs_list and generates a compressed EV certs
     whitelist for all EV certs in all the logs."""
     all_hashes_set = set()
-    for log_url in logs_list:
+    for log_url, last_acceptable_entry in logs_list.items():
         output_dir = None
         if FLAGS.output_directory:
             parsed_url = urlparse.urlsplit(log_url)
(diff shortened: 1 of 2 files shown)


pphaneuf:
> There's also a small bugfix in case the chain submitted consisted of a single certificate (so the extra_data is empty). That happens when a root was submitted to the log.

Could have split that out! :wink:

python/utilities/ev_whitelist/create_ev_whitelist_zip.py
```diff
@@ -21,13 +21,13 @@
...
 NUM_FETCHING_PROCESSES = 40
-LOGS_LIST = [
-    "https://ct.googleapis.com/pilot",
-    "https://ct.googleapis.com/aviator",
-    "https://ct1.digicert-ct.com/log",
-    "https://log.certly.io",
-    "https://ct.izenpe.com",
-]
+LOGS_LIST = {
+    "https://ct.googleapis.com/pilot":6082932,
```
phad: how did you come up with these 'last_acceptable_entry' index values?
when do they get refreshed?

phad: they really do fit the criteria of 'magic number' !  I think you should add some docs to explain where they come from and how they might change.

python/utilities/ev_whitelist/generate_ev_whitelist.py
```diff
@@ -66,7 +68,8 @@ def _write_cert_and_chain(
              "wb"))
 
 def _ev_match(
-        output_dir, certificate, entry_type, extra_data, certificate_index):
+        output_dir, last_entry_index, certificate, entry_type, extra_data,
```
phad: you might use the word 'acceptable' in last_entry_index here too, to be consistent with above, and to make the code better self-documented.

python/utilities/ev_whitelist/create_ev_whitelist_zip.py
eranmes: (1) I manually found them from looking at logs of my January 1st run.
(2) They would never get refreshed - the whitelist was generated once to include all EV certs logged prior to January 1st. The goal of regenerating it is reducing its size (hence memory footprint on Chrome clients).

phad: you might consider having these values passed in via the command line or extracted from a configuration file.

python/utilities/ev_whitelist/generate_ev_whitelist.py
```diff
@@ -66,7 +68,8 @@ def _write_cert_and_chain(
...
+        certificate_index):
     """Matcher function for the scanner. Returns the certificate's hash if
     it is a valid EV certificate, None otherwise."""
```
phad: Comment needs revising as it's now possible for a cert to be a valid EV cert yet not get returned because of its index being deemed 'not acceptable'

python/utilities/ev_whitelist/create_ev_whitelist_zip.py
phad: I see, so they are constants that will never change.
You should document this better, as it's not clear to the reader who doesn't know the history of the EV whitelist tool.

phad: re: reducing the whitelist size.

AFAICT this can only happen by 'natural wastage' ie. no longer including certs in the whitelist that once were. eg. if they expire.

eranmes: All good points, I have added documentation that will hopefully clarify this.

python/utilities/ev_whitelist/generate_ev_whitelist.py
eranmes: Done.

python/utilities/ev_whitelist/generate_ev_whitelist.py
eranmes: Done.

eranmes:
@phad PTAL.

PR #777, opened 2015-06-12
Use DB Iterator when scanning entries.

AlCutter:


cpp/log/tree_signer-inl.h
```diff
@@ -237,14 +237,12 @@ typename TreeSigner<Logged>::UpdateResult TreeSigner<Logged>::UpdateTree() {
   uint64_t min_timestamp = LastUpdateTime() + 1;
 
   // Sequence any new sequenced entries from our local DB.
+  auto it(db_->ScanEntries(cert_tree_->LeafCount()));
   for (int64_t i(cert_tree_->LeafCount());; ++i) {
     Logged logged;
-    typename Database<Logged>::LookupResult result(
-        db_->LookupByIndex(i, &logged));
-    if (result == Database<Logged>::NOT_FOUND) {
+    if (!it->GetNextEntry(&logged)) {
```
pphaneuf: Ah ha! This is insufficient... `GetNextEntry()` will, um, give you the next entry in the database, not the next entry in index order, so you might hit gaps. You have to check that you're contiguous here.

cpp/log/log_lookup-inl.h
```diff
@@ -68,8 +69,7 @@ void LogLookup<Logged>::UpdateFromSTH(const ct::SignedTreeHead& sth) {
     // handled more gracefully. E.g. we could retry a failed update
     // a number of times -- but until we know under which conditions
     // the database might fail (database busy?), just die.
-    CHECK_EQ(Database<Logged>::LOOKUP_OK,
-             db_->LookupByIndex(sequence_number, &logged))
+    CHECK(it->GetNextEntry(&logged))
```
pphaneuf: This used to get the entry by index, using `sequence_number`, so the `CHECK_EQ` at line 77 was just a sanity check ("I asked for 10 and got 11?!?")... But now, you might have a gap, and get some other entry index...

That said, it was a `LookupByIndex()` wrapped in a `CHECK_EQ`, so it was already going to crash (I guess that it's a precondition for this method that the database has the entries we need for the STH we get passed in?), so it might be okay (or, at least, no worse)?

cpp/server/handler.cc
```diff
@@ -494,11 +494,11 @@ void HttpHandler::AddPreChain(evhttp_request* req) {
 void HttpHandler::BlockingGetEntries(evhttp_request* req, int64_t start,
                                      int64_t end, bool include_scts) const {
   JsonArray json_entries;
+  auto it(db_->ScanEntries(start));
   for (int64_t i = start; i <= end; ++i) {
     LoggedCertificate cert;
 
-    if (db_->LookupByIndex(i, &cert) !=
-        ReadOnlyDatabase<LoggedCertificate>::LOOKUP_OK) {
+    if (!it->GetNextEntry(&cert)) {
```
pphaneuf: Should be like this:

```C++
if (!it->GetNextEntry(&cert) || cert->sequence_number() != i) {
```

pphaneuf:
Review done. Curious to see if this indeed improve the startup time (and it might improve the replication speed, but I'm not sure this was a bottleneck).

pphaneuf:
I'll take another look once you've rebased on top of (at least) #781.

cpp/log/tree_signer-inl.h
AlCutter: Yeah; currently it'll `CHECK` if it hits a gap, but that's not entirely ideal given the potentially sparse nature of the data above the STH.

cpp/log/log_lookup-inl.h
AlCutter: If we're committed to updating to that passed in STH and we don't have the data for it then we're screwed (since calls to this method happen via the new-sth-callback from the DB when we've decided to write out a new STH locally - so the data must be *somewhere*)

cpp/server/handler.cc
AlCutter: Yah, forgot that here too.

cpp/log/log_lookup-inl.h
pphaneuf: Yeah, I agree, if we have an STH in the database, but not the entries for it, something is seriously wrong, so blowing up is okay.

pphaneuf:
LGTM, with extra bonus points if you can get an idea of how much this helps? :smiley:

AlCutter:
Quite unscientifically, it seems to be about 10% faster (~24m -> ~22m) on restart time over a mirror of Pilot's log.  But using the iterator might let people using their own DBs gain some speed up.

pphaneuf:
Your mirror was using leveldb, right? :smiley:

PR #778, opened 2015-06-12
Push potentially blocking bits of Fetch() off onto a thread pool.

AlCutter:


cpp/net/url_fetcher.cc
```diff
@@ -231,10 +242,14 @@ UrlFetcher::~UrlFetcher() {
...
-  State* const state(new State(&impl_->pool_, req, resp, task));
+  State* const state(new State(impl_->base_, &impl_->pool_, req, resp, task));
   task->DeleteWhenDone(state);
 
-  impl_->base_->Add(bind(&State::MakeRequest, state));
+  // Run State::MakeRequest() on the task's executor because it may
+  // block doing DNS resolution etc.
+  // TODO(alcutter): this can go back to being put straight on the event Base
+  // once evhtp supports creating SSL connections to a DNS name.
+  task->executor()->Add(bind(&State::MakeRequest, state));
```
pphaneuf: Unfortunately, as we all know, everything is bad. There's quite a number of callers that use the `libevent::Base` as the executor on their `util::Task`, so you're back in the same place. :disappointed:

pphaneuf:
Also, the Travis failure looks like it's related:

https://travis-ci.org/google/certificate-transparency/builds/66559727#L1832

PR #779, opened 2015-06-12
Use SSL_set_ex_data() and friends to associate a Connection with its SSL

AlCutter:


pphaneuf:
LGTM, maybe @benlaurie ought to take a look? :wink:

cpp/net/connection_pool.cc
```diff
@@ -113,29 +101,21 @@ int ConnectionPool::Connection::SSLVerifyCallback(const int preverify_ok,
...
-    {
-      lock_guard<mutex> lock(ssl_mutex_);
-      CHECK(connections_by_ssl_.insert(make_pair(ssl_, this)).second);
-    }
-    SSL_set_tlsext_host_name(ssl_, other_end_.first.c_str());
-  }
+// static
+int ConnectionPool::Connection::GetSSLConnectionIndex() {
+  static const int ssl_connection_index(
+      SSL_get_ex_new_index(0, nullptr, nullptr, nullptr, nullptr));
```
benlaurie: Mildly concerned that this gets called before SSL is initialized.

AlCutter: Fairly sure this gets initialised the first time the that `GetSSLConnectionIndex()` is called, so it should be well after SSL is initialised (it's also threadsafe since C++0x.)

AlCutter: http://en.cppreference.com/w/cpp/language/storage_duration (see "Static Local Variables" section)

benlaurie: Hmm, that's not the clearest reference ever, but I believe you are correct.

PR #780, opened 2015-06-12
Add support for Google Cloud Monitoring

AlCutter:


pphaneuf:
Looks like a potential build error, but if I recall correctly, we said that this PR would need some refactoring, right? So hold off on review until it's sorted?

Dockerfile-ct-mirror
```diff
@@ -47,6 +49,9 @@ CMD cd /mnt/ctmirror/ && \
         --num_http_server_threads=${NUM_HTTP_SERVER_THREADS} \
         --target_public_key=/usr/local/etc/keys/${TARGET_LOG_PUBLIC_KEY} \
         --target_log_uri=${TARGET_LOG_URL} \
+        --monitoring=${MONITORING} \
+        --google_compute_monitoring_push_url="https://www.googleapis.com/cloudmonitoring/v2beta2/projects/${PROJECT}/timeseries:write" \
+        --google_compute_monitoring_metrics_url="https://www.googleapis.com/cloudmonitoring/v2beta2/projects/${PROJECT}/metricDescriptors" \
```
phad: maybe pull out the common parts of these two URLs to some other var

Makefile.am
```diff
@@ -138,6 +139,8 @@ cpp_libcore_a_SOURCES = \
 	cpp/merkletree/serial_hasher.cc \
 	cpp/merkletree/tree_hasher.cc \
 	cpp/monitoring/prometheus/exporter.cc \
+	cpp/monitoring/gcm/exporter.cc \
```
phad: gcm < prometheus

cloud/google/config.sh
```diff
@@ -28,26 +31,31 @@ for i in $(seq ${LOG_NUM_REPLICAS}); do
...
-export PROMETHEUS_DISK_SIZE=50GB
-export PROMETHEUS_BASE_NAME="${CLUSTER}-prometheus"
-export PROMETHEUS_MACHINE_TYPE=n1-standard-1
-declare -a PROMETHEUS_MACHINES LOG_DISKS
-export PROMETHEUS_MACHINES LOG_DISKS
-for i in $(seq ${PROMETHEUS_NUM_REPLICAS}); do
-  PROMETHEUS_MACHINES[$i]="${CLUSTER}-prometheus-$i"
-  PROMETHEUS_DISKS[$i]="${CLUSTER}-prometheus-disk-$i"
-done
+if [ "${MONITORING}" == "prometheus" ]; then
```
phad: isn't there any environment you need to set for "gcm" ?

cloud/google/mirror-config.sh
```diff
@@ -31,16 +34,18 @@ for i in $(seq ${MIRROR_NUM_REPLICAS}); do
...
-export PROMETHEUS_DISK_SIZE=50GB
-export PROMETHEUS_BASE_NAME="${CLUSTER}-prometheus"
-export PROMETHEUS_MACHINE_TYPE=n1-standard-1
-declare -a PROMETHEUS_MACHINES MIRROR_DISKS
-export PROMETHEUS_MACHINES MIRROR_DISKS
-for i in $(seq ${PROMETHEUS_NUM_REPLICAS}); do
-  PROMETHEUS_MACHINES[$i]="${CLUSTER}-prometheus-$i"
-  PROMETHEUS_DISKS[$i]="${CLUSTER}-prometheus-disk-$i"
-done
+if [ "${MONITORING}" == "prometheus" ]; then
```
phad: there seems to be a lot of duplication between config.sh and mirror-config.sh
can you extract common stuff to a separate .sh script?

phad: (and in general, between foo and mirror-foo)

cloud/google/start_log.sh
```diff
@@ -37,6 +39,7 @@ for i in `seq ${LOG_NUM_REPLICAS}`; do
       --image container-vm \
       --disk name=${LOG_DISKS[${i}]},mode=rw,boot=no,auto-delete=yes \
       --tags log-node \
+      --scopes "monitoring,storage-ro,compute-ro,logging-write" \
```
phad: what's this for?

cpp/monitoring/gauge.h
```diff
@@ -25,12 +25,13 @@ class Gauge : public Metric {
 
   void Set(const LabelTypes&... labels, double value);
 
+  // TODO(alcutter): Not over the moon about having this here.
```
phad: Nice TODO.  Do you have any idea what might be done about it?  The TODO would be more informative if you can write what steps are needed to remove it later.

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+#include "monitoring/registry.h"
+#include "net/url.h"
+#include "util/json_wrapper.h"
+#include "util/sync_task.h"
+
+DEFINE_string(google_compute_monitoring_push_url, "",
+              "GCM custom timeseries write URL.");
+DEFINE_string(google_compute_monitoring_metrics_url, "",
+              "GCM custom timeseries metrics URL.");
+DEFINE_int32(google_compute_monitoring_push_interval_seconds, 5,
```
phad: The 5s interval seems very short!  Is that a recommendation from GCM?  I assume this consumes some kind of quota faster than say a 30s interval would do.

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+              "http://metadata/computeMetadata/v1/instance/service-accounts",
+              "URL of GCE metadata server.");
+DEFINE_string(google_compute_monitoring_service_account, "default",
+              "Which GCE service account to use for pushing metrics.");
+
+namespace cert_trans {
+
+using std::chrono::seconds;
+using std::chrono::system_clock;
+using std::chrono::minutes;
```
phad: minutes < seconds

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+
+
+// Label and Metric -name prefix.
+const char kCloudPrefix[] = "custom.cloudmonitoring.googleapis.com/ct/";
+
+
+inline std::string RFC3339Time(const system_clock::time_point& when) {
+  const std::time_t now_c(system_clock::to_time_t(when));
+  char buf[256];
+  CHECK(std::strftime(buf, sizeof(buf), "%FT%T.00Z", std::localtime(&now_c)));
```
phad: Github is highlighting the % of %T in red for me.  Don't know why, but maybe you should check.

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+namespace {
+
+
+// Label and Metric -name prefix.
+const char kCloudPrefix[] = "custom.cloudmonitoring.googleapis.com/ct/";
+
+
+inline std::string RFC3339Time(const system_clock::time_point& when) {
+  const std::time_t now_c(system_clock::to_time_t(when));
+  char buf[256];
```
phad: "256 should be big enough for anyone!"

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+  CHECK(std::strftime(buf, sizeof(buf), "%FT%T.00Z", std::localtime(&now_c)));
+  return buf;
+}
+
+
+}  // namespace
+
+
+GCMExporter::GCMExporter(const string& instance_name, UrlFetcher* fetcher,
```
phad: Can you make UrlFetcher* const.  You're not taking ownership of it.
It'd be 'UrlFetcher const* fetcher' right?

phad: The corresponding field in the .h is const.  So you should be able to here as well.

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+                         Executor* executor)
+    : instance_name_(instance_name),
+      fetcher_(CHECK_NOTNULL(fetcher)),
+      executor_(CHECK_NOTNULL(executor)),
+      task_(executor_) {
+  RefreshCredentials();
+  CreateMetrics();
+  PushMetrics(nullptr);
```
phad: Can the arg be something better self-documented than 'nullptr'?  I have to look at the definition to find out what that is going to do.  A const, or an overload of PushMetrics with no args and a more descriptive name maybe.

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+      (URL(FLAGS_google_compute_metadata_url + "/" +
+           FLAGS_google_compute_monitoring_service_account + "/token")));
+  req.headers.insert(make_pair("Metadata-Flavor", "Google"));
+
+  UrlFetcher::Response resp;
+  SyncTask task(executor_);
+  fetcher_->Fetch(req, &resp, task.task());
+  task.Wait();
+
+  // TODO(alcutter): Handle this better
```
phad: Yeah you're going to have to return a util::Status or something.  As it stands this function claims it cannot fail!
Is it worth making the util::Status return now and just have it return Status::OK until you do this TODO?

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+  VLOG(1) << "GCM credentials refreshed";
+}
+
+
+namespace {
+
+
+void AddLabelDescription(const string& key, const string& desc,
+                         JsonArray* labels) {
+  JsonObject l;
```
phad: 'l'  you're not coding in Go here!

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+
+    // See
+    // https://cloud.google.com/monitoring/v2beta2/metricDescriptors#resource
+    // for a description of the structure we're building here.
+
+    JsonArray labels;
+    AddLabelDescription("instance",
+                        "Instance from which the sample originates.", &labels);
+    for (const auto& label : m->LabelNames()) {
+      AddLabelDescription(label, label, &labels);
```
phad: label, label means you have some fairly low-quality descriptions in there.

phad: doesn't the Metric support having a description field?

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+using std::unique_lock;
+using util::Executor;
+using util::SyncTask;
+
+
+namespace {
+
+
+// Label and Metric -name prefix.
+const char kCloudPrefix[] = "custom.cloudmonitoring.googleapis.com/ct/";
```
phad: Is this a landgrab for /ct, or is it scoped to within our monitoring app?  If it's scoped to within our app, why do you need the custom.blah.blah.com/ prefix?

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+    req.verb = UrlFetcher::Verb::POST;
+    req.headers.insert(make_pair("Content-Type", "application/json"));
+    req.headers.insert(make_pair("Authorization", "Bearer " + bearer_token_));
+    req.body = metric.ToString();
+
+    UrlFetcher::Response resp;
+    SyncTask task(executor_);
+    VLOG(1) << "Creating metric m.Name()...";
+    VLOG(2) << req.body;
+    fetcher_->Fetch(req, &resp, task.task());
```
phad: "Fetch" feels weird when you're doing a POST ...

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+    task.Wait();
+    CHECK_EQ(util::Status::OK, task.status());
```
phad: What if it's not OK?

cpp/monitoring/gcm/exporter.cc
```diff
@@ -0,0 +1,264 @@
...
+
+
+void GCMExporter::PushMetrics(util::Task* child_task) {
+  if (task_.task()->CancelRequested()) {
+    task_.task()->Return(util::Status::CANCELLED);
+    return;
+  }
+
+  if (system_clock::now() - token_refreshed_at_ > minutes(3)) {
+    RefreshCredentials();
```
phad: Wow, the credentials expire this frequently?

cpp/monitoring/gcm/exporter_test.cc
```diff
@@ -0,0 +1,184 @@
...
+#include <memory>
+
+#include "monitoring/monitoring.h"
+#include "net/mock_url_fetcher.h"
+#include "util/json_wrapper.h"
+#include "util/testing.h"
+#include "util/thread_pool.h"
+
+DECLARE_string(google_compute_monitoring_push_url);
+DECLARE_string(google_compute_monitoring_metrics_url);
```
phad: metrics < push

cpp/monitoring/gcm/exporter_test.cc
```diff
@@ -0,0 +1,184 @@
...
+DECLARE_int32(google_compute_monitoring_push_interval_seconds);
```
phad: push_interval < push_url

cpp/monitoring/gcm/exporter.cc
pphaneuf: The field in the header is `UrlFetcher* const` (a const pointer to a non-const object). `UrlFetcher const*` could be a non-const pointer to a const object (the same as `const UrlFetcher*`), which is something else (and wouldn't assign).

Making the flag here `UrlFetcher* const` would work, but isn't really useful, as the pointer itself is, uh, "passed by value", if you will (you can't change the pointer in the caller, so it makes no additional promises to the caller).

cpp/monitoring/gcm/exporter.cc
pphaneuf: If you POST 10 bytes and receive 10 gigs, would you feel better? :wink:

cpp/monitoring/gcm/exporter.cc
phad: My point is that the convention of passing (Foo* foo) generally denotes an ownership transfer, which doesn't occur in this case.

cpp/monitoring/gcm/exporter.cc
pphaneuf: It's just 4 lines, so I'm not overly worried, but I usually try to make sure it's something that can be searched for in a relatively useful way. :smiley:

cpp/monitoring/gcm/exporter.cc
pphaneuf: In this code base, we've actually used `unique_ptr<Foo>&& foo` to denote ownership transfer when we could (the new world of C++11!), and passing down a raw pointer (`Foo* foo`) implies no ownership transfer by default (we do specify when ownership is taken).

Dockerfile-ct-mirror
AlCutter: Good call, done.

Makefile.am
AlCutter: done.

cloud/google/config.sh
AlCutter: Not really - we just don't run stuff. Arguably you might want to configure the [new] `..._base_url` flag, but I'd say we'll add that when/if we need it.

cloud/google/mirror-config.sh
AlCutter: Yeah, there's definitely some things which can be pulled out, but I'll leave that for a follow-up if that's ok.

cloud/google/start_log.sh
AlCutter: Just gets the instance permission to call the listed Cloud APIs.
Needed to add `monitoring`, but then I need to explicitly list all the others too.
I suppose we may not need `monitoring` if you chose `prometheus`, but it won't hurt.

cpp/monitoring/gauge.h
AlCutter: Yeah, I don't really have a plan at the moment.
It won't actually become a problem until the day that we try to add things like Histogram support, which could quite possibly be Friday the 15th of Never.

cpp/monitoring/gcm/exporter.cc
AlCutter: 15s, that's my *final* offer. ;)

cpp/monitoring/gcm/exporter.cc
AlCutter: done

cpp/monitoring/gcm/exporter.cc
AlCutter: For me too, but `%T` is in the standard: http://www.cplusplus.com/reference/ctime/strftime/
Just another github diff quirk?

cloud/google/mirror-config.sh
phad: Sure.  Maybe TODO it then.

cpp/monitoring/gcm/exporter.cc
phad: Done.

cpp/monitoring/gcm/exporter.cc
phad: Quite possibly.  I can well believe that, seeing the kinda inconsistent approach to diff highlighting in general.

cpp/monitoring/gcm/exporter.cc
AlCutter: We use this same construction elsewhere too, but I've added a small comment anyway.

cpp/monitoring/gcm/exporter.cc
AlCutter: It can't fail - it'll destroy the universe before failing :)
I'll leave it for the `TODO` followup PR.

cpp/monitoring/gcm/exporter.cc
AlCutter: There you go.

cpp/monitoring/gcm/exporter.cc
AlCutter: The `Metric` has a description (see L152), but there's no description of the `labels` other than its name (which already should really provide some insight into what it's there for.)

cpp/monitoring/gcm/exporter.cc
AlCutter: Yeah, this is annoying.
You *need* to provide this scope, and it *must* start with `custom.cloudmonitoring.googleapis.com/`, then you need an identifier, for which I originally wanted to use something like `certificate-transparency.org`, but insanely, there's also a fairly tight maximum length allowed, and we blew past it so I had to use this.

cpp/monitoring/gcm/exporter.cc
AlCutter: Could be handled better in a follow up PR, but if you failed to create the metrics then you're not going to be able to write to them, and your monitoring will have no idea what's going on (despite all the jobs apparently being "up"), seemed better to just bomb out and at least cause all the bells and lights to go off.

cpp/monitoring/gcm/exporter.cc
AlCutter: I've added a flag and dropped the default to 30m (I think they expire after 60 by default.)

cpp/monitoring/gcm/exporter_test.cc
AlCutter: Done.

cloud/google/mirror-config.sh
AlCutter: Added.

cpp/monitoring/gcm/exporter_test.cc
AlCutter: obsoleted.

AlCutter:
PTAL, petal.

cpp/monitoring/gcm/exporter.cc
phad: OK

cpp/monitoring/gcm/exporter.cc
phad: OK

cloud/google/config.sh
phad: OK

cloud/google/start_log.sh
phad: OK

cpp/monitoring/gauge.h
phad: LOL, OK

PR #783, opened 2015-06-16
Refactor monitoring a bit in preparation for Google Cloud Monitoring implementation.

AlCutter:


cpp/monitoring/counter.h
```diff
@@ -1,42 +1,92 @@
...
 #include "base/macros.h"
 #include "monitoring/metric.h"
+#include "monitoring/labelled_values.h"
 
 namespace cert_trans {
 
-
-// A metric which can only increase (e.g. total_requests_served).
-template <class... Labels>
+template <class... LabelTypes>
```
phad: Did you for some reason decide that doc comments weren't necessary any more?

cpp/monitoring/gauge.h
```diff
@@ -18,22 +21,63 @@ class Gauge : public Metric {
       const typename NameType<LabelTypes>::name&... label_names,
       const std::string& help);
 
-  virtual double Get(const LabelTypes&...) const = 0;
```
phad: Was Gauge never subclassed in the end?

cpp/monitoring/counter.h
AlCutter: No, but I did replace the contents of the file.
I *thought* I'd added that comment back, maybe I only added it to `gauge.h`.

phad: Hmm.  Comments please.

cpp/monitoring/gauge.h
AlCutter: It used to be that you'd subclass `Counter`/`Gauge` for your new concrete (e.g. Prometheus) implementation, but I've canned that now and there's just a single concrete `Counter` and `Gauge` impl and the fancy pants magic Prometheus/... bits are in their own exporters.

cpp/monitoring/registry.cc
```diff
@@ -11,43 +15,20 @@ Registry* Registry::Instance() {
...
 
-void Registry::Export(std::ostream* os) const {
-  std::lock_guard<std::mutex> lock(mutex_);
-  for (const auto* m : metrics_) {
-    m->Export(os);
-  }
-}
-
-
-void Registry::ExportHTML(std::ostream* os) const {
```
phad: Was this not being used anywhere?

cpp/monitoring/registry.cc
```diff
@@ -11,43 +15,20 @@ Registry* Registry::Instance() {
...
 
 
 void Registry::ResetForTestingOnly() {
-  std::lock_guard<std::mutex> lock(mutex_);
+  lock_guard<mutex> lock(mutex_);
   metrics_.clear();
 }
 
 
-void Registry::Export(std::ostream* os) const {
```
phad: Was this not being used anywhere?

cpp/monitoring/prometheus/counter-inl.h
```diff
@@ -1,74 +0,0 @@
...
-
-#include <glog/logging.h>
-
-#include "util/protobuf_util.h"
-
-namespace cert_trans {
-
-// static
-template <class... LabelTypes>
-PrometheusCounter<LabelTypes...>* PrometheusCounter<LabelTypes...>::New(
```
phad: Where have all these inline definitions been moved to?

cpp/server/proxy.cc
```diff
@@ -5,6 +5,7 @@
 #include <event2/http.h>
 #include <event2/http_compat.h>
 #include <event2/keyvalq_struct.h>
+#include <glog/logging.h>
```
phad: Is this an IWYU?  I assume it must have come in transitively from elsewhere.

cpp/monitoring/registry.cc
AlCutter: Previously the concrete `Metric` implementations knew how to render themselves to an `ostream` and this would just get all known metrics to do that, but now the exporter does that job.

cpp/monitoring/registry.cc
AlCutter: ditto

cpp/monitoring/prometheus/counter-inl.h
AlCutter: They've gone away now because there aren't `Counter`/`Gauge` subclasses any longer.

cpp/server/proxy.cc
AlCutter: Yeah.

PR #786, opened 2015-06-17
Python: Small change to auto-generated code

eranmes:
Needed so that the Chromium class handling URLs (GURL) will correctly
append the command to be called to the log URL.


eranmes:
Manually merged as https://github.com/google/certificate-transparency/commit/ae5d23723bf02d51e5ccb3d6345b144de5e18bcd

PR #789, opened 2015-06-17
Run across multiple GCE zones

AlCutter:


cloud/google/create_new_cluster.sh
```diff
@@ -49,7 +62,7 @@ echo "============================================================="
 
 # Set gcloud defaults:
 ${GCLOUD} config set project ${PROJECT}
-${GCLOUD} config set compute/zone ${ZONE}
+${GCLOUD} config unset compute/zone
```
phad: you did this just above in the workaround

cloud/google/util.sh
```diff
@@ -10,18 +10,21 @@ function Header() {
...
   TYPE=$1
   NAME=$2
-  WANT_STATUS=$3
+  ZONE=$3
+  WANT_STATUS=$4
   STATUS=""
   until [[ "${STATUS}" =~ "${WANT_STATUS}" ]]; do
     sleep 1
-    STATUS=$(${GCLOUD} compute ${TYPE} describe ${NAME} | grep "status:")
+    STATUS=$(${GCLOUD} compute ${TYPE} describe ${NAME} --zone ${ZONE} | grep "status:")
```
phad: You should perhaps specify the grep you want, eg.  /usr/bin/grep

cloud/google/create_new_cluster.sh
AlCutter: Yeah, I just wanted to make sure there wasn't one set by default coming in here.

cloud/google/util.sh
AlCutter: I think all greps should do what I'm after here (or it has no business calling itself grep!)
Plus if I wanted the script to use my own 1337 grep I'd probably try shoving it in the `PATH` first.

PR #792, opened 2015-06-19
Factor out some common config script

AlCutter:
And add stock mirror configs for several logs.


cloud/google/configs/ct-mirror-certly1.sh
```diff
@@ -0,0 +1,9 @@
+export INSTANCE_TYPE="mirror"
+export PROJECT="ct-mirror-certly1"
```
phad: when did this log start being called 'certly1' ?  Haven't seen the number suffix before.

phad: oh looks like you're doing it on the basis of the _1.pem file, and you do the same below.
What's the story there?

cloud/google/configs/ct-mirror-izenpe1.sh
```diff
@@ -0,0 +1,9 @@
+export INSTANCE_TYPE="mirror"
+export PROJECT="ct-mirror-izenpe1"
+export REGION="us-central1"
+export CLUSTER="ct-mirror-izenpe1"
+export ZONES="a c f"
```
phad: what's your strategy for choosing zones here?  unique triples taken from a, b, c, f apparently?

cloud/google/configs/ct-mirror-certly1.sh
AlCutter: Just in case there's a second certly (etc.) log at some point in the future.

cloud/google/configs/ct-mirror-izenpe1.sh
AlCutter: yeah, that's pretty much it - some random subset of the available zones [which are currently abcf - as you noticed :)]

phad:
Why do you have config for some but not all the accepted Logs?

And, should these config files even be in this repo?  Seems odd mixing config relating to particular instances of logs into a repo containing code for building logs.

AlCutter:
Which other accepted log did I miss?

Thought it would be helpful to provide configs in case others want to run mirrors too.

cloud/google/configs/ct-mirror-certly1.sh
phad: Got it.

phad:
My bad, I failed to notice Pilot & Rocketeer.
So you have them covered.
Still not convinced these files belong here though.  An example file would be enough.

PR #793, opened 2015-06-19
Don't proxy on the libevent thread.

AlCutter:


phad:
Don't suppose you have any test that exhibits deadlocks or such without this fix?

pphaneuf:
LGTM

PR #796, opened 2015-06-23
Fix up initialization of ct-clustertool.

pphaneuf:


AlCutter:
Can you make it exit by itself too? Looks like it just hangs around forever now when you do an `initialize_log`.

pphaneuf:
One thing per PR, Al, one thing per PR. :wink:

PR #800, opened 2015-06-24
Report etcd errors more accurately.

AlCutter:


cpp/util/etcd.cc
```diff
@@ -97,12 +83,73 @@ util::error::Code ErrorCodeForHttpResponseCode(int response_code) {
...
+Status StatusFromResponse(const int response_code,
+                          const shared_ptr<JsonObject>& json) {
+  // Prefer the etcd errorCode if there is one:
+  if (json->Ok()) {
+    const JsonInt error_code(*json, "errorCode");
+    if (error_code.Ok()) {
+      const util::error::Code status_code(
+          StatusCodeFromEtcdErrorCode(error_code.Value()));
+      const string msg(EtcdErrorMessage(json));
+      return Status(status_code, msg);
```
pphaneuf: Could skip the temporaries without losing much readability there.

cpp/util/etcd.cc
```diff
@@ -97,12 +83,73 @@ util::error::Code ErrorCodeForHttpResponseCode(int response_code) {
...
+    ret = json->DebugString();
+  }
+  return ret;
+}
+
+
+Status StatusFromResponse(const int response_code,
+                          const shared_ptr<JsonObject>& json) {
+  // Prefer the etcd errorCode if there is one:
+  if (json->Ok()) {
```
pphaneuf: This would bomb if `json` is null, maybe `StatusFromResponse` should take a `const JsonObject&`?

cpp/util/etcd.cc
AlCutter: Skipped

cpp/util/etcd.cc
AlCutter: Ok, I don't think that can happen unless we're in OOM, but done anyway.

cpp/util/etcd.cc
```diff
@@ -97,12 +83,70 @@ util::error::Code ErrorCodeForHttpResponseCode(int response_code) {
...
+      ret += ", cause: " + string(cause.Value());
+    }
+  } else {
+    ret = json.DebugString();
+  }
+  return ret;
+}
+
+
+Status StatusFromResponse(const int response_code, const JsonObject& json) {
```
phad: const int here but not in the function above.  which is it to be?

phad:
Looks fine, will you squash?

cpp/util/etcd.cc
AlCutter: Doesn't really matter since it's by value, I'll just remove this one.

AlCutter:
Not very much; I've got bones.

AlCutter:
I mean yes.

PR #801, opened 2015-06-25
Push ct process logs to GCM Logging

AlCutter:


cloud/google/node_init.sh
```diff
@@ -6,4 +6,21 @@ sudo mkdir ${DATA_DIR}
 sudo /usr/share/google/safe_format_and_mount \
   -m "mkfs.ext4 -F" ${DATA_DEV} ${DATA_DIR}
 
-
+# Install google-fluentd which pushes application log files up into the Google
+# Cloud Logs Monitor.
+curl -sSO https://storage.googleapis.com/signals-agents/logging/google-fluentd-install.sh
```
phad: -sS0 ?

cloud/google/node_init.sh
```diff
@@ -6,4 +6,21 @@ sudo mkdir ${DATA_DIR}
...
+sha256sum ./google-fluentd-install.sh
```
phad: I assume this outputs the SHA256 of the script you just downloaded.
You don't do anything with the output and go on to do sudo things.
Seems risky.

phad: Isn't it better to just instruct the log operator to install google-fluentd themselves and assert here that it's present on their system?

cloud/google/node_init.sh
```diff
@@ -6,4 +6,21 @@ sudo mkdir ${DATA_DIR}
...
+<source>
+  type tail
+  format none
+  path /data/ctlog/logs/ct-server.*.INFO.*
+  pos_file /data/ctlog/logs/ct-server.INFO.pos
+  read_from_head true
+  tag ct
+</source>
+EOF
+sudo cp /tmp/ct-info.conf /etc/google-fluentd/config.d/ct-info.conf
```
phad: And leave all these sudo steps to the operator to perform manually.

phad:
This looks scary.  I think you shouldn't have anything 'sudo'd in such a script.  Leave it to documentation of pre-requisites for the Log operator to do themselves.

cloud/google/node_init.sh
AlCutter: This fetch & install stuff just comes verbatim from the docs:
https://cloud.google.com/logging/docs/agent/installation

cloud/google/node_init.sh
AlCutter: The bit in the docs where it talking about doing this via the start-up script (as I've done) don't verify the sha256, I guess because you'd have to keep updating the script as they pushed new versions.  so I took the approach of logging the sha256 so at least you could look back and see.

I don't really mind either way.

AlCutter: Having the operator do this by hand for all node instances would be a ball ache.

AlCutter:
The script already sudoes to format and mount the persistent disk...

cloud/google/node_init.sh
phad: Can you still that URL in the comments then.

cloud/google/node_init.sh
phad: Well, yes I can see that, but I think having sudo in a script is a bad idea.

cloud/google/node_init.sh
phad: OK I withdraw this.

How about you remove all sudo from this script and require the operator to sudo bash -c node_init.sh instead.
Then they are the ones typing sudo and not having some script do it for them.

cloud/google/node_init.sh
AlCutter: I'll add the sha checking.

cloud/google/node_init.sh
AlCutter: Because the whole point of a start-up script is that it runs at start-up automatically :)
This is a GCE/container-vm thing.

phad: You mean it's meant to hook into some init.d style start-up? or a newer 'service foo start' kind of thing?  What / who is going to run it?

AlCutter: Yes, something in the container-vm image runs this based on the "startup-script" metadata entry (which has this script's name as its value.)

phad:
OK if you're sure this is all a good idea.

PR #802, opened 2015-06-25
Populate TBS and Cert Signature fields in X509Description.

phad:
These fields were already defined but not being populated.
Add 'basic_constraint_ca' field to the X509Description proto as well.

### DIFF
--- a/python/ct/client/db/cert_desc.py
+++ b/python/ct/client/db/cert_desc.py
@@ -2,7 +2,9 @@ import re
 import hashlib
 import time
 from ct.crypto import cert
+from ct.crypto.asn1 import x509_common
 from ct.proto import certificate_pb2
+
 
 def from_cert(certificate, observations=[]):
     """Pulls out interesting fields from certificate, so format of data will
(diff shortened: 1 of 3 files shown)


python/ct/client/db/cert_desc.py
```diff
@@ -48,6 +50,27 @@ def from_cert(certificate, observations=[]):
...
+        proto.cert_signature.algorithm_id = cert_alg.long_name
+
+        cert_params = certificate.signature_algorithm()["parameters"]
+        proto.cert_signature.parameters = str(cert_params)
+    except cert.CertificateError:
+        pass
+
+    try:
+        is_ca = certificate.basic_constraint_ca()
+        proto.basic_constraint_ca = is_ca is not None and is_ca.__bool__()
```
eranmes: the 'is not None' bit is superfluous. Also, if type(is_ca) is not bool, you'll get an AttributeError here.
Wouldn't you get the same result from this expression by using bool(is_ca) ?

phad: basic_constraint_ca() can return None, ASN1.Bool (true) or ASN1.Bool(false)
I'm trying to handle each of those.

python/ct/client/db/cert_desc.py
```diff
@@ -48,6 +50,27 @@ def from_cert(certificate, observations=[]):
...
+        proto.cert_signature.algorithm_id = cert_alg.long_name
+
+        cert_params = certificate.signature_algorithm()["parameters"]
+        proto.cert_signature.parameters = str(cert_params)
+    except cert.CertificateError:
+        pass
+
+    try:
+        is_ca = certificate.basic_constraint_ca()
+        proto.basic_constraint_ca = bool(is_ca)
```
eranmes: Nit: inline is_ca?

python/ct/client/db/cert_desc_test.py
```diff
@@ -8,57 +8,71 @@
...
     def test_from_cert(self):
-        observations = []
-        for check in all_checks.ALL_CHECKS:
-            observations += check.check(CERT) or []
-        observations.append(observation.Observation(
-            "AE", u'ćę©ß→æ→ćąßę-ß©ąńśþa©ęńć←', (u'əę”ąłęµ', u'…łą↓ð→↓ś→ę')))
-        proto = cert_desc.from_cert(CERT, observations)
-        self.assertEqual(proto.der, CERT.to_der())
+        for test_case in [(CERT, False), (CA_CERT, True)]:
+            source = test_case[0]
```
eranmes: Nit: You can unpack both values together:
(source, expect_ca_true) = test_case

eranmes:
LGTM, see two minor comments.

python/ct/client/db/cert_desc_test.py
phad: Thanks, I forget about tuple assignment like this.

PR #805, opened 2015-06-26
Make ConnectionPool cache extra connections for a while.

AlCutter:
Try to handle connection churn (e.g. from etcd deletes) a bit better.

### DIFF
--- a/cpp/net/connection_pool.cc
+++ b/cpp/net/connection_pool.cc
@@ -1,8 +1,10 @@
 #include "net/connection_pool.h"
 
+#include <chrono>
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 
+#include "monitoring/monitoring.h"
 #include "util/openssl_util.h"
 
 extern "C" {
@@ -12,6 +14,9 @@ extern "C" {
 
 
 using std::bind;
+using std::chrono::duration_cast;
+using std::chrono::seconds;
+using std::chrono::system_clock;
 using std::lock_guard;
 using std::make_pair;
 using std::map;
@@ -21,6 +26,7 @@ using std::pair;
 using std::placeholders::_1;
 using std::placeholders::_2;
 using std::string;
+using std::to_string;
 using std::unique_lock;
 using std::unique_ptr;
 using util::ClearOpenSSLErrors;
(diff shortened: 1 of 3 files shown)


cpp/net/connection_pool.cc
```diff
@@ -240,9 +256,9 @@ unique_ptr<ConnectionPool::Connection> ConnectionPool::Get(const URL& url) {
   }
 
   VLOG(1) << "cached evhtp_connection for " << key.first << ":" << key.second;
-  unique_ptr<ConnectionPool::Connection> retval(move(it->second.back()));
+  unique_ptr<ConnectionPool::Connection> retval(
+      move(it->second.back().second));
```
phad: What if there's no second.back()

cpp/net/connection_pool.cc
```diff
@@ -279,14 +298,22 @@ void ConnectionPool::Put(unique_ptr<ConnectionPool::Connection>&& conn) {
 void ConnectionPool::Cleanup() {
   unique_lock<mutex> lock(lock_);
   cleanup_scheduled_ = false;
+  const system_clock::time_point cutoff(
+      system_clock::now() -
+      seconds(FLAGS_connection_pool_max_unused_age_seconds));
 
-  // std::map<HostPortPair, std::deque<unique_ptr<Connection>>> conns_;
+  // std::map<HostPortPair, std::deque<TimestampedConnection>> conns_;
```
phad: Remove this commented out stuff

cpp/net/connection_pool.cc
```diff
@@ -266,7 +282,10 @@ void ConnectionPool::Put(unique_ptr<ConnectionPool::Connection>&& conn) {
   auto& entry(conns_[key]);
 
   CHECK_GE(FLAGS_url_fetcher_max_conn_per_host_port, 0);
-  entry.emplace_back(move(conn));
+  entry.emplace_back(make_pair(system_clock::now(), move(conn)));
```
phad: If you were to inject a system_clock this would be more easily tested.

cpp/net/connection_pool.cc
AlCutter: There has to be at this point, see L244

cpp/net/connection_pool.cc
AlCutter: I think that was put there as a reference for what the `auto` in the for loop below would be.

cpp/net/connection_pool.cc
AlCutter: Ack.

cpp/net/connection_pool.cc
phad: Remove the conns_; bit then so it looks less like commented-out code.

AlCutter: oke doke

PR #806, opened 2015-06-26
Add read & write timeouts to connections.

AlCutter:


cpp/net/connection_pool.cc
```diff
@@ -32,16 +32,20 @@ using std::unique_ptr;
 using util::ClearOpenSSLErrors;
 using util::DumpOpenSSLErrorStack;
 
-DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
-             "maximum number of URL fetcher connections per host:port");
+DEFINE_int32(connection_read_timeout_seconds, 60,
+             "Connection read timeout in seconds, only applies while willing to read.");
```
phad: What does 'willing to' mean here?

cpp/net/connection_pool.cc
```diff
@@ -248,6 +252,9 @@ unique_ptr<ConnectionPool::Connection> ConnectionPool::Get(const URL& url) {
             ? base_->HttpsConnectionNew(key.first, key.second, ssl_ctx_.get())
             : base_->HttpConnectionNew(key.first, key.second),
         move(key)));
+    struct timeval read_timeout = { FLAGS_connection_read_timeout_seconds, 0 };
```
phad: are the zero values millisecs?

Maybe kZeroMillis?

cpp/net/connection_pool.cc
AlCutter: That's yoinked from the libevent docs (which is what actually implements it):
http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/bufferevent_8h.html#a3686d744e67e9a530ebf8ab888ade2cf

AlCutter: But in essence; if you're not actively trying to read or write, then it won't time out - so for example a connection idling with no requests on it wouldn't timeout.

cpp/net/connection_pool.cc
AlCutter: Good call - done.

PR #807, opened 2015-06-29
Watch fail returns ABORTED status (like it used to.)

AlCutter:


phad:
Where are all the tests this change must have broken?  :)

AlCutter:
Yeah, AFAIK all of the `Watch()` stuff is untested, right @pphaneuf ?

PR #808, opened 2015-06-29
Update scripts for GCE

AlCutter:
A simple script to update the log/mirror jobs.

### DIFF
--- a/cloud/google/update_log.sh
+++ b/cloud/google/update_log.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+if [ "$1" == "" ]; then
+  echo "Usage $0: <config-file>"
+  exit 1
+fi
+source ${DIR}/util.sh
+source ${DIR}/config.sh $1
+
+set -e
+GCLOUD="gcloud"
+
+Header "Updating log instances..."
+# TODO(alcutter): Wait for each task (or zone?) to be healthy before
+# proceeding.
+for i in `seq 0 $((${LOG_NUM_REPLICAS} - 1))`; do
+  echo "Updating ${LOG_MACHINES[${i}]}"
+  gcloud compute ssh ${LOG_MACHINES[${i}]} \
+      --zone ${LOG_ZONES[${i}]} \
+      --command \
+          'sudo docker pull gcr.io/'${PROJECT}'/super_duper:test &&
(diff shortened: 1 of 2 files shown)


cloud/google/update_log.sh
```diff
@@ -0,0 +1,25 @@
...
+
+Header "Updating log instances..."
+# TODO(alcutter): Wait for each task (or zone?) to be healthy before
+# proceeding.
+for i in `seq 0 $((${LOG_NUM_REPLICAS} - 1))`; do
+  echo "Updating ${LOG_MACHINES[${i}]}"
+  gcloud compute ssh ${LOG_MACHINES[${i}]} \
+      --zone ${LOG_ZONES[${i}]} \
+      --command \
+          'sudo docker pull gcr.io/'${PROJECT}'/super_duper:test &&
```
phad: What is gcr.io ?

cloud/google/update_mirror.sh
```diff
@@ -0,0 +1,25 @@
...
+
+Header "Updating mirror instances..."
+# TODO(alcutter): Wait for each task (or zone?) to be healthy before
+# proceeding.
+for i in `seq 0 $((${MIRROR_NUM_REPLICAS} - 1))`; do
+  echo "Updating ${MIRROR_MACHINES[${i}]}"
+  gcloud compute ssh ${MIRROR_MACHINES[${i}]} \
+      --zone ${MIRROR_ZONES[${i}]} \
+      --command \
+          'sudo docker pull gcr.io/'${PROJECT}'/super_mirror:test &&
```
phad: Was s/duper/mirror/g the only change between these two files?
I think you might find some common stuff to extract in that case :)

cloud/google/update_log.sh
AlCutter: It's Google's new container registry, docs at... https://gcr.io

cloud/google/update_mirror.sh
AlCutter: `/MIRROR_/LOG_/` too.
But the reasons I didn't factor it out are:
 - passing arrays is painful in bash, though I could just copy the contents of the `LOG_*` and `MIRROR_*` to e.g. `NODE_*` before calling, but
 -  when the `TODO`s are `TODONE`, there could be different checks necessary depending on you're updating (especially if that could also be an `etcd` instance.), again I guess you could define a function and then pass in the command to call it which gets `eval`'d  or something, but you have to question whether it's worth it to dedupe this pretty small loop...

PR #810, opened 2015-06-29
Migrate from 'requests' for an HTTP client to 'httplib2'

phad:
This is being done because 'requests' depends on 'urllib3' which is thought to have security issues.

### DIFF
--- a/python/ct/client/log_client.py
+++ b/python/ct/client/log_client.py
@@ -1,14 +1,16 @@
 """RFC 6962 client API."""
 import base64
 import json
-
+import collections
 from ct.client.db import database
 from ct.crypto import verify
 from ct.proto import client_pb2
 import gflags
+import httplib2
 import logging
 import random
-import requests
+import urllib
+import urlparse
 
 from twisted.internet import defer
 from twisted.internet import error
@@ -196,9 +198,8 @@ def _parse_consistency_proof(response, servername):
 # A class that we can mock out to generate fake responses.
 class RequestHandler(object):
     """HTTPS requests."""
-    def __init__(self, connection_timeout=60, ca_bundle=True):
-      self._timeout = connection_timeout
-      self._ca_bundle = ca_bundle
+    def __init__(self, connection_timeout=60, ca_bundle=None):
+        self._http = httplib2.Http(timeout=connection_timeout, ca_certs=ca_bundle)
 
     def __repr__(self):
         return "%r()" % self.__class__.__name__
@@ -209,17 +210,20 @@ class RequestHandler(object):
     def get_response(self, uri, params=None):
         """Get an HTTP response for a GET request."""
         try:
-            return requests.get(uri, params=params, timeout=self._timeout,
-                                verify=self._ca_bundle)
-        except requests.exceptions.RequestException as e:
+            return self._build_requests_style_response(
+                self._http.request(self._uri_with_params(uri, params)))
+        except httplib2.HttpLib2Error as e:
             raise HTTPError("Connection to %s failed: %s" % (uri, e))
 
     def post_response(self, uri, post_data):
         try:
-            return requests.post(uri, data=json.dumps(post_data),
-                                 timeout=self._timeout, verify=self._ca_bundle)
-        except requests.exceptions.RequestException as e:
+            return self._build_requests_style_response(
+                self._http.request(uri, "POST", json.dumps(post_data)))
+        except httplib2.HttpLib2Error as e:
             raise HTTPError("POST to %s failed: %s" % (uri, e))
+
+    # Mimic the Response class from the requests API.
+    Response = collections.namedtuple('Response', ['status_code', 'reason', 'content', 'headers'])
 
     @staticmethod
     def check_response_status(code, reason, content='', headers=''):
@@ -231,6 +235,22 @@ class RequestHandler(object):
             raise HTTPServerError("%s (%s) %s" % (reason, content, headers))
         else:
             raise HTTPError("%s (%s) %s" % (reason, content, headers))
+
+    @staticmethod
+    def _uri_with_params(uri, params=None):
+        if not params:
+            return uri
+        components = list(urlparse.urlparse(uri))
+        if params:
+            # Update the URI query, which is at index 4 of the tuple.
+            components[4] = urllib.urlencode(params)
+        return urlparse.urlunparse(components)
+
+    @staticmethod
+    def _build_requests_style_response((resp_hdr, resp_body)):
+        status_code = int(resp_hdr.pop("status")) if "status" in resp_hdr else 0
+        reason = resp_hdr["reason"] if "reason" in resp_hdr else ""
+        return RequestHandler.Response(status_code, reason, resp_body, resp_hdr)
 
     def get_response_body(self, uri, params=None):
         response = self.get_response(uri, params=params)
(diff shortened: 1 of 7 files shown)


python/ct/client/log_client.py
```diff
@@ -197,8 +199,10 @@ def _parse_consistency_proof(response, servername):
 class RequestHandler(object):
     """HTTPS requests."""
     def __init__(self, connection_timeout=60, ca_bundle=True):
-      self._timeout = connection_timeout
-      self._ca_bundle = ca_bundle
+        if isinstance(ca_bundle, str):
```
eranmes: (1) the API doesn't make sense anymore and (2) the isinstance check counters Python's duck-typing.
I propose to set ca_bundle to None by default and just pass it in (should be OK according to module documentation: http://httplib2.googlecode.com/hg/doc/html/libhttplib2.html)

python/ct/client/log_client.py
```diff
@@ -209,18 +213,29 @@ def __str__(self):
...
         try:
-            return requests.post(uri, data=json.dumps(post_data),
-                                 timeout=self._timeout, verify=self._ca_bundle)
-        except requests.exceptions.RequestException as e:
+            return RequestHandler._build_requests_style_response(
+                self._http.request(uri, "POST", json.dumps(post_data)))
+        except Exception as e:
             raise HTTPError("POST to %s failed: %s" % (uri, e))
 
+    class Response(object):
```
eranmes: How about using namedtuple for this? https://docs.python.org/2/library/collections.html#collections.namedtuple

python/ct/client/log_client.py
```diff
@@ -232,6 +247,25 @@ def check_response_status(code, reason, content='', headers=''):
...
+            # Update the URI query, which is at index 4 of the tuple.
+            components[4] = urllib.urlencode(params)
+        return urlparse.urlunparse(components)
+
+    @staticmethod
+    def _build_requests_style_response((resp_hdr, resp_body)):
+        status_code = 0
+        if resp_hdr.has_key("status"):
+            status_code = int(resp_hdr["status"])
+            resp_hdr.pop("status")
```
eranmes: status_code = int(resp_hdr.pop("status")) if "status" in resp_hdr else 0

python/ct/client/log_client.py
```diff
@@ -209,18 +213,29 @@ def __str__(self):
     def get_response(self, uri, params=None):
         """Get an HTTP response for a GET request."""
         try:
-            return requests.get(uri, params=params, timeout=self._timeout,
-                                verify=self._ca_bundle)
-        except requests.exceptions.RequestException as e:
+            return RequestHandler._build_requests_style_response(
+                self._http.request(RequestHandler._uri_with_params(uri, params)))
+        except Exception as e:
```
eranmes: Isn't this except clause a bit broad? Does httplib2 not specify the type of exceptions it'll raise?

python/ct/client/scanner.py
```diff
@@ -105,11 +105,11 @@ def _scan(entry_queue, log_url, range_description):
         entries = client.get_entries(range_start, range_end)
         scanned = range_start
         for entry in entries:
-            scanned += 1
             # Can't pickle protocol buffers with protobuf module version < 2.5.0
             # (https://code.google.com/p/protobuf/issues/detail?id=418)
             # so send serialized entry.
             entry_queue.put((scanned, entry.SerializeToString()))
+            scanned += 1
```
eranmes: That belonged in another pull request, I believe?

python/utilities/ev_whitelist/create_ev_whitelist_zip.py
```diff
@@ -43,12 +43,15 @@
 class EVCRXHandler(object):
     """Produces updated EV whitelist Chrome extension."""
     def __init__(self, crx_url=CRX_URL):
-        req = requests.get(crx_url)
-        if not req.ok:
+        http = httplib2.Http()
+        http.force_exception_to_status_code = True
+        resp_hdr, resp_body = http.request(crx_url)
+        status = (resp_hdr["status"] and int(resp_hdr["status"])) or 0
```
eranmes: nit: Use the same idiom of int(resp_hdr["status"]) if "status" in resp_hdr or 0?

python/utilities/ev_whitelist/create_ev_whitelist_zip.py
```diff
@@ -43,12 +43,15 @@
 class EVCRXHandler(object):
     """Produces updated EV whitelist Chrome extension."""
     def __init__(self, crx_url=CRX_URL):
-        req = requests.get(crx_url)
-        if not req.ok:
+        http = httplib2.Http()
+        http.force_exception_to_status_code = True
+        resp_hdr, resp_body = http.request(crx_url)
```
eranmes: extra-pedantic nit: expand the names to response_headers, response_body. The characters saved do not seem justified (I also read that somewhere, will have to look up the reference).

python/utilities/logobserver/logobserver.py
```diff
@@ -39,13 +39,15 @@ def create_directory(directory):
...
-                f.write(list_.content)
-    except requests.exceptions.RequestException:
+    http = httplib2.Http(timeout=5)
+    http.force_exception_to_status_code = True
+    resp_hdr, resp_body = http.request(tld_list.TLD_LIST_ADDR)
+    if resp_hdr["status"] and int(resp_hdr["status"]) == 200:
+        create_directory(FLAGS.tld_list_dir)
+        with open('/'.join((FLAGS.tld_list_dir, "tld_list")), 'w') as f:
+            f.write(resp_body)
+            logging.info("Got TLD list: %s", resp_body[:256])
```
eranmes: nit: Remove the logging statement.

python/ct/client/log_client.py
phad: It wasn't clear to me why it was OK for a parameter to be either true/false or a string.
I'll make the change not least because the result will be a lot nicer.

python/ct/client/log_client.py
phad: Great tip, thanks!

python/ct/client/log_client.py
phad: Yes, that's better, thanks.

python/ct/client/log_client.py
phad: There's a top-level httplib2 exception class which I missed, so now using that.

python/ct/client/scanner.py
phad: Damn.

python/utilities/ev_whitelist/create_ev_whitelist_zip.py
phad: done

python/utilities/ev_whitelist/create_ev_whitelist_zip.py
phad: OK.  My opinion is that it's readable enough, and certainly not as terse as Go. :)

python/utilities/logobserver/logobserver.py
phad: done

python/ct/client/log_client.py
```diff
@@ -209,18 +210,21 @@ def __str__(self):
     def get_response(self, uri, params=None):
         """Get an HTTP response for a GET request."""
         try:
-            return requests.get(uri, params=params, timeout=self._timeout,
-                                verify=self._ca_bundle)
-        except requests.exceptions.RequestException as e:
+            return RequestHandler._build_requests_style_response(
```
eranmes: That's a correct but a bit awkward way for invoking a static method. I suggest using self._build_requests... to match how check_response_status is called.

phad: Done.

phad:
Any more comments on this PR?

python/ct/client/scanner.py
```diff
@@ -120,7 +120,7 @@ def _scan(entry_queue, log_url, range_description):
 
 
 def _get_tree_size(log_url):
-    client = log_client.LogClient(log_url)
+    client = log_client.LogClient(log_url, ca_bundle=None)
```
eranmes: No need to specify the ca_bundle parameter if you're passing in a None value.

python/ct/client/log_client.py
```diff
@@ -232,6 +236,22 @@ def check_response_status(code, reason, content='', headers=''):
...
+        components = list(urlparse.urlparse(uri))
+        if params:
+            # Update the URI query, which is at index 4 of the tuple.
+            components[4] = urllib.urlencode(params)
+        return urlparse.urlunparse(components)
+
+    @staticmethod
+    def _build_requests_style_response((resp_hdr, resp_body)):
+        status_code = int(resp_hdr.pop("status")) if "status" in resp_hdr else 0
+        reason = resp_hdr["reason"] if "reason" in resp_hdr else ""
```
eranmes: Please find attached the letter 'o', 'n', 's', 'e' and 'a' for complete and consistent naming of the response header and response body parameters.

phad: :)

PR #812, opened 2015-07-03
Add OIDs and algorithm IDs for algorithms in RFCs 5758 and 3447.

phad:


python/ct/crypto/asn1/oid.py
```diff
@@ -165,17 +165,24 @@ def _decode_value(cls, buf, strict=True):
...
 ID_RSASSA_PSS = ObjectIdentifier(value="1.2.840.113549.1.1.10")
 # RFC 4491
 ID_GOSTR3411_94_WITH_GOSTR3410_94 = ObjectIdentifier(value="1.2.643.2.2.4")
 ID_GOSTR3411_94_WITH_GOSTR3410_2001 = ObjectIdentifier(value="1.2.643.2.2.3")
-
+# RFC 3447
+ID_SHA256_WITH_RSA_ENCRYPTION = ObjectIdentifier(value="1.2.840.113549.1.1.11")
+ID_SHA384_WITH_RSA_ENCRYPTION = ObjectIdentifier(value="1.2.840.113549.1.1.12")
+ID_SHA512_WITH_RSA_ENCRYPTION = ObjectIdentifier(value="1.2.840.113549.1.1.13")
+ID_SHA224_WITH_RSA_ENCRYPTION = ObjectIdentifier(value="1.2.840.113549.1.1.14")
```
RJPercival: The SHA224 OID is from RFC 4055 (it does not appear in RFC 3447), unlike the other three.

phad: I've fixed that now, thanks for spotting it.

PR #813, opened 2015-07-03
Signature params can be NULL according to various RFCs.

phad:
Check before trying to copy from them.
In the interest of robustness do the same for signature
algorithm itself, although you'd expect that to always be present.


eranmes:
From an offlien discussion: We want to find a certificate where the parameters are missing and add it as a test case.

PR #817, opened 2015-07-06
Quieten down quiescent log noise a bit.

AlCutter:


cpp/log/cluster_state_controller-inl.h
```diff
@@ -348,8 +348,9 @@ void ClusterStateController<Logged>::OnServingSthUpdated(
     }
 
     actual_serving_sth_.reset(new ct::SignedTreeHead(update.handle_.Entry()));
-    LOG(INFO) << "Received new Serving STH:\n"
-              << actual_serving_sth_->DebugString();
+    LOG(INFO) << "Received new Serving STH: timestamp: "
+              << actual_serving_sth_->timestamp()
+              << " tree_size: " << actual_serving_sth_->tree_size();
```
pphaneuf: Maybe the ShortDebugString() would be helpful here? Not 100% sure if that's the right name, but there's a variant that uses a single line, for an STH, that shouldn't be excessive, and give us all the info (you just know that if you skip something, that the bit you'll be needing!)?

pphaneuf:
One idea, but looks good to me as it is, it's your call on that one.

cpp/log/cluster_state_controller-inl.h
AlCutter: It's gonna be quite a long line, but yeah, you just know that we'll not have the bit we need!
Done.

Issue #787, opened 2015-06-17
add a build identifier


pphaneuf:
We need a way to identify builds, at the very minimum reported by some sort of `--version` flag, but possibly reported in metrics, `Server`/`User-Agent` headers, dashboard pages, etc...

pphaneuf:
This script might be helpful:

https://stuff.mit.edu/afs/athena/astaff/project/opssrc/cups/autoconf-2.65/build-aux/git-version-gen

PR #819, opened 2015-07-06
Include (and print) git version info.

AlCutter:
Addresses #787

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,8 +12,15 @@ BUILT_SOURCES = \
 	cpp/gtest-all.cc \
 	cpp/monitoring/prometheus/metrics.pb.cc \
 	cpp/monitoring/prometheus/metrics.pb.h \
+	cpp/version.h \
 	proto/ct.pb.cc \
 	proto/ct.pb.h
+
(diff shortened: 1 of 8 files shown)


Makefile.am
```diff
@@ -12,9 +12,16 @@ BUILT_SOURCES = \
...
 	cpp/monitoring/prometheus/metrics.pb.h \
+	cpp/version.h \
 	proto/ct.pb.cc \
 	proto/ct.pb.h
 
+BUILD_VERSION=`git describe --abbrev=0 --dirty --always --tags`
+
+.PHONY: cpp/version.h
+cpp/version.h:
+	echo "const char kBuildVersion[] = \"$(BUILD_VERSION)\";" > $@-t && mv $@-t $@
```
pphaneuf: Put in an `$(AM_V_GEN)`, like for other generated files.

pphaneuf:
Would have preferred a `.cc` file (to avoid recompiling all those somewhat heavier files, but just compile one and relink instead), but seems okay for now.

Makefile.am
AlCutter: Done

pphaneuf:
LGTM

Issue #770, opened 2015-06-04
LevelDB syncing


pphaneuf:
Right now, LevelDB is operating as the equivalent of the SQLite [synchronous OFF](https://www.sqlite.org/pragma.html#pragma_synchronous) mode. I don't think we need to sync on every write, but when we claim that we have a certain STH, we should do so after some kind of barrier.

I'm hoping that re-writing the current STH with the "sync" option should be sufficient? Need to validate this...

PR #820, opened 2015-07-06
Write STHs to LevelDB synchronously.

AlCutter:
#770 
Docs at https://github.com/google/leveldb/blob/master/doc/index.html suggest that this "hybrid" approach as they call it is fine.  (Otherwise there's always the synchronous "WriteBatch" approach.)


pphaneuf:
The only problem I see with this is that this will sync *after* we've claimed that we had a certain tree size... We write the (serving) STH to disk well after we've told other nodes that we had this tree size.

I guess this is better than nothing, though?

AlCutter:
This really needs the DB API widening to include a `Sync()` method, we'll do that in a follow up - this at least brings the LevelDB impl up to parity with SQLite.

pphaneuf:
I agree, LGTM.

PR #822, opened 2015-07-08
Add support for precertificate verification.

plietar:
In addition to the already supported plain X.509 certificate,
precertificates and certificates embedding an SCT list can be verified.
PreCertificates signed with special purpose CA certificates are also
supported.

### DIFF
--- a/python/ct/crypto/cert.py
+++ b/python/ct/crypto/cert.py
@@ -189,6 +189,10 @@ class Certificate(object):
     def is_identical_to(self, other_cert):
         return self.to_der() == other_cert.to_der()
 
+    def to_asn1(self):
+        """Get a copy of the ASN.1 representation of the certificate."""
+        return x509.Certificate.decode(self._asn1_cert.encode())
+
     def get_extensions(self):
         """Get all extensions.
 
@@ -617,6 +621,20 @@ class Certificate(object):
         h.update(self._asn1_cert.encode())
         return h.digest()
 
+    def key_hash(self, hashfunc="sha1"):
+        """Get the certificate's public key hash.
+
+        Args:
+            hashfunc: name of a hash function. Algorithms always present are
+                'md5', 'sha1', 'sha224', 'sha256', 'sha384', and 'sha512'.
+        Returns:
+            a (binary) hash digest of the public key.
+        """
+        h = hashlib.new(hashfunc)
+        h.update(
+            self._asn1_cert["tbsCertificate"]["subjectPublicKeyInfo"].encode())
+        return h.digest()
+
     def key_usage(self, key_usage):
         """Whether the certificate has the given key usage asserted.
 
@@ -826,6 +844,19 @@ class Certificate(object):
         return [a[x509_ext.ACCESS_LOCATION] for a in aia
                 if a[x509_ext.ACCESS_METHOD] == oid.ID_AD_OCSP]
 
+    def has_extension(self, extn_id):
+        """Check if certificate contains a given extnsion.
+
+        Args:
+            extn_id: extension OID.
+
+        Returns:
+            True or False
+
+        Raises:
+            CertificateError: corrupt extension, or multiple extension values.
+        """
+        return self._get_decoded_extension_value(extn_id) is not None
 
 def certs_from_pem(pem_string, skip_invalid_blobs=False, strict_der=True):
     """Read multiple PEM-encoded certificates from a string.
--- a/python/ct/crypto/verify.py
+++ b/python/ct/crypto/verify.py
@@ -7,6 +7,9 @@ import struct
 from ct.crypto import error
 from ct.crypto import merkle
 from ct.crypto import pem
+from ct.crypto.asn1 import oid
+from ct.crypto.asn1 import x509_extension as x509_ext
+from ct.crypto.asn1 import x509_name
 from ct.proto import client_pb2
 from ct.proto import ct_pb2
 from ct.serialization import tls_message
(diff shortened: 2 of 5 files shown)


python/ct/crypto/asn1/oid.py
```diff
@@ -327,7 +332,11 @@ def _decode_value(cls, buf, strict=True):
     ID_AD_CA_ISSUERS: ("id-ad-caIssuers", "caIssuers"),
     ID_QT_CPS: ("id-qt-cps", "CPS"),
     ID_QT_UNOTICE: ("id-qt-unotice", "UserNotice"),
-    ANY_POLICY: ("anyPolicy", "anyPolicy")
+    ANY_POLICY: ("anyPolicy", "anyPolicy"),
+
+    EMBEDDED_SCT_LIST: ("ctEmbeddedSCT", "ctEmbeddedSCT"),
+    POISON: ("ctPoison", "ctPoison"),
+    PRECERTIFICATE_SIGNING: ("ctPresign", "ctPresign")
```
eranmes: Nit: The more informative name is Precertificate Signing Cert

python/ct/crypto/tools/verify_util.py
```diff
@@ -13,7 +13,7 @@
 
   For example:
 
-  verify_util.py verify_sct --sct=cert_sct.tls --log_key=log_key.pem cert.pem
+  verify_util.py verify_sct --sct=cert_sct.tls --log_key=log_key.pem cert.pem [cert.pem...]
```
eranmes: This chain should not be necessary - a single PEM file can contain multiple certificates and that's the convention around the codebase.

python/ct/crypto/tools/verify_util.py
```diff
@@ -61,15 +61,20 @@ def main(argv):
...
     if command != "verify_sct":
         exit_with_message("Unknown command %s" % command)
 
-    if not cert_file:
+    if not chain:
         exit_with_message("No certificate file given")
 
-    verify_sct(cert.Certificate.from_pem_file(cert_file, strict_der=False),
+    chain = map(lambda name:
+                    cert.Certificate.from_pem_file(name, strict_der = True),
```
eranmes: The Certificate.from_pem_file returns an iterable of certs from the file, which is why you don't need to specify multiple files.

python/ct/crypto/cert.py
```diff
@@ -184,11 +184,15 @@ def to_der(self):
         return self._asn1_cert.encode()
 
     def to_pem(self):
-      return pem.to_pem(self._asn1_cert.encode(), self.PEM_MARKERS[0])
+        return pem.to_pem(self._asn1_cert.encode(), self.PEM_MARKERS[0])
```
eranmes: Unnecessary whitespace change.

python/ct/crypto/verify.py
```diff
@@ -211,24 +333,15 @@ def verify_sct(self, sct, chain):
             ct.crypto.error.EncodingError: failed to encode signature input,
                 or decode the signature.
             ct.crypto.error.SignatureError: invalid signature.
+            ct.crypto.error.MissingCertificateError: a certificate is missing.
```
eranmes: nit: Add "from the chain".
A better error would be "MissingIssuerCertificate" so it's clear which one is missing (and why).

python/ct/crypto/verify.py
```diff
@@ -50,6 +53,128 @@ def decode_signature(signature):
...
+            })
+        ]
+
+    return asn1["tbsCertificate"].encode()
+
+def is_precertificate(cert):
+    return (cert.has_extension(oid.CT_POISON) or
+            cert.has_extension(oid.CT_EMBEDDED_SCT_LIST))
+
+def create_dst_entry(sct, chain):
```
eranmes: Make this a private method.

python/ct/crypto/verify.py
plietar: The error is also used when the leaf certificate is missing

eranmes: In this case, how about IncompleteChainError? that would be more indicative of the error.

python/ct/crypto/cert.py
```diff
@@ -378,21 +382,33 @@ def subject_alternative_names(self):
         sans = self._get_decoded_extension_values(oid.ID_CE_SUBJECT_ALT_NAME)
         return sum([list(san) for san in sans], [])
 
-    def subject(self):
-        """Returns list of subject field values in fashion similar to
-        subject_alternative_names fashion.
+    def subject(self, flatten=True):
```
eranmes: The change here is for a very specific use case so I recommend leaving it out of the general API.
Instead, get the value from the asn1 the in verify.py - the same way you do for the other extensions.

python/ct/crypto/verify.py
```diff
@@ -50,6 +53,129 @@ def decode_signature(signature):
                                    len(remaining)))
     return (hash_algo, sig_algo, remaining)
 
+def get_issuer(chain):
```
eranmes: prefix with underscore to indicate private method and name _get_precertificate_issuer so it's clear this is useful in precertificate serialization context.

python/ct/crypto/verify.py
```diff
@@ -50,6 +53,129 @@ def decode_signature(signature):
...
+        return issuer
+    else:
+        try:
+            return chain[2]
+        except IndexError:
+            raise error.MissingCertificateError(
+                "Chain with PreCertificate signed by PreCertificate "
+                "Signing Cert must contain issuer.")
+
+def find_extension(asn1, extn_id):
```
eranmes: prefix with underscore to indicate it shouldn't be used outside this module.

python/ct/crypto/verify.py
```diff
@@ -50,6 +53,129 @@ def decode_signature(signature):
...
+        The decoded value of the extension, or None if not found.
+        This is a reference and can be modified.
+    """
+    for e in asn1["tbsCertificate"]["extensions"]:
+        if e["extnID"] == extn_id:
+            return e["extnValue"].decoded_value
+
+    return None
+
+def remove_extension(asn1, extn_id):
```
eranmes: Same as for find_extension.

python/ct/crypto/verify.py
```diff
@@ -50,6 +53,129 @@ def decode_signature(signature):
...
+    Args:
+        asn1: x509.Certificate instance.
+        extn_id: OID of the extension to be removed.
+    """
+    asn1["tbsCertificate"]["extensions"] = (
+        filter(lambda e: e["extnID"] != extn_id,
+               asn1["tbsCertificate"]["extensions"])
+    )
+
+def encode_tbs_certificate_for_validation(cert, issuer):
```
eranmes: Same.

eranmes:
(hopefully) final round of review comments - should be good to go once those are addressed.

python/ct/crypto/verify_test.py
```diff
@@ -249,37 +249,71 @@ def test_verify_sth_consistency_invalid_proof(self):
...
-    def test_verify_sct_valid_signature(self):
-        root_cert = cert.Certificate.from_pem_file(
-            os.path.join(FLAGS.testdata_dir, 'ca-cert.pem'))
-        test_cert = cert.Certificate.from_pem_file(
-            os.path.join(FLAGS.testdata_dir, 'test-cert.pem'))
+    def _test_verify_sct(self, proof, chain, timestamp = None):
         sct = client_pb2.SignedCertificateTimestamp()
-        tls_message.decode(read_testdata_file('test-cert.proof'), sct)
+        tls_message.decode(read_testdata_file(proof), sct)
+        if timestamp is not None:
```
eranmes: In which case would the timestamp be None? you should have it for all test data, don't you?

plietar: `timestamp` is a fake value used to modify the SCT to make timestamp validation fail.
The real value is loaded just above.

plietar: I'll change the name to make it more obvious

PR #823, opened 2015-07-08
Set the gflags version string.

pphaneuf:
This makes `--version` return the build version string.


pphaneuf:
Maybe we should use google::VersionString() when logging the version? This also has the smell that we might want some sort of global init function to do these sorts of things...

pphaneuf:
This is to help with #787, of course.

AlCutter:
Using `google::VersionString()` SGTM, feel free to add it.

PR #825, opened 2015-07-08
Put the build version in an object file.

pphaneuf:
This builds a bit more quickly.

### DIFF
--- a/cpp/version.cc
+++ b/cpp/version.cc
@@ -1,0 +1,13 @@
+#include "version.h"
+
+#ifndef BUILD_VERSION
+#error Must specify -DBUILD_VERSION=... when building this file.
+#endif
+
+namespace cert_trans {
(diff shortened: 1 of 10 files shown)


cpp/version.cc
```diff
@@ -0,0 +1,12 @@
+#include "version.h"
+
+#ifndef BUILD_VERSION
+#error Must specify -DBUILD_VERSION=... when building this file.
+#endif
+
+namespace cert_trans {
+
```
AlCutter: Extra blank line here or remove one of the two on lines 8 & 9 in the `version.h` file

AlCutter:
LGTM, one nit.

cpp/version.cc
pphaneuf: Added.

PR #829, opened 2015-07-10
Add method to extract and verify SCTs from certificate.

plietar:
This is based on PR #822.

### DIFF
--- a/python/ct/crypto/cert.py
+++ b/python/ct/crypto/cert.py
@@ -850,6 +850,22 @@ class Certificate(object):
         return [a[x509_ext.ACCESS_LOCATION] for a in aia
                 if a[x509_ext.ACCESS_METHOD] == oid.ID_AD_OCSP]
 
+    def embedded_sct_list(self):
+        """Get the encoded list of embedded timestamps
+
+        Returns:
+            bytes representing a TLS encoded SignedCertificateTimestampList,
+            or None if not present
+
+        Raises:
+            CertificateError: corrupt extension, or multiple extension values.
+        """
+        signed_certificate_timestamp_list = (
+                self._get_decoded_extension_value(oid.CT_EMBEDDED_SCT_LIST))
+        if signed_certificate_timestamp_list is None:
+            return None
+        return signed_certificate_timestamp_list.value
+
     def has_extension(self, extn_id):
         """Check if certificate contains a given extnsion.
 
--- a/python/ct/crypto/verify.py
+++ b/python/ct/crypto/verify.py
@@ -347,4 +347,46 @@ class LogVerifier(object):
 
         return self._verify(signature_input, sct.signature.signature)
 
+    def verify_embedded_scts(self, chain):
+        """Extract and verify SCTs embedded in an X.509 certificate.
+
+        Args:
+            chain: list of cert.Certificate instances.
+
+        Returns:
+            List of (SignedCertificateTimestamp, bool) pairs, one for each SCT
+                present in the certificate. The boolean is True if the
+                corresponding SCT is valid, False otherwise.
+
+        Raises:
+            ct.crypto.error.EncodingError: failed to encode signature input,
+                or decode the signature.
+            ct.crypto.error.IncompleteChainError: the chain is empty.
+        """
+
+        try:
+            leaf_cert = chain[0]
+        except IndexError:
+            raise error.IncompleteChainError(
+                    "Chain must contain leaf certificate.")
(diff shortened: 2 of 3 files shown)


CONTRIBUTORS
```diff
@@ -39,6 +39,7 @@ Nicholas Galbreath <nickg@client9.com>
 Oliver Weidner <Oliver.Weidner@gmail.com>
 Pascal Leroy <phl@google.com>
 Paul Hadfield <hadfieldp@google.com> <paul@phad.org.uk>
+Paul Lietar <lietar@google.com>
```
pphaneuf: Feel free to also put your personal email, of course (or not, as you prefer).

python/ct/crypto/cert.py
```diff
@@ -844,6 +844,21 @@ def ocsp_responders(self):
...
+    def embedded_sct_list(self):
+        """Get the encoded list of embedded timestamps
+
+        Returns:
+            bytes representing a TLS encoded SignedCertificateTimestampList
+
+        Raises:
+            CertificateError: corrupt extension, or multiple extension values.
+        """
+        esl = self._get_decoded_extension_value(oid.CT_EMBEDDED_SCT_LIST)
```
eranmes: Naming: esl is not a comprehensible variable name. embedded_sct_list is ok, signed_certificate_timestamp_list is even better because it maps directly to a data structure in RFC6962 so it's clear what it holds.

python/ct/crypto/cert.py
```diff
@@ -844,6 +844,21 @@ def ocsp_responders(self):
         return [a[x509_ext.ACCESS_LOCATION] for a in aia
                 if a[x509_ext.ACCESS_METHOD] == oid.ID_AD_OCSP]
 
+    def embedded_sct_list(self):
+        """Get the encoded list of embedded timestamps
+
+        Returns:
+            bytes representing a TLS encoded SignedCertificateTimestampList
```
eranmes: add: ... or None if not present.

python/ct/crypto/verify.py
```diff
@@ -347,4 +347,41 @@ def verify_sct(self, sct, chain):
...
+            List of (SignedCertificateTimestamp, bool) pairs, one for each SCT
+                present in the certificate. The boolean is True if the
+                corresponding SCT is valid, False otherwise.
+
+        Raises:
+            ct.crypto.error.EncodingError: failed to encode signature input,
+                or decode the signature.
+        """
+
+        leaf_cert = chain[0]
```
eranmes: Since you've added the IncompleteChainError, why not add it here?

eranmes:
Small comments, otherwise looks good.

python/ct/crypto/verify.py
plietar: `IncompleteChainError` is a subclass of `VerifyError`, so it gets caught by the try-except block, and False is set inside the list.

python/ct/crypto/cert.py
plietar: I've mimicked the methods above, but sure, I'll fix it

python/ct/crypto/cert.py
```diff
@@ -850,6 +850,23 @@ def ocsp_responders(self):
...
+            or None if not present
+
+        Raises:
+            CertificateError: corrupt extension, or multiple extension values.
+        """
+        signed_certificate_timestamp_list = (
+                self._get_decoded_extension_value(oid.CT_EMBEDDED_SCT_LIST))
+        if signed_certificate_timestamp_list is None:
+            return None
+        else:
```
eranmes: The else here is superfluous - remove it.

python/ct/crypto/verify.py
eranmes: If the chain is empty you'll get an IndexError before the try/catch block.

python/ct/crypto/verify.py
```diff
@@ -347,4 +347,41 @@ def verify_sct(self, sct, chain):
...
+
+        scts_blob = leaf_cert.embedded_sct_list()
+        if scts_blob is None:
+            return []
+
+        scts = client_pb2.SignedCertificateTimestampList()
+        tls_message.decode(scts_blob, scts)
+
+        result = list()
```
eranmes: Nit: 'result = []' is shorter.

eranmes:
A couple of small comments, LGTM otherwise.
Let me know when you've updated the PR and I'll merge.

Issue #811, opened 2015-07-02
The log server frequently crash when the master election.


tedwang2015:
The log server frequently crash for the master election.  

Please see the stack trace below. 

Environment ct log + etcd. 

```text
F0701 09:15:11.091508 22393 masterelection.cc:363] Check failed: task->status().ok() /root/election/c0b90f7d-a261-4a9e-9887-d9fdb5710ab9: FAILED_PRECONDITION: Invalid JSON: Couldn't find 'node'
*** Check failure stack trace: ***
    @     0x7fcc585ae20d  google::LogMessage::Fail()
    @     0x7fcc585affac  google::LogMessage::SendToLog()
    @     0x7fcc585ade2b  google::LogMessage::Flush()
    @     0x7fcc585b08be  google::LogMessageFatal::~LogMessageFatal()
    @           0x4f4952  cert_trans::MasterElection::ProposalUpdateDone()
    @           0x4fb0e5  _ZNKSt7_Mem_fnIMN10cert_trans14MasterElectionEFvRKSsPNS0_10EtcdClient8ResponseEPN4util4TaskEEEclIJRSsRS6_S9_EvEEvPS1_DpOT_
    @           0x4fab74  _ZNSt5_BindIFSt7_Mem_fnIMN10cert_trans14MasterElectionEFvRKSsPNS1_10EtcdClient8ResponseEPN4util4TaskEEEPS2_SsS7_St12_PlaceholderILi1EEEE6__callIvJOSA_EJLm0ELm1ELm2ELm3EEEET_OSt5tupleIJDpT0_EESt12_Index_tupleIJXspT1_EEE
    @           0x4fa031  _ZNSt5_BindIFSt7_Mem_fnIMN10cert_trans14MasterElectionEFvRKSsPNS1_10EtcdClient8ResponseEPN4util4TaskEEEPS2_SsS7_St12_PlaceholderILi1EEEEclIJSA_EvEET0_DpOT_
    @           0x4f938d  std::_Function_handler<>::_M_invoke()
    @           0x4fe77f  std::function<>::operator()()
    @           0x4fd725  util::Task::RunCleanupAndDoneCallbacks()
    @           0x500b5f  _ZNKSt7_Mem_fnIMN4util4TaskEFvvEEclIIEvEEvPS1_DpOT_
[...]

PR #832, opened 2015-07-15
Report the watched key when an error occur.

pphaneuf:
Trying to get some more insight on #811...


AlCutter:
Sorry, just saw this - you'll have to update it now, but it looked good if that's any consolation ;)

pphaneuf:
I'm more concerned about leaving that weird behaviour in the domain of the unknown...

PR #838, opened 2015-07-17
Add some pretty printers for fetcher stuff to make Mock failures read…

AlCutter:
…able.

### DIFF
--- a/cpp/net/url_fetcher.cc
+++ b/cpp/net/url_fetcher.cc
@@ -277,4 +277,42 @@ ostream& operator<<(ostream& output, const UrlFetcher::Response& resp) {
 }
 
 
+ostream& operator<<(ostream& output, const UrlFetcher::Request& req) {
+  output << "verb: " << req.verb << endl
+         << "url: " << req.url << endl
+         << "headers: " << req.headers << endl
+         << "body: <<EOF" << endl
+         << req.body << "EOF" << endl;
+  return output;
+}
+
+
+ostream& operator<<(ostream& os, const UrlFetcher::Verb& verb) {
+  switch (verb) {
+    case UrlFetcher::Verb::GET:
+      os << "GET";
+      break;
+    case UrlFetcher::Verb::POST:
+      os << "POST";
+      break;
+    case UrlFetcher::Verb::PUT:
+      os << "PUT";
+      break;
+    case UrlFetcher::Verb::DELETE:
+      os << "DELETE";
+      break;
+  }
+  return os;
+}
+
+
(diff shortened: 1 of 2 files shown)


cpp/net/url_fetcher.cc
```diff
@@ -276,5 +276,41 @@ ostream& operator<<(ostream& output, const UrlFetcher::Response& resp) {
...
+         << "headers: " << req.headers << endl
+         << "body: <<EOF" << endl
+         << req.body << "EOF" << endl;
+  return output;
+}
+
+ostream& operator<<(ostream& os, const UrlFetcher::Verb& verb) {
+  switch (verb) {
+    case UrlFetcher::Verb::GET:
+      os << "GET ";
```
pphaneuf: I don't like those trailing spaces... I can add them when needed, but I removing them would be annoying.

cpp/net/url_fetcher.cc
```diff
@@ -276,5 +276,41 @@ ostream& operator<<(ostream& output, const UrlFetcher::Response& resp) {
...
+      break;
+  }
+  return os;
+}
+
+
+ostream& operator<<(ostream& os, const UrlFetcher::Headers& headers) {
+  os << "{";
+  for (const auto& h : headers) {
+    os << h.first << ": " << h.second;
```
pphaneuf: Nothing to separate each header?

pphaneuf:
Review done, just minor things...

cpp/net/url_fetcher.cc
AlCutter: removed

cpp/net/url_fetcher.cc
AlCutter: added `endl`s

AlCutter:
Squashed them in already since they're minor, ready to merge.

pphaneuf:
Hmm, something happened in your squashing? Some other commits in there?

cpp/net/url_fetcher.cc
```diff
@@ -276,5 +276,41 @@ ostream& operator<<(ostream& output, const UrlFetcher::Response& resp) {
...
 
+ostream& operator<<(ostream& output, const UrlFetcher::Request& req) {
+  output << "verb: " << req.verb << endl
+         << "url: " << req.url << endl
+         << "headers: " << req.headers << endl
+         << "body: <<EOF" << endl
+         << req.body << "EOF" << endl;
+  return output;
+}
+
```
pphaneuf: Only one line here (and before the preceding definition), but two later on? We should probably pick one for the file... (I vote for two!) :smiley:

PR #839, opened 2015-07-17
Log etcd version(s)

AlCutter:
Logs the version of the etcd server we're currently talking to.
Does the reporting lazily - i.e. waits for the first "real" request before requesting the version.
If we get redirected to another etcd it'll log the version string from there too.

### DIFF
--- a/cpp/util/etcd.h
+++ b/cpp/util/etcd.h
@@ -14,6 +14,7 @@
 #include "base/macros.h"
 #include "net/url_fetcher.h"
 #include "util/status.h"
+#include "util/sync_task.h"
 #include "util/task.h"
 #include "util/util.h"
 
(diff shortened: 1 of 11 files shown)


cpp/util/etcd.h
```diff
@@ -145,9 +147,14 @@ class EtcdClient {
   void WatchRequestDone(WatchState* state, GetResponse* gen_resp,
                         util::Task* child_task);
 
+  void MaybeLogEtcdVersion();
+
+  util::Executor* const executor_;
+  std::unique_ptr<util::SyncTask> log_version_task_;
```
pphaneuf: Why have this in a `unique_ptr`, instead of by value? It's created at startup, and lasts the whole time...

Incidentally, maybe add a comment to `~EtcdClient` to mention it could block.

pphaneuf:
Why was it lazy, again? I think it might have had to do with not having an executor, but you put one in. Seems to be like it'd be simpler if it was called in the spots where the bool is reset? And it wouldn't be "maybe", of course. :smiley:

pphaneuf:
That last comment is kind of optional, I'd say... Just thinking this might be some extra complication for an obsolete reason, but this should work fine. The only bit I don't like is the `SyncTask` that's not a value.

pphaneuf:
And will need a rebase, it looks like!

AlCutter:
It's lazy just to make testing easier, otherwise it'll start trying to fetch version info in the c'tor and race the setting of the mock expectation.

I guess it doesn't need to continue being lazy after that, but I figured it'd be less code like that than special-case it.

The `SyncTask` is a `unique_ptr` because there's a testing default c'tor in `EtcdClient` which sets the `executor_` to `nullptr`, and that wouldn't play very nicely with the `CHECK_NOTNULL` in `SyncTask::SyncTask()`...

pphaneuf:
It'd have to change the test class, but you could set the expectation on UrlFetcher before instantiating... Anyway, LGTM.

PR #841, opened 2015-07-21
Factor out common binary-init stuff

AlCutter:


cpp/util/util.cc
```diff
@@ -191,4 +199,27 @@ string ToBase64(const string& from) {
   return ret;
 }
 
+void InitCT(int* argc, char** argv[]) {
+  google::SetVersionString(cert_trans::kBuildVersion);
+  google::ParseCommandLineFlags(argc, argv, true);
+  google::InitGoogleLogging(*argv[0]);
+  google::InstallFailureSignalHandler();
+
+
```
pphaneuf: Spurious line.

AlCutter: Zapped

cpp/client/ct.cc
```diff
@@ -981,10 +981,8 @@ int Monitor() {
 // Exit code upon abnormal exit (CHECK failures): != 0
 // (on UNIX, 134 is expected)
 int main(int argc, char** argv) {
+  util::InitCT(&argc, &argv);
```
pphaneuf: Pretty sure you have to set the usage message *before* calling `google::ParseCommandLineFlags`?

cpp/util/util.cc
```diff
@@ -191,4 +199,5 @@ string ToBase64(const string& from) {
   return ret;
 }
 
+
```
pphaneuf: All that I like two lines between top-level declarations, I think this file can probably be dropped from your PR? :wink:

pphaneuf:
Done with this review.

cpp/client/ct.cc
AlCutter: Good catch, done.

cpp/util/util.cc
AlCutter: That line makes all the difference, man.

pphaneuf:
LGTM

PR #842, opened 2015-07-21
Kill -Wextra warnings for g++ and clang++

### LINKED ISSUE #685: clean up warnings given by -Wextra
Both gcc and Clang have a ```-Wextra``` flag, for even more warnings than ```-Wall```, but it's quite a lot of warnings, so I'm leaving this here as a note to get back to it.

AlCutter:
Fixes #685

### DIFF
--- a/cpp/fetcher/remote_peer.cc
+++ b/cpp/fetcher/remote_peer.cc
@@ -10,6 +10,7 @@
 
 using cert_trans::LoggedCertificate;
 using ct::SignedTreeHead;
+using std::bind;
 using std::chrono::seconds;
 using std::lock_guard;
 using std::make_shared;
@@ -56,13 +57,13 @@ struct RemotePeer::Impl {
   mutex lock_;
   shared_ptr<SignedTreeHead> sth_;
 
-  void FetchSTH(util::Task* task);
+  void FetchSTH();
   void DoneGetSTH(const std::shared_ptr<ct::SignedTreeHead>& on_new_sth,
                   AsyncLogClient::Status status);
 };
 
 
-void RemotePeer::Impl::FetchSTH(util::Task* task) {
+void RemotePeer::Impl::FetchSTH() {
   if (CHECK_NOTNULL(task_)->CancelRequested()) {
     task_->Return(util::Status::CANCELLED);
     return;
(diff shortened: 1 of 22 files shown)


cpp/server/ct-server.cc
```diff
@@ -159,7 +160,8 @@ static bool ValidatePort(const char* flagname, int port) {
 static const bool port_dummy =
     RegisterFlagValidator(&FLAGS_port, &ValidatePort);
 
-static bool ValidateRead(const char* flagname, const string& path) {
+static bool ValidateRead(const char* flagname __attribute__((unused)),
```
benlaurie: But its used on line 166.

AlCutter: `__attribute__` induced blindness, removed.

benlaurie: Does the compiler not complain that you used an unused variable?

AlCutter: Surprisingly, no... But then the docs for that attribute say that is means that "...this variable is meant to possibly be unused, GCC *[and clang, empirically]* will not produce a warning for this variable."

That said, I'm not over the moon about the extra grit in the lotion these attributes bring (I originally like the documentation aspect of it), so I think I might just remove the parameter names instead.

cpp/server/ct-dns-server.cc
```diff
@@ -22,7 +22,7 @@ DEFINE_string(domain, "", "Domain");
 DEFINE_string(db, "", "Database for certificate and tree storage");
 
 // Basic sanity checks on flag values.
-static bool ValidatePort(const char* flagname, int32_t port) {
+static bool ValidatePort(const char*, int32_t port) {
```
pphaneuf: You know, now that I think about it, saying "this port is invalid" without telling them *which* port, that's a bit rude, especially if we know which port! :smiley:

cpp/fetcher/remote_peer.cc
```diff
@@ -67,35 +68,37 @@ void RemotePeer::Impl::DoneGetSTH(
...
-    LOG(WARNING) << "Unexpected tree size in new_sth:\n"
-                 << new_sth->DebugString();
-  } else if (sth_ && new_sth->tree_size() < sth_->tree_size()) {
-    LOG(WARNING) << "Received old STH:\n" << new_sth->DebugString();
-  } else if (!sth_ || new_sth->timestamp() > sth_->timestamp()) {
-    // This STH is good, we'll take it.
-    sth_ = new_sth;
-    if (on_new_sth_) {
-      on_new_sth_(*sth_);
+  if (status == AsyncLogClient::OK) {
```
pphaneuf: What's the warning that this fixed? It makes sense, somewhat, but I'm really curious of what turned this up!

pphaneuf: Maybe some log message, if this wasn't the case?

cpp/net/connection_pool.cc
```diff
@@ -188,7 +188,7 @@ ConnectionPool::ConnectionPool(libevent::Base* base)
 
 // static
 evhtp_res ConnectionPool::Connection::ConnectionErrorHook(
-    evhtp_connection_t* conn, evhtp_error_flags errtype, void* arg) {
+    evhtp_connection_t* conn, evhtp_error_flags, void* arg) {
```
pphaneuf: This looks like exactly the kind of information one would like to know when debugging... :smiley:

cpp/server/ct-dns-server.cc
```diff
@@ -22,7 +22,7 @@ DEFINE_string(domain, "", "Domain");
...
   if (port <= 0 || port > 65535) {
     std::cerr << "Port value " << port << " is invalid. " << std::endl;
```
pphaneuf: Or, could probably skip this line entirely, since the message that gflags give us is this (which is no less informative):

    ERROR: failed validation of new value '70000' for flag 'port'

pphaneuf:
Ok, done reviewing again.

cpp/net/connection_pool.cc
AlCutter: Actually, it turns out to always be some unhelpful value like "unknown" or whatever, or at least it was when I was writing the SSL stuff.

And if you want it, you can always put it back in while you're debugging and remove it once you're done.

cpp/server/ct-dns-server.cc
AlCutter: Removed.

pphaneuf:
LGTM

Issue #772, opened 2015-06-08
pass multiple etcd IPs to ct-server


pphaneuf:
At the moment, `ct-server` takes only one IP for etcd, but if that instance happens to be down when starting up (or if it is the leader and it goes down while running), this doesn't work (can't start up, or crashing during operation).

The `etcd_host` flag could have multiple IPs (doesn't have to be all of them, but the more you'd provide, the more robust it would be). Ideally, when `EtcdClient` starts up, it would request the list of peers from etcd, and possibly refresh that list every so often.

PR #843, opened 2015-07-21
EtcdClient supports passing in multiple etcd servers.

AlCutter:
Addresses #772 

Will update the `Dockerfile` etc. to actually use this in another PR.

### DIFF
--- a/cpp/server/ct-mirror.cc
+++ b/cpp/server/ct-mirror.cc
@@ -74,8 +74,8 @@ DEFINE_int32(log_stats_frequency_seconds, 3600,
              "Must be greater than 0.");
 DEFINE_int32(target_poll_frequency_seconds, 10,
              "How often should the target log be polled for updates.");
-DEFINE_string(etcd_hosts, "",
-              "Comma separated list of 'Hostname:Port' of the etcd server(s)");
+DEFINE_string(etcd_servers, "",
+              "Comma separated list of 'hostname:port' of the etcd server(s)");
 DEFINE_string(etcd_root, "/root", "Root of cluster entries in etcd.");
 DEFINE_int32(num_http_server_threads, 16,
              "Number of threads for servicing the incoming HTTP requests.");
@@ -286,7 +286,7 @@ int main(int argc, char* argv[]) {
         new FileStorage(FLAGS_meta_dir, 0));
   }
 
-  const bool stand_alone_mode(FLAGS_etcd_hosts.empty());
+  const bool stand_alone_mode(FLAGS_etcd_servers.empty());
   const shared_ptr<libevent::Base> event_base(make_shared<libevent::Base>());
   ThreadPool internal_pool(8);
   UrlFetcher url_fetcher(event_base.get(), &internal_pool);
@@ -294,7 +294,7 @@ int main(int argc, char* argv[]) {
   const std::unique_ptr<EtcdClient> etcd_client(
       stand_alone_mode
           ? new FakeEtcdClient(event_base.get())
-          : new EtcdClient(&url_fetcher, SplitHosts(FLAGS_etcd_hosts)));
+          : new EtcdClient(&url_fetcher, SplitHosts(FLAGS_etcd_servers)));
 
   Server<LoggedCertificate>::Options options;
   options.server = FLAGS_server;
--- a/cpp/server/ct-mirror.cc
+++ b/cpp/server/ct-mirror.cc
@@ -74,8 +74,8 @@ DEFINE_int32(log_stats_frequency_seconds, 3600,
              "Must be greater than 0.");
 DEFINE_int32(target_poll_frequency_seconds, 10,
              "How often should the target log be polled for updates.");
-DEFINE_string(etcd_host, "", "Hostname of the etcd server");
-DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
+DEFINE_string(etcd_hosts, "",
+              "Comma separated list of 'Hostname:Port' of the etcd server(s)");
 DEFINE_string(etcd_root, "/root", "Root of cluster entries in etcd.");
 DEFINE_int32(num_http_server_threads, 16,
              "Number of threads for servicing the incoming HTTP requests.");
@@ -108,9 +108,10 @@ using cert_trans::PeriodicClosure;
 using cert_trans::Proxy;
 using cert_trans::ReadPublicKey;
 using cert_trans::RemotePeer;
-using cert_trans::Server;
 using cert_trans::ScopedLatency;
+using cert_trans::Server;
 using cert_trans::StrictConsistentStore;
+using cert_trans::SplitHosts;
 using cert_trans::ThreadPool;
 using cert_trans::Update;
 using cert_trans::UrlFetcher;
@@ -285,7 +286,7 @@ int main(int argc, char* argv[]) {
         new FileStorage(FLAGS_meta_dir, 0));
   }
 
-  const bool stand_alone_mode(FLAGS_etcd_host.empty());
+  const bool stand_alone_mode(FLAGS_etcd_hosts.empty());
   const shared_ptr<libevent::Base> event_base(make_shared<libevent::Base>());
   ThreadPool internal_pool(8);
   UrlFetcher url_fetcher(event_base.get(), &internal_pool);
@@ -293,7 +294,7 @@ int main(int argc, char* argv[]) {
   const std::unique_ptr<EtcdClient> etcd_client(
       stand_alone_mode
           ? new FakeEtcdClient(event_base.get())
-          : new EtcdClient(&url_fetcher, FLAGS_etcd_host, FLAGS_etcd_port));
+          : new EtcdClient(&url_fetcher, SplitHosts(FLAGS_etcd_hosts)));
 
   Server<LoggedCertificate>::Options options;
   options.server = FLAGS_server;
--- a/cpp/server/ct-server.cc
+++ b/cpp/server/ct-server.cc
@@ -77,8 +77,8 @@ DEFINE_int32(tree_signing_frequency_seconds, 600,
 DEFINE_double(guard_window_seconds, 60,
               "Unsequenced entries newer than this "
               "number of seconds will not be sequenced.");
-DEFINE_string(etcd_host, "", "Hostname of the etcd server");
-DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
+DEFINE_string(etcd_hosts, "",
+              "Comma separated list of 'Hostname:Port' of the etcd server(s)");
 DEFINE_string(etcd_root, "/root", "Root of cluster entries in etcd.");
 DEFINE_int32(num_http_server_threads, 16,
              "Number of threads for servicing the incoming HTTP requests.");
@@ -101,8 +101,9 @@ using cert_trans::HttpHandler;
 using cert_trans::Latency;
 using cert_trans::LoggedCertificate;
 using cert_trans::ReadPrivateKey;
-using cert_trans::Server;
 using cert_trans::ScopedLatency;
+using cert_trans::Server;
+using cert_trans::SplitHosts;
 using cert_trans::ThreadPool;
 using cert_trans::TreeSigner;
 using cert_trans::Update;
@@ -370,7 +371,7 @@ int main(int argc, char* argv[]) {
   ThreadPool internal_pool(8);
   UrlFetcher url_fetcher(event_base.get(), &internal_pool);
 
-  const bool stand_alone_mode(FLAGS_etcd_host.empty());
+  const bool stand_alone_mode(FLAGS_etcd_hosts.empty());
   if (stand_alone_mode && !FLAGS_i_know_stand_alone_mode_can_lose_data) {
     LOG(FATAL) << "attempted to run in stand-alone mode without the "
                   "--i_know_stand_alone_mode_can_lose_data flag";
@@ -381,7 +382,7 @@ int main(int argc, char* argv[]) {
   std::unique_ptr<EtcdClient> etcd_client(
       stand_alone_mode
           ? new FakeEtcdClient(event_base.get())
-          : new EtcdClient(&url_fetcher, FLAGS_etcd_host, FLAGS_etcd_port));
+          : new EtcdClient(&url_fetcher, SplitHosts(FLAGS_etcd_hosts)));
 
   Server<LoggedCertificate>::Options options;
   options.server = FLAGS_server;
--- a/cpp/server/ct-server.cc
+++ b/cpp/server/ct-server.cc
@@ -77,8 +77,8 @@ DEFINE_int32(tree_signing_frequency_seconds, 600,
 DEFINE_double(guard_window_seconds, 60,
               "Unsequenced entries newer than this "
               "number of seconds will not be sequenced.");
-DEFINE_string(etcd_hosts, "",
-              "Comma separated list of 'Hostname:Port' of the etcd server(s)");
+DEFINE_string(etcd_servers, "",
+              "Comma separated list of 'hostname:port' of the etcd server(s)");
 DEFINE_string(etcd_root, "/root", "Root of cluster entries in etcd.");
 DEFINE_int32(num_http_server_threads, 16,
              "Number of threads for servicing the incoming HTTP requests.");
@@ -371,7 +371,7 @@ int main(int argc, char* argv[]) {
   ThreadPool internal_pool(8);
   UrlFetcher url_fetcher(event_base.get(), &internal_pool);
 
-  const bool stand_alone_mode(FLAGS_etcd_hosts.empty());
+  const bool stand_alone_mode(FLAGS_etcd_servers.empty());
   if (stand_alone_mode && !FLAGS_i_know_stand_alone_mode_can_lose_data) {
     LOG(FATAL) << "attempted to run in stand-alone mode without the "
                   "--i_know_stand_alone_mode_can_lose_data flag";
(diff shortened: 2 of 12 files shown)


cpp/server/ct-mirror.cc
```diff
@@ -74,8 +74,8 @@ DEFINE_int32(log_stats_frequency_seconds, 3600,
              "Must be greater than 0.");
 DEFINE_int32(target_poll_frequency_seconds, 10,
              "How often should the target log be polled for updates.");
-DEFINE_string(etcd_host, "", "Hostname of the etcd server");
-DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
+DEFINE_string(etcd_hosts, "",
+              "Comma separated list of 'Hostname:Port' of the etcd server(s)");
```
pphaneuf: Wild capitalisation appears!

cpp/util/etcd_test.cc
```diff
@@ -698,6 +820,16 @@ TEST_F(EtcdTest, GetStoreStats) {
   EXPECT_EQ(16, response.stats["getsFail"]);
 }
 
+TEST_F(EtcdTest, SplitHosts) {
```
pphaneuf: Maybe another test with something bad? Like just a ",", "foo,,bar", etc...

cpp/util/etcd.cc
```diff
@@ -918,5 +966,17 @@ void EtcdClient::Generic(const string& key, const string& key_space,
                       bind(&EtcdClient::FetchDone, this, etcd_req, _1)));
 }
 
+list<EtcdClient::HostPortPair> SplitHosts(const string& hosts_string) {
+  vector<string> hosts(util::split(hosts_string, ','));
+  CHECK(!hosts.empty());
```
pphaneuf: Maybe print something (`<< "blabla"`) here? This could happen with invalid flags, best to be as informative as we can.

cpp/util/etcd.cc
```diff
@@ -918,5 +966,17 @@ void EtcdClient::Generic(const string& key, const string& key_space,
...
+
+  list<EtcdClient::HostPortPair> ret;
+  for (const auto& h : hosts) {
+    vector<string> hp(util::split(h, ':'));
+    CHECK_EQ(2, hp.size());
```
pphaneuf: Same thing here, explain this better, and print out `h`, so users can find the problem quicker.

cpp/util/etcd.cc
```diff
@@ -918,5 +966,17 @@ void EtcdClient::Generic(const string& key, const string& key_space,
...
+    ret.emplace_back(EtcdClient::HostPortPair(hp[0], stoi(hp[1])));
```
pphaneuf: With `stoi`, you might get 0 for something that doesn't parse, right? Might be a good idea to `CHECK_NE(0, port)` (with explanation)...

cpp/util/etcd.h
```diff
@@ -76,9 +81,10 @@ class EtcdClient {
 
   typedef std::function<void(const std::vector<Node>& updates)> WatchCallback;
 
-  // TODO(pphaneuf): This should take a set of servers, not just one.
   EtcdClient(UrlFetcher* fetcher, const std::string& host, uint16_t port);
 
+  EtcdClient(UrlFetcher* fetcher, const std::list<HostPortPair>& etcds);
```
pphaneuf: I'm almost tempted to say there should be just the one constructor... Could still use it fairly easily, with something like `{{host, port}}`, or `{make_pair(host, port)}`.

cpp/util/etcd.h
```diff
@@ -154,6 +160,12 @@ class EtcdClient {
 };
 
 
+// Splits strings of the form "host:port,host:port,..." to a list of
+// HostPortPairs.
+std::list<EtcdClient::HostPortPair> SplitHosts(
```
pphaneuf: I'd have made that one return an `std::vector`, but not a big deal...

cpp/util/etcd.h
```diff
@@ -2,6 +2,8 @@
 #define CERT_TRANS_UTIL_ETCD_H_
 
 #include <chrono>
+#include <glog/logging.h>
```
pphaneuf: What's that one for?

cpp/util/etcd.h
```diff
@@ -2,6 +2,8 @@
...
+#include <list>
```
pphaneuf: We'd need `<utility>`, for `pair<>`...

cpp/util/etcd.cc
```diff
@@ -668,10 +680,36 @@ bool EtcdClient::Node::HasExpiry() const {
 
 
 EtcdClient::EtcdClient(UrlFetcher* fetcher, const string& host, uint16_t port)
-    : fetcher_(CHECK_NOTNULL(fetcher)), endpoint_(host, port) {
-  CHECK(!endpoint_.first.empty());
-  CHECK_GT(endpoint_.second, 0);
+    : EtcdClient(fetcher, list<HostPortPair>{HostPortPair(host, port)}) {
```
pphaneuf: Do you even need the `list<HostPortPair>` here?

cpp/server/ct-mirror.cc
```diff
@@ -74,8 +74,8 @@ DEFINE_int32(log_stats_frequency_seconds, 3600,
              "Must be greater than 0.");
 DEFINE_int32(target_poll_frequency_seconds, 10,
              "How often should the target log be polled for updates.");
-DEFINE_string(etcd_host, "", "Hostname of the etcd server");
-DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
+DEFINE_string(etcd_hosts, "",
```
pphaneuf: Vague feeling that this isn't just hosts... `etcd_servers`? If you do change it, make it the same everywhere, of course.

cpp/server/ct-server.cc
```diff
@@ -77,8 +77,8 @@ DEFINE_int32(tree_signing_frequency_seconds, 600,
 DEFINE_double(guard_window_seconds, 60,
               "Unsequenced entries newer than this "
               "number of seconds will not be sequenced.");
-DEFINE_string(etcd_host, "", "Hostname of the etcd server");
-DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
+DEFINE_string(etcd_hosts, "",
+              "Comma separated list of 'Hostname:Port' of the etcd server(s)");
```
pphaneuf: Decapitalise here too.

cpp/server/ct-mirror.cc
AlCutter: aNy bEtTeR LiKe ThAt?

cpp/util/etcd_test.cc
AlCutter: Added some

cpp/tools/clustertool_main.cc
```diff
@@ -49,8 +50,8 @@ using util::Status;
 DEFINE_string(cluster_config, "",
               "Path of file containing the cluster config (in ASCII proto "
               "format.)");
-DEFINE_string(etcd_host, "", "Hostname of the etcd server");
-DEFINE_int32(etcd_port, 0, "Port of the etcd server.");
+DEFINE_string(etcd_hosts, "",
+              "Comma separated list of 'Hostname:Port' of the etcd server(s)");
```
pphaneuf: And here.

cpp/util/etcd.cc
```diff
@@ -668,10 +680,36 @@ bool EtcdClient::Node::HasExpiry() const {
...
+}
+
+
+EtcdClient::EtcdClient(UrlFetcher* fetcher, const list<HostPortPair>& etcds)
+    : fetcher_(CHECK_NOTNULL(fetcher)), etcds_(etcds) {
+  CHECK(!etcds_.empty());
```
pphaneuf: I know I didn't have them originally, but some explanation strings on those startup-time `CHECK`s would be nice.

cpp/util/etcd.cc
AlCutter: Removed that `CHECK`; it's impossible to be more specific here, and `EtcdClient()` does a `CHECK` that the passed in list isn't empty.

cpp/util/etcd.cc
```diff
@@ -668,10 +680,36 @@ bool EtcdClient::Node::HasExpiry() const {
...
+  }
+
+  ChooseNextServer();
+}
+
+
+void EtcdClient::ChooseNextServer() {
+  // Grab the next etcd server to try:
+  HostPortPair new_endpoint(move(etcds_.front()));
+  etcds_.pop_front();
```
pphaneuf: This juggling should be done with the lock held, I'd say.

cpp/util/etcd.cc
AlCutter: Done.

cpp/util/etcd.cc
```diff
@@ -688,14 +726,24 @@ void EtcdClient::FetchDone(RequestState* etcd_req, Task* task) {
...
 
   if (!task->status().ok()) {
-    // TODO(pphaneuf): If there is a connection problem, we should re-do the
-    // request on another node.
+    if (task->status().error_code() == util::error::UNAVAILABLE) {
+      // Seems etcd wasn't available; pick a new etcd server and retry
+      LOG(WARNING) << "Etcd fetch failed: " << task->status() << ", retrying "
+                   << "on next etcd server.";
+      ChooseNextServer();
+      etcd_req->SetHostPort(endpoint_);
```
pphaneuf: Accessing `endpoint_` without the lock here (assuming we want `endpoint_` to continue existing, which I don't think so at the moment).

cpp/util/etcd.cc
AlCutter: Checked range too.

cpp/util/etcd.h
AlCutter: I did this just to avoid ballooning the PR with all the call sites too.

cpp/util/etcd.h
AlCutter: SplitHosts was originally templated so you could have whichever container you wanted, and that was here for the `CHECK`s in the inlined code.

AlCutter: Removed.

cpp/util/etcd.h
AlCutter: Added

cpp/util/etcd.cc
AlCutter: Apparently not!

cpp/util/etcd.cc
```diff
@@ -688,14 +726,24 @@ void EtcdClient::FetchDone(RequestState* etcd_req, Task* task) {
...
+      fetcher_->Fetch(etcd_req->req_, &etcd_req->resp_,
```
pphaneuf: No delay, ever? It's a bit complicated to Do It Right (tm), needing things like knowing when you've gone through the whole list (so keeping track of what server you've tried)... Maybe this is sufficient, but that's the main reason I hadn't tackled this yet (didn't want to add an infinite busy loop here!).

cpp/server/ct-mirror.cc
AlCutter: done

cpp/server/ct-server.cc
AlCutter: done

cpp/util/etcd.cc
```diff
@@ -474,6 +476,16 @@ void EtcdClient::WatchInitialGetDone(WatchState* state, GetResponse* resp,
   // TODO(pphaneuf): Need better error handling here. Have to review
   // what the possible errors are, for now we'll just retry after a delay.
   if (!task->status().ok()) {
+    if (task->status().error_code() == util::error::UNAVAILABLE) {
```
pphaneuf: I think this should never happen, since this is called when the `Get` is done (which goes through `GenericRequest`, and thus, `FetchDone`).

I'm also seeing the retry just below what you've added, and considering that it's also duplicate, seems sketchy!

cpp/tools/clustertool_main.cc
AlCutter: done already

cpp/util/etcd.cc
AlCutter: Added

cpp/util/etcd.cc
```diff
@@ -688,14 +726,24 @@ void EtcdClient::FetchDone(RequestState* etcd_req, Task* task) {
   VLOG(2) << "EtcdClient::FetchDone: " << task->status();
 
   if (!task->status().ok()) {
-    // TODO(pphaneuf): If there is a connection problem, we should re-do the
-    // request on another node.
+    if (task->status().error_code() == util::error::UNAVAILABLE) {
+      // Seems etcd wasn't available; pick a new etcd server and retry
+      LOG(WARNING) << "Etcd fetch failed: " << task->status() << ", retrying "
+                   << "on next etcd server.";
+      ChooseNextServer();
```
pphaneuf: Why call this here? If you keep `endpoint_`, you could just call `UpdateEndpoint`, but I take `endpoint_` out entirely, and use `etcds_.front()` instead...

Note that ripping it out will properly find everything that's using it, as a side-effect, so would catch little details like `etcd_req->SetHostPort(endpoint_);` without having updated `etcds_`...

pphaneuf: Oh, and not only do you need to use `etcds_`, but when you get a redirect, you should add (or move) it to the front of the list.

pphaneuf:
I think I'll call it done for now. :wink:

cpp/util/etcd.cc
AlCutter: Added

cpp/util/etcd.cc
AlCutter: Added the lock.

AlCutter: Changed this all slightly.

cpp/util/etcd.cc
AlCutter: PRs accepted.

cpp/util/etcd.cc
AlCutter: I think this is a weird merge artifact, removed this block.

cpp/util/etcd.cc
AlCutter: Yeah I already did these as part of earlier stuff.

cpp/util/etcd.h
```diff
@@ -145,15 +150,24 @@ class EtcdClient {
...
                         util::Task* child_task);
 
+  HostPortPair endpoint() const;
+
   UrlFetcher* const fetcher_;
 
   mutable std::mutex lock_;
-  HostPortPair endpoint_;
+  std::list<HostPortPair> etcds_;
+
```
pphaneuf: Ding this extra line.

AlCutter: dinged

cpp/util/etcd.h
pphaneuf: SGTM

cpp/util/etcd.cc
```diff
@@ -668,10 +670,36 @@ bool EtcdClient::Node::HasExpiry() const {
...
   VLOG(1) << "EtcdClient: " << this;
+
+  for (const auto& e : etcds_) {
+    CHECK(!e.first.empty()) << "Empty host specified";
+    CHECK_GT(e.second, 0) << "Invalid port specified";
+  }
+}
+
+
+EtcdClient::HostPortPair EtcdClient::endpoint() const {
```
pphaneuf: That looks a **lot** like `GetEndpoint()`! I don't mind renaming it, but not duplicating. :wink:

cpp/util/etcd.cc
```diff
@@ -668,10 +670,36 @@ bool EtcdClient::Node::HasExpiry() const {
...
+  lock_guard<mutex> lock(lock_);
+  return etcds_.front();
+}
+
+
+void EtcdClient::ChooseNextServer() {
```
pphaneuf: This could return `etcds_.front()`, to save having to take the lock immediately after.

cpp/util/etcd.cc
```diff
@@ -746,17 +784,28 @@ void EtcdClient::FetchDone(RequestState* etcd_req, Task* task) {
...
 
 
 EtcdClient::HostPortPair EtcdClient::UpdateEndpoint(const string& host,
                                                     uint16_t port) {
   lock_guard<mutex> lock(lock_);
-  VLOG_IF(1, endpoint_.first != host || endpoint_.second != port)
-      << "new endpoint: " << host << ":" << port;
-  endpoint_ = make_pair(host, port);
-  return endpoint_;
+  HostPortPair new_endpoint(host, port);
```
pphaneuf: Could make that the parameter?

cpp/util/etcd.cc
```diff
@@ -746,17 +784,28 @@ void EtcdClient::FetchDone(RequestState* etcd_req, Task* task) {
...
+  auto it(find(etcds_.begin(), etcds_.end(), new_endpoint));
+  if (it == etcds_.end()) {
+    // TODO(alcutter): We don't really have a way of knowing when to remove
+    // etcd endpoints at the moment.  We should really be querying the etcd
+    // cluster for its members and using that list.
+    etcds_.emplace_front(move(new_endpoint));
+  } else {
+    etcds_.emplace_front(move(*it));
+    etcds_.erase(it);
```
pphaneuf: I think you could use `std::list::splice()` to do this, but this is all right too.

pphaneuf:
Done review, only one truly bad thing (the duplicated method), rest is "would be nice".

cpp/util/etcd.cc
AlCutter: Whoops, missed that.

cpp/util/etcd.cc
AlCutter: done

cpp/util/etcd.cc
AlCutter: done

cpp/util/etcd.cc
AlCutter: Nice, wasn't aware of that before, thanks.

pphaneuf:
Pretty classy, LGTM.

PR #845, opened 2015-07-24
Convert log IDs to URL-safe Base64 before using them in a file path

### LINKED ISSUE #771: [python] Monitor state file not found when log_id contains slash characters?
When running python/utilities/logobserver/logobserver.py against Certly (or DigiCert, etc) log servers, it would fail like:
```
WARNING:root:Monitor state file not found, assuming first run.
ERROR:root:get-sth from https://log.certly.io failed: [Errno 2] No such file or directory: u'/tmp/ct_monitor/zbUXm3/BwEb+6jETaj+PAC5hgvr4iW/syLL1tatgSQA='
ERROR:root:Update failed
ERROR:root:Failed to update data for https://log.certly.io: latest timestamp is Wed Dec 31 16:00:00 1969
```
I'm guessing that this is because the monitor state files are named by the `log_id` but some IDs have slashes in them?

RJPercival:
Fixes #771 by converting log IDs from standard Base64 to URL-safe Base64 before passing them to the state keeper, which uses them as part of a file path.

### DIFF
--- a/python/ct/client/prober.py
+++ b/python/ct/client/prober.py
@@ -42,7 +42,7 @@ class ProberThread(threading.Thread):
             hasher = merkle.TreeHasher()
             verifier = verify.LogVerifier(log.public_key_info,
                                           merkle.MerkleVerifier(hasher))
-            state_keeper = state_keeper_class(FLAGS.monitor_state_dir +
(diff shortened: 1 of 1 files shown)


pphaneuf:
LGTM

PR #847, opened 2015-07-24
Check for DER chain being empty in ct.client.reporter._scan_der_cert

### LINKED ISSUE #769: Are IndexErrors at ct/client/reporter.py", line 64, in _scan_der_cert expected?
I ran python/utilities/logobserver/logobserver.py against https://ct.googleapis.com/testtube and noticed some IndexErrors during the run, for example:

```
ERROR:root:Traceback (most recent call last):
  File "/<MY_FOLDER>/ct/client/reporter.py", line 64, in _scan_der_cert
    root = cert.Certificate(der_chain[-1], strict_der=False)
IndexError: list index out of range

ERROR:root:list index out of range
ERROR:root:Entry 50614 in batch <50000, 50999> raised an exception during scan
```
The "extra_data" for log entries 50614 was "AAAA"

Nothing serious but just wondering, is throwing IndexError the desired behavior here? Thanks!

RJPercival:
Fixes #769. Previously, an IndexError would have been raised if the DER chain was empty but now the certificate will be assumed to be a root certificate in this case.

### DIFF
--- a/python/ct/client/reporter.py
+++ b/python/ct/client/reporter.py
@@ -38,6 +38,7 @@ def _scan_der_cert(der_certs, checks):
         for log_index, der_cert, der_chain, entry_type in der_certs:
             current = log_index
             partial_result = []
+            certificate = None
             strict_failure = False
             try:
                 certificate = cert.Certificate(der_cert)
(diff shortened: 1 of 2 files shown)


pphaneuf:
LGTM

PR #851, opened 2015-07-27
Fixes bug in ct.client.db.cert_desc when local timezone is not UTC

### LINKED ISSUE #765: cert_desc_test.py is failing
Error message:

```
FAIL: test_from_cert (__main__.CertificateDescriptionTest)
  File "./ct/client/db/cert_desc_test.py", line 51, in test_from_cert
    CERT.not_before())
AssertionError: time.struct_time(tm_year=2013, tm_mon=5, tm_mday=22, tm_hour=23, tm_min=49, tm_sec=4, tm_wday=2, tm_yday=142, tm_isdst=0) != time.struct_time(tm_year=2013, tm_mon=5, tm_mday=22, tm_hour=15, tm_min=49, tm_sec=4, tm_wday=2, tm_yday=142, tm_isdst=0)

```

Seems to be a timezone-related issue, possibly similar to #627

RJPercival:
Fixes #765. This also modifies a test to use a non-UTC timezone, which then fails without the
included fix.

### DIFF
--- a/python/ct/cert_analysis/algorithm.py
+++ b/python/ct/cert_analysis/algorithm.py
@@ -1,3 +1,4 @@
+import calendar
 import time
 
 from ct.crypto import cert
@@ -19,7 +20,8 @@ class AlgorithmMismatch(AlgorithmObservation):
 
 class SHA1Observation(AlgorithmObservation):
     def _format_details(self):
-        return time.strftime("%Y-%m-%dT%H:%M:%S%Z", self.details)
+        # Times are in UTC, so use "UTC" instead of %Z (which gives local TZ)
+        return time.strftime("%Y-%m-%dT%H:%M:%SUTC", self.details)
 
 
 class RsaSHA1(SHA1Observation):
@@ -57,7 +59,7 @@ class CheckSignatureAlgorithmsMismatch(object):
 def check_sha1_after_2017(not_after, algorithm):
     invalid_after = time.strptime("2017-01-01T00:00:00UTC",
                                   "%Y-%m-%dT%H:%M:%S%Z")
-    if time.mktime(not_after) - time.mktime(invalid_after) >= 0:
+    if calendar.timegm(not_after) - calendar.timegm(invalid_after) >= 0:
         if algorithm == oid.SHA1_WITH_RSA_ENCRYPTION:
             return RsaSHA1(not_after)
         elif algorithm == oid.ID_DSA_WITH_SHA1:
--- a/python/ct/client/db/cert_desc.py
+++ b/python/ct/client/db/cert_desc.py
@@ -1,6 +1,6 @@
 import re
 import hashlib
-import time
+import calendar
 from ct.crypto import cert
 from ct.crypto.asn1 import x509_common
 from ct.proto import certificate_pb2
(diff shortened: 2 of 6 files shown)


python/ct/client/db/cert_desc.py
```diff
@@ -1,6 +1,6 @@
 import re
 import hashlib
-import time
+import calendar
```
phad: Move up to sort imports (hashlib too)

python/ct/client/db/cert_desc_test.py
```diff
@@ -13,81 +14,86 @@
...
-                             source.signature_algorithm()["algorithm"].long_name)
-            self.assertEqual(proto.tbs_signature.algorithm_id,
-                             proto.cert_signature.algorithm_id)
-
-            if source.signature()["parameters"]:
-                self.assertEqual(proto.tbs_signature.parameters,
-                                 source.signature()["parameters"])
-            else:
-                self.assertFalse(proto.tbs_signature.HasField('parameters'))
+        # Test in a non-UTC timezone, to detect timezone issues
```
phad: Shouldn't the tests run in at least one other timezone?

python/ct/cert_analysis/algorithm.py
```diff
@@ -57,7 +59,7 @@ def check(certificate):
 def check_sha1_after_2017(not_after, algorithm):
     invalid_after = time.strptime("2017-01-01T00:00:00UTC",
                                   "%Y-%m-%dT%H:%M:%S%Z")
```
eranmes: Shouldn't you fix here as well, so this check would not use the local timezone?

python/ct/client/db/cert_desc_test.py
```diff
@@ -13,81 +14,86 @@
...
+        with time_utils.timezone("America/Los_Angeles"):
+            for test_case in [(CERT, False), (DSA_SHA256_CERT, False), (CA_CERT, True)]:
```
eranmes: This test method is getting rather substantial. It has a 'with' clause, a for clause (for 3 different cases), a couple of if/else statements and quite a bit of code in between. How about breaking it up?

python/ct/test/time_utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def get_timezone_environ():
+    return os.environ.get("TZ", '')
+
+def set_timezone_environ(new_timezone):
+    os.environ["TZ"] = new_timezone
+    time.tzset()
+
+@contextlib.contextmanager
```
eranmes: Nit: Are you going to re-use this context manager somewhere else?
If not, I how about folding timezone setting into the setUp / tearDown of the only test class that currently uses it?

RJPercival: As per @phad's comment, I'll be making the tests try at least two different timezones, so putting this directly into setUp/tearDown won't work. It seems worthwhile keeping this code in a utility module as this lowers the barrier to re-use and thereby encourages others to use it in their tests, should they also write some code that involves timestamps.

phad: Not sure what timezone @linshunghuang is in, but he's reported TZ-related bugs in the cert_desc.py tests a couple of times, so it'd be good to add his TZ to the test case.

python/ct/client/db/cert_desc_test.py
```diff
@@ -6,88 +6,132 @@
...
+        with time_utils.timezone("UTC"):
+            self.assert_description_matches_source(CERT, False)
+
+        # Test in a non-UTC timezone, to detect timezone issues
+        with time_utils.timezone("America/Los_Angeles"):
+            self.assert_description_matches_source(DSA_SHA256_CERT, False)
+
+    def test_from_cert_with_ca_cert(self):
+        with time_utils.timezone("UTC"):
+            self.assert_description_matches_source(CERT, False)
```
eranmes: Why are you using CERT here but CA_CERT below?

RJPercival: Oops, copy-and-paste error - fixed.

PR #852, opened 2015-07-27
Catch manifests up with latest GCE/Kubernetes requirements.

AlCutter:
+ Add a `metadata.name` field
+ Shorten port names to 15 chars or less.


pphaneuf:
LGTM

PR #853, opened 2015-07-27
Connection pool fixes

AlCutter:


cpp/net/connection_pool.cc
```diff
@@ -205,7 +234,15 @@ evhtp_res ConnectionPool::Connection::ConnectionErrorHook(
     // If someone hasn't already modified the default status, set it to a
     // generic "something went wrong" value here:
     if (conn->request->status == 200) {
-      conn->request->status = kUnknownErrorStatus;
+      VLOG(1) << "error flag (0x" << std::hex << static_cast<int>(flags)
+              << "):" << ErrorFlagDescription(flags);
```
pphaneuf: The `std::hex` only affects the thing immediately after it, right?

cpp/util/etcd.cc
```diff
@@ -573,9 +573,8 @@ void EtcdClient::WatchRequestDone(WatchState* state, GetResponse* get_resp,
     return;
   }
 
-  // This is probably due to a timeout, just retry.
-  if (!child_task->status().ok()) {
-    LOG_EVERY_N(INFO, 10) << "Watch request failed: " << child_task->status();
+  if (child_task->status().error_code() == util::error::DEADLINE_EXCEEDED) {
+    VLOG(1) << "Watch request timedout: " << child_task->status();
```
pphaneuf: Shouldn't we restart other kinds of errors too? Or can *no* other error get here, somehow?

cpp/net/connection_pool.cc
AlCutter: I think it affects any scalar values sent to the stream after the manipulator, it won't affect the string description though.

cpp/util/etcd.cc
AlCutter: I restricted this to timeouts-only based on the comment that was there, but we could extend that to other errors if that was the right thing to do.  I don't really have a strong opinion either way other than fear of hiding a permanently borked etcd from the higher layers.

This change was mainly intended to clear up the misleading `UNAVAILABLE` coming from the read timeout on the hanging get.

pphaneuf: I think for watches, we want to keep them trying forever, unless there was some sort of error that could be truly final?

For example, all etcd might go offline for a bit, nothing much is going to happen, but we should keep trying, and resume notifying once they come back, I'd say.

AlCutter: Fair enough, I've squashed that in.

pphaneuf:
LGTM, pending Travis.

PR #854, opened 2015-07-29
Converts some PEM test data to Unix line endings

RJPercival:
This makes them consistent with the other PEM files.


pphaneuf:
LGTM

PR #856, opened 2015-07-30
Use registered field number for Protobuf TLS extension fields

RJPercival:
The extension fields have been registered as follows:
Project:  Certificate Transparency
Contact:  Paul Hadfield (hadfieldp@google.com)
Web site: https://github.com/google/certificate-transparency
Extensions: 1023 (all types)

### DIFF
--- a/python/ct/proto/tls_options.proto
+++ b/python/ct/proto/tls_options.proto
@@ -4,7 +4,6 @@ import "google/protobuf/descriptor.proto";
 
 package ct;
 
-
 // TLS field options specify the wire format for parsing TLS wire messages
 // to protocol buffers, using the custom decoder in tls_message.py.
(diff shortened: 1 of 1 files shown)


benlaurie:
Curious why its OK to use the same number twice?

phad:
So, I originally requested two distinct extension IDs, thinking that they
should differ.
Here's what the registry told me:
---
Hi Paul,

Sorry for the late response. I have added an extension field number entry
for you:

--------------------------------
Project:  Certificate Transparency
Contact:  Paul Hadfield
Web site: https://github.com/google/certificate-transparency
Extensions: 1023 (all types)
--------------------------------

You could use 1023 for both extensions because they are extending different
option types.

PR #857, opened 2015-07-30
Add an outline of a Gossip server in Go.

AlCutter:
Outline of the HTTPS server functionality described in http://datatracker.ietf.org/doc/draft-linus-trans-gossip-ct/?include_text=1 (version 02)

Missing lots of functionality; no validation of anything AT ALL, no Trusted Auditor support, etc. etc., but it's a start.

### DIFF
--- a/go/gossip/handler.go
+++ b/go/gossip/handler.go
@@ -1,0 +1,92 @@
+package gossip
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"net/http"
+)
+
+var defaultNumPollinationsToReturn = flag.Int("default_num_pollinations_to_return", 10,
+	"Number of randomly selected STH pollination entries to return for sth-pollination requests.")
+
+// Handler for the gossip HTTP requests.
+type Handler struct {
+	storage *Storage
+}
+
+func writeWrongMethodResponse(rw *http.ResponseWriter, allowed string) {
+	(*rw).Header().Add("Allow", allowed)
+	(*rw).WriteHeader(http.StatusMethodNotAllowed)
+}
+
+func writeErrorResponse(rw *http.ResponseWriter, status int, body string) {
+	(*rw).WriteHeader(status)
+	(*rw).Write([]byte(body))
+}
+
+// HandleSCTFeedback handles requests POSTed to .../sct-feedback.
+// It attempts to store the provided SCT Feedback
+func (h *Handler) HandleSCTFeedback(rw http.ResponseWriter, req *http.Request) {
+	if req.Method != "POST" {
+		writeWrongMethodResponse(&rw, "POST")
+		return
+	}
+
+	decoder := json.NewDecoder(req.Body)
+	var feedback SCTFeedback
+	if err := decoder.Decode(&feedback); err != nil {
+		writeErrorResponse(&rw, http.StatusBadRequest, fmt.Sprintf("Invalid SCT Feedback received: %v", err))
+		return
+	}
+
+	// TODO(alcutter): 5.1.1 Validate leaf chains up to a trusted root
+	// TODO(alcutter): 5.1.1/2 Verify each SCT is valid and from a known log, discard those which aren't
+	// TODO(alcutter): 5.1.1/3 Discard leaves for domains other than ours.
+	if err := h.storage.AddSCTFeedback(feedback); err != nil {
+		writeErrorResponse(&rw, http.StatusInternalServerError, fmt.Sprintf("Unable to store feedback: %v", err))
+		return
+	}
+	rw.WriteHeader(http.StatusOK)
+}
+
+// HandleSTHPollination handles requests POSTed to .../sth-pollination.
+// It attempts to store the provided pollination info, and returns a random set of
+// pollination data from the last 14 days (i.e. "fresh" by the definition of the gossip RFC.)
+func (h *Handler) HandleSTHPollination(rw http.ResponseWriter, req *http.Request) {
+	if req.Method != "POST" {
+		writeWrongMethodResponse(&rw, "POST")
+		return
+	}
+
+	decoder := json.NewDecoder(req.Body)
+	var p STHPollination
+	if err := decoder.Decode(&p); err != nil {
+		writeErrorResponse(&rw, http.StatusBadRequest, fmt.Sprintf("Invalid STH Pollination received: %v", err))
+		return
+	}
+
+	err := h.storage.AddSTHPollination(p)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


go/gossip/handler_test.go
```diff
@@ -0,0 +1,309 @@
...
+        ]
+      }`
+)
+
+func CreateAndOpenStorage() *Storage {
+	dir, err := ioutil.TempDir("", "handlertest")
+	if err != nil {
+		log.Fatalf("Failed to get temporary dir for test: %v", err)
+	}
+	*dbName = dir + "/gossip.db"
```
benlaurie: Eww. Why is this not local to Storage?

go/gossip/storage.go
```diff
@@ -0,0 +1,339 @@
...
+		return errors.New("attempting to call Open() on an already Open()'d Storage")
+	}
+	s.db, err = sql.Open("sqlite3", *dbName)
+	if err != nil {
+		return err
+	}
+	if _, err := s.db.Exec(schema); err != nil {
+		return err
+	}
+	if s.insertChain, err = s.db.Prepare(insertChain); err != nil {
```
benlaurie: Man. Seems like their ought to be an easier way than all this repetition?

go/gossip/storage.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	return id, nil
+}
+
+// insertThingOrSelectID will attempt to execute the insert Statement (under transaction tx), if that fails due to
+// a unique primary key constraint, it will look up that primary key by executing the getID Statement.
+// Returns the ID associated with persistent thing, or an error describing the failure.
+func insertThingOrSelectID(tx *sql.Tx, insert *sql.Stmt, getID *sql.Stmt, thing interface{}) (int64, error) {
+	txInsert := tx.Stmt(insert)
+	txGetID := tx.Stmt(getID)
```
benlaurie: O.M.G.

You should use txInsert.LastInsertId().

Don't believe it needs to be in a transaction, either.

go/gossip/storage.go
```diff
@@ -0,0 +1,339 @@
...
+	r, err := txInsert.Exec(thing)
+	if err != nil {
+		switch err.(type) {
+		case sqlite3.Error:
+			if err.(sqlite3.Error).Code == sqlite3.ErrConstraint {
```
benlaurie: Can you explain this notation to me? :-)

go/gossip/storage.go
benlaurie: err. I meant r.LastInsertId(), below.

benlaurie: Oh, I see I am completely confused. :-) Never mind.

go/gossip/storage.go
```diff
@@ -0,0 +1,339 @@
...
+// AddSCTFeedback stores the passed in feedback object.
+func (s *Storage) AddSCTFeedback(feedback SCTFeedback) error {
+	tx, err := s.db.Begin()
+	if err != nil {
+		return err
+	}
+	for _, f := range feedback.Feedback {
+		chainID, err := s.addChainIfNotExists(tx, f.X509Chain)
+		if err != nil {
+			return err
```
benlaurie: Leaves transaction open.

go/gossip/storage.go
```diff
@@ -0,0 +1,339 @@
...
+		default:
+			return err
+		}
+	}
+	return nil
+}
+
+// AddSCTFeedback stores the passed in feedback object.
+func (s *Storage) AddSCTFeedback(feedback SCTFeedback) error {
+	tx, err := s.db.Begin()
```
benlaurie: Why does this need to be in a transaction?

go/gossip/storage.go
benlaurie: One obvious possibility...

Most of these SQL strings could be synthesized on the fly (e.g "SELECT COUNT(*) FROM <table>"), you could keep a map from strings to prepared statements and prepare them on the fly.

go/gossip/handler_test.go
AlCutter: Fair point, is now.

go/gossip/storage.go
AlCutter: I guess it doesn't really need to be at the moment, but if we ever allowed the server to ditch old feedback/pollen then we'd potentially have a race here.

go/gossip/storage.go
AlCutter: I assumed that it'd `Rollback()` automatically when the `Tx` went out of scope, but you're right - it doesn't.
Fixed that up with some `defer` trickery.

go/gossip/storage.go
AlCutter: Thing is with that approach is that you lose the compile-time checking that you didn't misspell the key somewhere.

I've tried something else, have a look and see what you think.

go/gossip/storage.go
AlCutter: This is a "type assertion" (kinda like a `dynamic_cast` I guess; gives you back a reference to `err` which has the `sqlite3.Error` type, or `nil` if you guessed wrong), but I've realised that I can have it for free by capturing the value of the `err.(type)` above.

go/gossip/handler.go
```diff
@@ -0,0 +1,90 @@
...
+		rw.Header().Add("Allow", "POST")
+		rw.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	decoder := json.NewDecoder(req.Body)
+	var feedback SCTFeedback
+	if err := decoder.Decode(&feedback); err != nil {
+		rw.WriteHeader(http.StatusBadRequest)
+		rw.Write([]byte(fmt.Sprintf("Invalid SCT Feedback received: %v", err)))
```
benlaurie: This stanza occurs a lot ... refactor?

AlCutter: Good call, done.

go/gossip/storage.go
```diff
@@ -0,0 +1,354 @@
...
+			sctID, err := s.addSCTIfNotExists(tx, sct)
+			if err != nil {
+				return err
+			}
+			if err = s.addSCTFeedbackIfNotExists(tx, chainID, sctID); err != nil {
+				return err
+			}
+		}
+	}
+	return tx.Commit()
```
benlaurie: First assert err == nil? Or put the commit in the defer also?

benlaurie:
Other than that comment...

go/gossip/storage.go
AlCutter: Moved this (and the same pattern in the other method which used a TX) to the `defer` block.

PR #859, opened 2015-07-31
Add Draft RFC extension validation logic to Cert class

Martin2112:
Starting work to implement the draft RFC features. Adds new extensions and
validation logic to Cert class to check compliance with RFC sections
2.3.3 and 2.3.4. Adds various test cases and the script that generates the
test certs.


Martin2112:
I signed it!

cpp/log/cert.cc
```diff
@@ -2,6 +2,7 @@
 #include "log/cert.h"
 
 #include <glog/logging.h>
+#include <util/util.h>
```
AlCutter: We try to keep the include ordering alphabetically ordered (and grouped according the Google C++ Style Guide.)

cpp/log/cert.cc
```diff
@@ -536,9 +555,10 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
   // "extension not found, found more than once or corrupt".
   Cert::Status status = HasExtension(extension_nid);
   if (status != TRUE)
```
phad: You might as well brace this 'if'

cpp/log/cert.cc
```diff
@@ -92,6 +93,24 @@ Cert::Status Cert::LoadFromDerString(const std::string& der_string) {
   return Cert::TRUE;
 }
 
+Cert::Status Cert::LoadFromDerBio(BIO *bio_in) {
+  if (x509_ != NULL) {
```
AlCutter: prefer `nullptr` over `NULL` if possible (C++11 spoken here ;))

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
     return FALSE;
   }
 
   return TRUE;
 }
 
+bool Cert::IsRedactedHost(const std::string& hostname) {
+  // Split the hostname on '.' characters
+  std::vector<string> tokens = util::split(hostname, '.');
```
phad: tokens could be const

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+        }
+      }
+    }
+
+    sk_GENERAL_NAME_free(subject_alt_names);
+  }
+
+  // The next stage of validation is that if the subject name CN exists it
+  // must match the first DNS id and have the same labels redacted
+  // TODO(mhs): Confirm it's valid to not have a CN.
```
phad: Who's mhs? :)

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  std::vector<std::string> dns_alt_names;
+  int redacted_name_count = 0;
+
+  // First. Check all the Subject Alt Name extension records. Any that are of
+  // type DNS must pass validation if they are attempting to redact labels
+  STACK_OF(GENERAL_NAME) * subject_alt_names = (STACK_OF(GENERAL_NAME) *)
+      X509_get_ext_d2i(x509_, NID_subject_alt_name, NULL, NULL);
+
+  if (subject_alt_names != NULL) {
+    int subject_alt_name_count = sk_GENERAL_NAME_num(subject_alt_names);
```
phad: this can be const

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      return true;
+    }
+  }
+
+  return false;
+}
+
+bool Cert::IsValidRedactedHost(const std::string& hostname) {
+  // Split the hostname on '.' characters
+  std::vector<string> tokens = util::split(hostname, '.');
```
phad: const again

cpp/log/cert.cc
```diff
@@ -92,6 +93,24 @@ Cert::Status Cert::LoadFromDerString(const std::string& der_string) {
   return Cert::TRUE;
 }
 
+Cert::Status Cert::LoadFromDerBio(BIO *bio_in) {
+  if (x509_ != NULL) {
+    X509_free(x509_);
```
AlCutter: Hmm, not on you, but looks like we should convert this class to use a `unique_ptr` with custom deallocator for x509_, could you pop a TODO in there against me?

cpp/log/cert.cc
```diff
@@ -92,6 +93,24 @@ Cert::Status Cert::LoadFromDerString(const std::string& der_string) {
...
+    x509_ = NULL;
+  }
+
+  x509_ = d2i_X509_bio(bio_in, &x509_);
+
+  if (x509_ == NULL) {
+    // At this point most likely the input was just corrupt. There are few
```
AlCutter: /are few/are a few/ ?

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // must match the first DNS id and have the same labels redacted
+  // TODO(mhs): Confirm it's valid to not have a CN.
+  X509_NAME* name = X509_get_subject_name(x509_);
+
+  if (name == NULL) {
+    LOG(ERROR) << "Missing X509 subject name";
+    return Cert::ERROR;
+  }
+
+  int name_pos = -1;
```
phad: wouldn't the X509_NAME_get_index_by_NID call return a -ve error code?

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+  if (name_pos >= 0) {
+    X509_NAME_ENTRY* name_entry = X509_NAME_get_entry(name, name_pos);
+
+    if (name_entry != NULL) {
+      ASN1_STRING *subject_name_asn1 = X509_NAME_ENTRY_get_data(name_entry);
+
+      if (subject_name_asn1 == NULL) {
+        LOG(WARNING) << "Missing subject name";
+        // TODO(mhs_: Check this is correct behaviour. Is it OK to not have
```
phad: _ -> ), mhs

cpp/log/cert.cc
AlCutter: Can you wrap this in `{}` - we use them even for 1 line blocks (goto: fail :)

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  }
+
+  // If it's not a CA cert or there is no name constraint extension then we
+  // don't need to apply the rules any further
+  if (HasBasicConstraintCATrue() != Cert::TRUE
+        || HasExtension(NID_name_constraints) == FALSE) {
+    return Cert::TRUE;
+  }
+
+  // So there now must be a CT exension and the name constraint must not be
```
phad: typo: extension

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
       LOG_OPENSSL_ERRORS(WARNING);
     }
+
     return FALSE;
   }
 
   return TRUE;
 }
 
+bool Cert::IsRedactedHost(const std::string& hostname) {
```
AlCutter: This doesn't need to be a member function I think, can you make it free standing?

AlCutter: Don't need the `std::` on string (there should be a `using std::string` above, but if not then feel free to add one.)

cpp/log/cert.cc
AlCutter: same for `std::` here, also you may wish to initialise via c'tor here:
```C++
vector<string> tokens(util::split(hostname, '.'));
```

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // Split the hostname on '.' characters
+  std::vector<string> tokens = util::split(hostname, '.');
+
+  for (string str : tokens) {
```
AlCutter: `const string& str : tokens` otherwise you're making copies

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    if (str == "?") {
```
AlCutter: Could this all be just something like:
`return hostname.find("?") != npos`
or
`return hostname.find(".?") != npos || hostname.find("?.") != npos || hostname == "?";`

Admittedly it's O(2N), but less code (or are you doing something slightly different to that?).

cpp/log/cert.cc
AlCutter: same `std::` and c'tor comments here too

AlCutter: And also doesn't need to be a member function

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  return false;
+}
+
+bool Cert::IsValidRedactedHost(const std::string& hostname) {
+  // Split the hostname on '.' characters
+  std::vector<string> tokens = util::split(hostname, '.');
+
+  // Enforces the following rules: '?' must be to left of non redactions
+  // If first label is '*' then treat it as if it was a redaction
+  bool can_redact = true;
```
AlCutter: nit: c'tor init and elsewhere (I'll stop calling these out now)

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+  for (string str : tokens) {
+    if (str == "?") {
+      return true;
+    }
+  }
+
+  return false;
+}
+
```
AlCutter: trying to standardise on two lines between top-level items in files

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  return true;
+}
+
+Cert::Status Cert::IsValidWildcardRedaction() const {
+  if (!IsLoaded()) {
+    LOG(ERROR) << "Cert not loaded";
+    return Cert::ERROR;
+  }
+
+  std::string common_name;
```
AlCutter: zap the `std::`s (I'll stop picking these out now)

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  std::vector<std::string> dns_alt_names;
+  int redacted_name_count = 0;
+
+  // First. Check all the Subject Alt Name extension records. Any that are of
+  // type DNS must pass validation if they are attempting to redact labels
+  STACK_OF(GENERAL_NAME) * subject_alt_names = (STACK_OF(GENERAL_NAME) *)
```
AlCutter: Attach * to `STACK_OF(GENERAL_NAME)` (we've standardised on pointer-is-the-type)

AlCutter: You *may* be able to pull the `unique_ptr` with a custom deallocation functor trick here to easy the memory management effort, haven't tried it with a STACK_OF yet though, so YMMV.

cpp/log/cert.cc
AlCutter: `int name_pos(X509_NAME_get_index_by_NID(name, NID_commonName, name_pos));`

cpp/log/cert.cc
Martin2112: I was trying to match exactly what the spec says as it refers to labels in the name. The RFC doesn't specify how to treat anything other than labels which are '?' e.g. "x?.example.com" doesn't fall into the definition of a redaction as I read it.

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    if (name_entry != NULL) {
+      ASN1_STRING *subject_name_asn1 = X509_NAME_ENTRY_get_data(name_entry);
+
+      if (subject_name_asn1 == NULL) {
+        LOG(WARNING) << "Missing subject name";
+        // TODO(mhs_: Check this is correct behaviour. Is it OK to not have
+        // a subject?
+      } else {
+        common_name = std::string(
+            reinterpret_cast<char*>(ASN1_STRING_data(subject_name_asn1)),
```
AlCutter: This ASN1ToStringAndCheckForNull block crops up a few times, possible to refactor it out into a free function?

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      return Cert::FALSE;
+    }
+
+    // All the integers in the sequence must be positive, check the sign
+    // after conversion to BIGNUM
+    for (int i = 0; i < num_integers; ++i) {
+      ASN1_INTEGER* redacted_labels = sk_ASN1_INTEGER_value(integers, i);
+      BIGNUM *value = ASN1_INTEGER_to_BN(redacted_labels, NULL);
+
+      int neg = value -> neg;
```
AlCutter: bool?

AlCutter: also can be const

cpp/log/cert.cc
```diff
@@ -546,12 +566,314 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    return Cert::FALSE;
+  }
+
+  // Unpack the extension contents, which should be SEQUENCE OF INTEGER
+  STACK_OF(ASN1_INTEGER)* integers = (STACK_OF(ASN1_INTEGER)*)
+      ASN1_seq_unpack_ASN1_INTEGER(exty->value->data, exty->value->length,
+                                   d2i_ASN1_INTEGER, ASN1_INTEGER_free);
+
+  if (integers != NULL) {
+    int num_integers = sk_ASN1_INTEGER_num(integers);
```
AlCutter: const all the things

cpp/log/cert_test.cc
```diff
@@ -325,6 +435,159 @@ TEST_F(CertTest, SignatureAlgorithmMatches) {
   EXPECT_EQ(Cert::FALSE, mismatched_algs.IsSignedBy(issuer));
 }
 
+TEST_F(CertTest, TestIsRedactedHost) {
+  EXPECT_FALSE(Cert::IsRedactedHost(""));
```
AlCutter: How about "?" (here and in `TestIsValidRedactedHost()`) ? :)

cpp/log/ct_extensions.cc
```diff
@@ -4,6 +4,7 @@
 #include <openssl/asn1.h>
 #include <openssl/objects.h>
 #include <openssl/x509v3.h>
+#include <openssl/asn1t.h>
```
AlCutter: ordering

cpp/log/ct_extensions.cc
```diff
@@ -12,6 +13,8 @@ int NID_ctSignedCertificateTimestampList = 0;
 int NID_ctEmbeddedSignedCertificateTimestampList = 0;
 int NID_ctPoison = 0;
 int NID_ctPrecertificateSigning = 0;
+int NID_ctPrecertificateRedactedLabelCount = 0;
```
AlCutter: ordering

cpp/log/ct_extensions.h
```diff
@@ -1,17 +1,33 @@
...
 namespace cert_trans {
 
 // One-time initializer for loading CT-specific certificate extensions.
 void LoadCtExtensions();
 
+// Defines structure for redacted label count to hold a stack of integers
+typedef struct RedactedLabelCount_st {
+  STACK_OF(ASN1_INTEGER)* redacted_labels;
+} REDACTED_LABEL_COUNT;
+//DECLARE_ASN1_ITEM(REDACTED_LABEL_COUNT)
```
AlCutter: remove the commented out line?

cpp/log/cert.cc
AlCutter: Oke doke.

cpp/log/cert.cc
Martin2112: There's existing usages of NULL in this file, are we converting them over?

AlCutter: at least new code should prefer `nullptr`, if you'd like to convert the others too then bonus points will be awarded! (but maybe do that in a followup PR if possible)

cpp/log/cert.cc
Martin2112: Do we use GitHub id or just plain TODO?

cpp/log/cert.cc
pphaneuf: Spaces around `->`? :smiley:

cpp/log/cert.cc
Martin2112: I think it does but I'm returning a different status for the two cases. Arguably they're the same.

pphaneuf:
For the CLA, the problem is that the commits have `Martin2112 <Martin2112@users.noreply.github.com>` as the author... You need to use an email registered with the CLA (and if you use an `@google.com` email, the bot will figure it out automatically, no need to sign the CLA).

It's possible to amend the commit authors, while they're still in review. :wink:

cpp/log/cert.cc
pphaneuf: Not to mention that this should just be `if (x509_)` anyway, neatly dodging the whole issue. :wink:

cpp/log/cert.cc
Martin2112: Hoping to defer that to a followup as I'm over my segfault quota this week.

cpp/log/ct_extensions.cc
Martin2112: Done. I was trying to separate the V2 OIDs but the comment also marks them.

Martin2112:
Can you let me know how to change the email address on this? I can't see
where to do this.

Thanks,

Martin

benlaurie:
git commit --amend --author=xxx maybe?

cpp/log/cert.cc
```diff
@@ -36,10 +37,29 @@ static int X509_get_signature_nid(const X509* x) {
 
 namespace cert_trans {
 
+// Convert string from ASN1 and check it doesn't contain nul characters
+string ASN1ToStringAndCheckForNulls(ASN1_STRING *asn1_string,
+                                    string tag,
```
AlCutter: const string&

cpp/log/cert.cc
```diff
@@ -36,10 +37,29 @@ static int X509_get_signature_nid(const X509* x) {
...
+                                    Cert::Status& status) {
```
AlCutter: Pass by pointer if you're going to modify the param.

cpp/log/cert.cc
```diff
@@ -36,10 +37,29 @@ static int X509_get_signature_nid(const X509* x) {
...
+  string cpp_string = string(
```
AlCutter:
```C++
const string cpp_string(reinter...., ASN1_STRING_len...);
```

cpp/log/cert.cc
```diff
@@ -36,10 +37,29 @@ static int X509_get_signature_nid(const X509* x) {
...
+  }
+
+  return cpp_string;
+}
+
 Cert::Cert(X509* x509) : x509_(x509) {
 }
 
-Cert::Cert(const std::string& pem_string) : x509_(NULL) {
+Cert::Cert(const string& pem_string) : x509_(NULL) {
```
AlCutter: `nullptr`

cpp/log/cert.cc
```diff
@@ -76,7 +96,7 @@ Cert* Cert::Clone() const {
   return clone;
 }
 
-Cert::Status Cert::LoadFromDerString(const std::string& der_string) {
+Cert::Status Cert::LoadFromDerString(const string& der_string) {
   if (x509_ != NULL) {
```
AlCutter:
```C++
if (x509_) {
...
```

cpp/log/cert.cc
```diff
@@ -92,6 +112,26 @@ Cert::Status Cert::LoadFromDerString(const std::string& der_string) {
   return Cert::TRUE;
 }
 
+Cert::Status Cert::LoadFromDerBio(BIO *bio_in) {
+  if (x509_) {
```
AlCutter:
```C++
CHECK_NOTNULL(bio_in);
```

cpp/log/cert.cc
```diff
@@ -92,6 +112,26 @@ Cert::Status Cert::LoadFromDerString(const std::string& der_string) {
...
+  }
+
+  x509_ = d2i_X509_bio(bio_in, &x509_);
+
+  if (!x509_) {
+    // At this point most likely the input was just corrupt. There are few
+    // real errors that may have happened (a malloc failure is one) and it is
+    // virtually impossible to fish them out.
+    LOG(WARNING) << "Input is not a valid encoded certificate";
+    LOG_OPENSSL_ERRORS(WARNING);
```
AlCutter:
```C++
return FALSE; 
```
?

Add a test which would've caught trying to load from bogus data?

Martin2112: This is part of the next change for CMS unwrapping, it's not used yet which is why this got through.

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
       LOG_OPENSSL_ERRORS(WARNING);
     }
+
     return FALSE;
   }
 
   return TRUE;
 }
 
+bool IsRedactedHost(const string& hostname) {
```
AlCutter: Pop these into an anonymous namespace (unless you need access to them from other places), that way the compiler will warn if they're unused at any point.

cpp/log/cert.cc
AlCutter: ping

cpp/log/cert.cc
AlCutter: ping

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  string common_name;
+  vector<string> dns_alt_names;
+  int redacted_name_count = 0;
+
+  // First. Check all the Subject Alt Name extension records. Any that are of
+  // type DNS must pass validation if they are attempting to redact labels
+  STACK_OF(GENERAL_NAME)* subject_alt_names = (STACK_OF(GENERAL_NAME)*)
+      X509_get_ext_d2i(x509_, NID_subject_alt_name, NULL, NULL);
+
+  if (subject_alt_names != nullptr) {
```
AlCutter: [channeling @pphaneuf :]
```C++
if (subject_alt_names) {
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    const int subject_alt_name_count = sk_GENERAL_NAME_num(subject_alt_names);
+
+    for (int i = 0; i < subject_alt_name_count; ++i) {
+      GENERAL_NAME* name = sk_GENERAL_NAME_value(subject_alt_names, i);
+
+      Cert::Status name_status;
+
+      if (name -> type == GEN_DNS) {
```
AlCutter:
```C++
name->type
```

cpp/log/cert.cc
Martin2112: Will that affect testing?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // First. Check all the Subject Alt Name extension records. Any that are of
+  // type DNS must pass validation if they are attempting to redact labels
+  STACK_OF(GENERAL_NAME)* subject_alt_names = (STACK_OF(GENERAL_NAME)*)
+      X509_get_ext_d2i(x509_, NID_subject_alt_name, NULL, NULL);
+
+  if (subject_alt_names != nullptr) {
+    const int subject_alt_name_count = sk_GENERAL_NAME_num(subject_alt_names);
+
+    for (int i = 0; i < subject_alt_name_count; ++i) {
+      GENERAL_NAME* name = sk_GENERAL_NAME_value(subject_alt_names, i);
```
AlCutter: could this potentially return nullptr?
Also, const all the things:
```C++
GENERAL_NAME* cost name(....);
```
(Because OpenSSL it's probably going to break/require const_cast everywhere if you make the pointee const, but give it a go...)

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+      Cert::Status name_status;
+
+      if (name -> type == GEN_DNS) {
+        const string dns_name = ASN1ToStringAndCheckForNulls(name->d.dNSName,
```
AlCutter: Would it make sense to pass in `&dns_name` and return the status?
If only this were a `util::Status`, then you could use `util::StatusOr<>`.
One day, one day...

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    return Cert::ERROR;
+  }
+
+  string common_name;
+  vector<string> dns_alt_names;
+  int redacted_name_count = 0;
+
+  // First. Check all the Subject Alt Name extension records. Any that are of
+  // type DNS must pass validation if they are attempting to redact labels
+  STACK_OF(GENERAL_NAME)* subject_alt_names = (STACK_OF(GENERAL_NAME)*)
```
AlCutter: unique_ptr w/ custom deallocator?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    LOG(ERROR) << "Missing X509 subject name";
+    return Cert::ERROR;
+  }
+
+  int name_pos = -1;
+
+  name_pos = X509_NAME_get_index_by_NID(name, NID_commonName, name_pos);
+
+  if (name_pos >= 0) {
+    X509_NAME_ENTRY* name_entry = X509_NAME_get_entry(name, name_pos);
```
AlCutter:
```C++
X509_NAME_ENTRY* const name_entry(...)
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+    if (name_entry != NULL) {
```
AlCutter:
```C++
if (name_entry) {
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      ASN1_STRING *subject_name_asn1 = X509_NAME_ENTRY_get_data(name_entry);
```
AlCutter:
```C++
ASN1_STRING* const subject...(...);
```

cpp/log/cert.cc
Martin2112: These are all macros, will try to add const. As we checked the stack wasn't nullptr above there would have to be a null data item in the data we get from OpenSSL. Not sure if that's possible

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+      if (subject_name_asn1 == nullptr) {
```
AlCutter:
```C++
if (!subject_name_asn1) {
...
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+        // '*' is only valid at the left
+        return false;
+      }
+    }
+  }
+
+  return true;
+}
+
+Cert::Status Cert::IsValidWildcardRedaction() const {
```
AlCutter: This method ~300 lines long, could it be split up further?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+    sk_GENERAL_NAME_free(subject_alt_names);
+  }
+
+  // The next stage of validation is that if the subject name CN exists it
+  // must match the first DNS id and have the same labels redacted
+  // TODO: Confirm it's valid to not have a CN.
+  X509_NAME* name = X509_get_subject_name(x509_);
+
+  if (name == nullptr) {
```
AlCutter:
```C++
if (!name) {
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // that the required extensions are present and specified correctly if
+  // we found any redacted names. First though if nothing is redacted
+  // then the rest of the rules need not be applied
+  if (redacted_name_count == 0 && !IsRedactedHost(common_name)) {
+    return Cert::TRUE;
+  }
+
+  // If we reach here then the RFC says the CT redaction count extension
+  // MUST BE present.
+  X509_EXTENSION* exty;
```
AlCutter: can you constify the ptr?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  if (GetExtension(NID_ctPrecertificateRedactedLabelCount,
+                   &exty) != Cert::TRUE) {
+    LOG(WARNING) << "Required CT redaction count extension missing from cert";
+    return Cert::FALSE;
+  }
+
+  // Unpack the extension contents, which should be SEQUENCE OF INTEGER
+  STACK_OF(ASN1_INTEGER)* integers = (STACK_OF(ASN1_INTEGER)*)
```
AlCutter: Can you constify the ptr?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      ASN1_seq_unpack_ASN1_INTEGER(exty->value->data, exty->value->length,
+                                   d2i_ASN1_INTEGER, ASN1_INTEGER_free);
+
+  if (integers != NULL) {
```
AlCutter:
```C++
if (integers) {
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    const int num_integers = sk_ASN1_INTEGER_num(integers);
```
AlCutter: personally, I much prefer c'tor initialisation:
```C++
const int num_integers(sk_ASN1_INTEGER_num(integers));
```
there are loads of places in the PR where you could do that instead, though it's just an aesthetic thing really.

cpp/log/cert.cc
AlCutter: Also could this be a `unique_ptr` w/ custom deallocator?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      LOG(WARNING) << "Too many integers in extension: " << num_integers
+          << " but only " << dns_alt_names.size() << " DNS names";
+      sk_ASN1_INTEGER_free(integers);
+      return Cert::FALSE;
+    }
+
+    // All the integers in the sequence must be positive, check the sign
+    // after conversion to BIGNUM
+    for (int i = 0; i < num_integers; ++i) {
+      ASN1_INTEGER* redacted_labels = sk_ASN1_INTEGER_value(integers, i);
```
AlCutter: constify the ptr?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      BIGNUM *value = ASN1_INTEGER_to_BN(redacted_labels, NULL);
```
AlCutter: `nullptr`, constify the ptr, c'tor init?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // So there now must be a CT extension and the name constraint must not be
+  // in error
+  if (HasExtension(NID_name_constraints) != Cert::TRUE
+      || HasExtension(NID_ctNameConstraintNologIntermediateCa) != Cert::TRUE) {
+    LOG(WARNING) < "Name constraint extension without CT extension";
+    return Cert::FALSE;
+  }
+
+  int crit;
+  NAME_CONSTRAINTS* nc = (NAME_CONSTRAINTS *) X509_get_ext_d2i(
```
AlCutter: constify the ptr?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      x509_, NID_name_constraints, &crit, NULL);
```
AlCutter: `nullptr`

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+  if (nc == NULL || crit == -1) {
```
AlCutter: `nullptr` or even `if (!nc || crit == -1)...`

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  }
+
+  // Search all the permitted subtrees, there must be at least one DNS
+  // entry and it must not be empty
+  bool seen_dns = false;
+
+  for (int permitted_subtree = 0;
+      permitted_subtree < sk_GENERAL_SUBTREE_num(nc->permittedSubtrees);
+      ++permitted_subtree) {
+    GENERAL_SUBTREE *perm_subtree = sk_GENERAL_SUBTREE_value(
```
AlCutter: constify ptr

cpp/log/cert.cc
Martin2112: GetExtension takes a non const pointer.

cpp/log/cert.cc
AlCutter: Could this return nullptr?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+        nc->permittedSubtrees, permitted_subtree);
+
+    if (perm_subtree->base != NULL && perm_subtree->base->type == GEN_DNS
```
AlCutter:
```C++
if (!perm_subtree->base ...)
```

cpp/log/cert.cc
Martin2112: It somehow looks wrong to me, like it's calling a method. Similarly in java you can't write if (ptr), which is why I do it.

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+  // TODO: Does not handle more complex cases at the moment and I'm
+  // not sure whether it should. E.g. a combination of multiple entries
+  // that end up covering the whole available range. For the moment
+  // things similar to the example in the RFC work.
+  for (int excluded_subtree = 0;
+      excluded_subtree < sk_GENERAL_SUBTREE_num(nc->excludedSubtrees);
+      ++excluded_subtree) {
+
+    GENERAL_SUBTREE *excl_subtree = sk_GENERAL_SUBTREE_value(
```
AlCutter: const ptr & c'tor init?
Could it be nullptr?

cpp/log/cert.cc
```diff
@@ -546,12 +588,311 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+        nc->excludedSubtrees, excluded_subtree);
+
+    // Only consider entries that are of type ipAddress (OCTET_STRING)
+    if (excl_subtree->base != NULL && excl_subtree->base->type == GEN_IPADD) {
```
AlCutter: if (!excl...)

cpp/log/cert.cc
AlCutter: unique_ptr w/ custom deallocator?

cpp/log/cert.cc
```diff
@@ -723,6 +1064,7 @@ Cert::Status TbsCertificate::ExtensionIndex(int extension_nid,
   return Cert::TRUE;
 }
 
+
 CertChain::CertChain(const string& pem_string) {
   // A read-only BIO.
   BIO* bio_in = BIO_new_mem_buf(const_cast<char*>(pem_string.data()),
```
AlCutter: consty pointer ctry init?

cpp/log/cert.cc
```diff
@@ -816,7 +1158,7 @@ Cert::Status CertChain::IsValidCaIssuerChainMaybeLegacyRoot() const {
   }
 
   Cert::Status status;
-  for (std::vector<Cert*>::const_iterator it = chain_.begin();
+  for (vector<Cert*>::const_iterator it = chain_.begin();
```
AlCutter:
```C++
for (auto it(...); ...
```
would be fine too.

cpp/log/cert.cc
Martin2112: This was there before. I just got rid of std:: via global replace, same for next comment.

cpp/log/cert.cc
Martin2112: Not sure what the magic incantation for this would be? The free functions for these objects are macro expansions, which I don't think helps.

pphaneuf: The compiler *should* be able to see through the expansions, but if this doesn't work, don't worry too much, I'd say...

```C++
unique_ptr<STACK_OF(GENERAL_NAME), void(STACK_OF(GENERAL_NAME)*)> subject_alt_names((STACK_OF(GENERAL_NAME)*)X509_get_ext_d2i(x509_, NID_subject_alt_name, NULL, NULL), sk_GENERAL_NAME_free);
```

I'll leave reformatting to you. :wink: 

Also, you've used a C-style cast, should use a C++-style one (probably `static_cast`, I'd guess here).

pphaneuf: And you might be able to make the `unique_ptr` itself `const`?

cpp/log/cert.cc
pphaneuf: It *is* calling a method: the `int` constructor. :wink:

cpp/log/cert.cc
pphaneuf: And the `*` should be on the `BIGNUM`.

cpp/log/cert.cc
Martin2112: I tried something similar to that earlier. It produces a bunch of errors that look template related with "sk_GENERAL_NAME_free was not declared in this scope" in the middle.

cpp/log/cert.cc
```diff
@@ -92,6 +111,27 @@ Cert::Status Cert::LoadFromDerString(const std::string& der_string) {
   return Cert::TRUE;
 }
 
+Cert::Status Cert::LoadFromDerBio(BIO *bio_in) {
```
pphaneuf: Style: `BIO* bio_in`

cpp/log/cert.cc
```diff
@@ -536,9 +577,10 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
   // "extension not found, found more than once or corrupt".
   Cert::Status status = HasExtension(extension_nid);
   if (status != TRUE)
-    return status;
+      return status;
```
pphaneuf: This got overly indented?

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+  int name_pos = -1;
+
+  name_pos = X509_NAME_get_index_by_NID(name, NID_commonName, name_pos);
+
+  if (name_pos >= 0) {
+    X509_NAME_ENTRY* const name_entry(X509_NAME_get_entry(name, name_pos));
+
+    if (name_entry) {
+      ASN1_STRING * const subject_name_asn1(
```
pphaneuf: Style: `ASN1_STRING* const`...

cpp/log/cert.cc
```diff
@@ -36,10 +37,28 @@ static int X509_get_signature_nid(const X509* x) {
 
 namespace cert_trans {
 
+// Convert string from ASN1 and check it doesn't contain nul characters
+string ASN1ToStringAndCheckForNulls(ASN1_STRING *asn1_string,
```
pphaneuf: Style: `*` goes with the type.

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      }
+    }
+  }
+
+  return true;
+}
+
+// Helper method for validating V2 redaction rules. If it returns true
+// then the result in status is final.
+bool Cert::ValidateRedactionSubjectAltNameAndCN(int *dns_alt_name_count,
```
pphaneuf: Style: `*` goes with the types (next line too).

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      }
+    }
+  }
+
+  // If both a subject CN and DNS ids are present in the cert then the
+  // first DNS id must exactly match the CN
+  if (!dns_alt_names.empty() && common_name.length() > 0) {
+    if (dns_alt_names[0] != common_name) {
+      LOG(WARNING) << "CN " << common_name << " does not match DNS.0 "
+          << dns_alt_names[0];
```
pphaneuf: Style: the `<<` should be indented to go along the first ones on the previous line.

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // that the required extensions are present and specified correctly if
+  // we found any redacted names. First though if nothing is redacted
+  // then the rest of the rules need not be applied
+  if (redacted_name_count == 0 && !IsRedactedHost(common_name)) {
+    *status = Cert::TRUE;
+    return true;
+  }
+
+  *dns_alt_name_count = dns_alt_names.size();
+  return false; // validation has no definite result yet
```
pphaneuf: Style: need two spaces before a comment.

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      ASN1_seq_unpack_ASN1_INTEGER(exty->value->data, exty->value->length,
+                                   d2i_ASN1_INTEGER, ASN1_INTEGER_free)));
+
+  if (integers != NULL) {
+    const int num_integers = sk_ASN1_INTEGER_num(integers);
+
+    // RFC text says there MUST NOT be more integers than there are DNS ids
+    if (num_integers > dns_alt_name_count) {
+      LOG(WARNING) << "Too many integers in extension: " << num_integers
+          << " but only " << dns_alt_name_count << " DNS names";
```
pphaneuf: Style: indent further.

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      return Cert::FALSE;
+    }
+
+    // All the integers in the sequence must be positive, check the sign
+    // after conversion to BIGNUM
+    for (int i = 0; i < num_integers; ++i) {
+      ASN1_INTEGER* const redacted_labels(sk_ASN1_INTEGER_value(integers, i));
+      BIGNUM* const value(ASN1_INTEGER_to_BN(redacted_labels, nullptr));
+
+      const bool neg = value -> neg;
```
pphaneuf: Style: no spaces around `->`.

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      ASN1_INTEGER_free(redacted_labels);
+
+      if (neg) {
+        LOG(WARNING) << "Invalid negative redaction label count: "
+            << BN_bn2hex(value);
```
pphaneuf: Style: indent further.

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+        if (excl_subtree->base->d.ip->data[i] != 0) {
+          all_zero = false;
+        }
+      }
+
+      if (all_zero) {
+        if (excl_subtree->base->d.ip->length == 32) {
+          // IPv6
+          seen_ipv6 = true;
+        } if (excl_subtree->base->d.ip->length == 8) {
```
pphaneuf: Did you mean an `else` there? If not, the `if` should go on its own line.

cpp/log/cert.cc
```diff
@@ -723,10 +1090,11 @@ Cert::Status TbsCertificate::ExtensionIndex(int extension_nid,
   return Cert::TRUE;
 }
 
+
 CertChain::CertChain(const string& pem_string) {
   // A read-only BIO.
-  BIO* bio_in = BIO_new_mem_buf(const_cast<char*>(pem_string.data()),
-                                pem_string.length());
+  BIO* const bio_in(BIO_new_mem_buf(const_cast<char*>(pem_string.data()),
+                                pem_string.length()));
```
pphaneuf: Style: indentation needs to be a bit further.

cpp/log/cert.h
```diff
@@ -52,9 +58,13 @@ class Cert {
   // underlying copy succeeded.
   Cert* Clone() const;
 
-  // Frees the old X509 and attempts to load anew.
+  // Frees the old X509 and attempts to load a new one.
   Status LoadFromDerString(const std::string& der_string);
 
+  // Frees the old X509 and attempts to load from BIO in DER form. Caller
+  // still owns the BIO afterwards.
+  Status LoadFromDerBio(BIO *bio_in);
```
pphaneuf: Style: `*` with type.

cpp/log/cert.h
```diff
@@ -196,6 +213,8 @@ class Cert {
   Status ExtensionIndex(int extension_nid, int* extension_index) const;
   Status GetExtension(int extension_nid, X509_EXTENSION** ext) const;
   Status ExtensionStructure(int extension_nid, void** ext_struct) const;
+  bool ValidateRedactionSubjectAltNameAndCN(int *dns_alt_name_count,
```
pphaneuf: Style: `*` with type (next line too).

pphaneuf:
Ran a bit of Stylerator (tm)... :smiley: 

It *was* confused by the OpenSSL `STACK_OF` macro, didn't recognize it as a type, hehe!

cpp/log/cert.h
Martin2112: All of these should be fixed now. BTW are you sure about the log indents?
I've got Eclipse configured with Google style and it disagrees. It's not
always right though.

Martin

Martin2112:
I think I've addressed all the issues on this now.

Martin2112:
Ooops.

cpp/log/ct_extensions.h
```diff
@@ -22,6 +37,8 @@ extern const char kEmbeddedSCTListOID[];
 extern const char kPoisonOID[];
 // Extended Key Usage value for Precertificate signing
 extern const char kPrecertificateSigningOID[];
+// Name constrained intermediate CA may not be logged
+extern const char kNameConstraintNologIntermediateOID[];
```
pphaneuf: I suppose you're doing like the others here, but I'm wondering... is there a reason to have those visible in the header, rather than ensconced in the warmth of an anonymous namespace, in the `.cc` file?

Martin2112: I don't know why this pattern was chosen but as the code interacts with the
OpenSSL C API I decided to follow what existed.

I think we should organize a mini fixit for a couple of days and try to
bring everything up to C++11 etc. Better use of namespaces could be part of
that.

Martin

pphaneuf: I agree, it's fine to do it as a separate effort (and keep it consistent in this PR), I was just wondering if there was some reference that I missed.

cpp/log/cert.cc
AlCutter: I guess that depends whether you're testing these methods directly or via the public API of this class...

cpp/log/cert.cc
AlCutter: two lines between top level items here

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // TODO: Confirm it's valid to not have a CN.
+  X509_NAME* const name(X509_get_subject_name(x509_));
+
+  if (!name) {
+    LOG(ERROR) << "Missing X509 subject name";
+    *status = Cert::ERROR;
+    return true;
+  }
+
+  int name_pos = -1;
```
AlCutter:
```C++
const int name_pos(X509_NAME_get_index_by_NID(...., -1));
```
?

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+    LOG(WARNING) << "Required CT redaction count extension missing from cert";
+    return Cert::FALSE;
+  }
+
+  // Unpack the extension contents, which should be SEQUENCE OF INTEGER
+  STACK_OF(ASN1_INTEGER)* const integers(static_cast<STACK_OF(ASN1_INTEGER)*>(
+      ASN1_seq_unpack_ASN1_INTEGER(exty->value->data, exty->value->length,
+                                   d2i_ASN1_INTEGER, ASN1_INTEGER_free)));
+
+  if (integers != NULL) {
```
AlCutter:
```C++
if (integers)
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+      } else if (pos > 0) {
+        // '*' is only valid at the left
+        return false;
+      }
+    }
+  }
+
+  return true;
+}
+
```
AlCutter: two lines between top level items (here and throughout)

cpp/log/cert.cc
AlCutter: There are a couple of large (100+ line) methods added here, can they be broken up a bit to make it easier to read?

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+          *status = cn_status;
+          return true;
+        }
+      }
+    }
+  }
+
+  // If both a subject CN and DNS ids are present in the cert then the
+  // first DNS id must exactly match the CN
+  if (!dns_alt_names.empty() && common_name.length() > 0) {
```
AlCutter:
```C++
... && !common_name.empty()...
```

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  // So there now must be a CT extension and the name constraint must not be
+  // in error
+  if (HasExtension(NID_name_constraints) != Cert::TRUE
+      || HasExtension(NID_ctNameConstraintNologIntermediateCa) != Cert::TRUE) {
+    LOG(WARNING) < "Name constraint extension without CT extension";
+    return Cert::FALSE;
+  }
+
+  int crit;
+  NAME_CONSTRAINTS* const nc((NAME_CONSTRAINTS *) X509_get_ext_d2i(
```
AlCutter: use C++ style casts

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+  }
+
+  // Search all the permitted subtrees, there must be at least one DNS
+  // entry and it must not be empty
+  bool seen_dns = false;
+
+  for (int permitted_subtree = 0;
+      permitted_subtree < sk_GENERAL_SUBTREE_num(nc->permittedSubtrees);
+      ++permitted_subtree) {
+    GENERAL_SUBTREE * const perm_subtree(sk_GENERAL_SUBTREE_value(
```
AlCutter: * attaches to type

cpp/log/cert.cc
```diff
@@ -546,12 +588,337 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
...
+
+  // TODO: Does not handle more complex cases at the moment and I'm
+  // not sure whether it should. E.g. a combination of multiple entries
+  // that end up covering the whole available range. For the moment
+  // things similar to the example in the RFC work.
+  for (int excluded_subtree = 0;
+      excluded_subtree < sk_GENERAL_SUBTREE_num(nc->excludedSubtrees);
+      ++excluded_subtree) {
+
+    GENERAL_SUBTREE * const excl_subtree(sk_GENERAL_SUBTREE_value(
```
AlCutter: * attaches to type

cpp/log/cert.cc
Martin2112: I'm testing the pattern matching separately to avoid the need to create a lot of test certs with bad values.

pphaneuf:
There's been a number of surrounding fixes to the code this touches, in other pull requests... Once you've address this last round of comments, maybe it'd be good to rebase it on top of the latest `upstream/master`, to give it a refresh?

Martin2112:
I've broken out another helper method to split up the biggest one. I prefer not to do much more of this as it involves passing various bits dug out of the certs in both directions and could get ugly. I think I've addressed the other issues. PTAL.

cpp/log/cert.cc
AlCutter: So should it be returning `FALSE`, then? (as opposed to checking in unused but "broken" code I mean)

Martin2112: Oh dear, cut and paste failure. I've fixed it up. I could move it to the next PR if you like.

test/generate_v2_unit_test_certs.sh
```diff
@@ -0,0 +1,100 @@
...
+# in tests only. Some of the output certs do not follow the rules in the RFC.
+# Test case numbering relates to spreadsheet. Not all cases may be tested so
+# there could be gaps in numbering.
+
+set -e
+
+TMP_REQ=/tmp/req.csr
+
+create_cert() {
+  config=$1
```
AlCutter: should these all be declared `local`? (same in other functions)

test/generate_v2_unit_test_certs.sh
```diff
@@ -0,0 +1,100 @@
+# This generates a bunch of certs for unit tests of validating the rules
```
AlCutter: no shbang on purpose?

test/generate_v2_unit_test_certs.sh
```diff
@@ -0,0 +1,100 @@
...
+# sections 3.2.2 and 3.2.3 of the V2 CT RFC. They are intended for use
+# in tests only. Some of the output certs do not follow the rules in the RFC.
+# Test case numbering relates to spreadsheet. Not all cases may be tested so
+# there could be gaps in numbering.
+
+set -e
+
+TMP_REQ=/tmp/req.csr
```
AlCutter: Worth using `$(mktemp)` in here? (or mktemp -d and then blowing the whole dir away afterwards?)

There are some other `/tmp/file` references too if you decide to do this.

Martin2112: It's a reasonable point. I've got the same issue in the CMS verify test cert script so I'll defer and address them both there.

benlaurie:
Perhaps a bit late, but I was kinda hoping we'd do future shell-type tests in Python, its a whole lot more readable than shell...

Martin2112:
This just creates the certs that the C++ tests use, same for the followup
CMS tests. Most people won't need to run these scripts.

Martin

PR #860, opened 2015-07-31
Add test coverage for copying of issuer and root certs subjects to certificate_pb2.X509Certificate

phad:


python/ct/client/reporter_test.py
```diff
@@ -95,5 +107,35 @@ def test_entry_type_propogated(self):
...
+        result = report.report()
+        self.assertEqual(len(sum(result.values(), [])), 0)
+
+        certs = report.certs()
+        self.assertEqual(len(certs), 1)
+
+        issuer_cert = cert_desc.from_cert(cert.Certificate(CHAIN_DERS[1]))
+        root_cert = cert_desc.from_cert(cert.Certificate(CHAIN_DERS[2]))
+
+        self.assertEqual(readable_dn(certs[0].issuer),
```
eranmes: Can you write out the explicitly-expected subject here?

phad: Yeah that's a good idea, I'll do that.

PR #863, opened 2015-08-06
Avoid trying to read flags when a module is imported

RJPercival:
Using a command-line flag as the default value for a parameter will not work, as these values are evaluated when the module is imported. This is almost always prior to the flags being parsed.

Also removes one such parameter that was not being used (pool_size).

### DIFF
--- a/python/ct/client/reporter.py
+++ b/python/ct/client/reporter.py
@@ -104,11 +104,10 @@ class CertificateReport(object):
     __metaclass__ = abc.ABCMeta
 
     def __init__(self, checks=all_checks.ALL_CHECKS,
-                 pool_size=FLAGS.reporter_workers,
-                 queue_size=FLAGS.reporter_queue_size):
+                 queue_size=None):
         self.reset()
         self.checks = checks
-        self._jobs = Queue(queue_size)
+        self._jobs = Queue(queue_size or FLAGS.reporter_queue_size)
         self._pool = None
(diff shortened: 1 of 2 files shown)


python/ct/client/reporter.py
```diff
@@ -104,11 +104,10 @@ class CertificateReport(object):
     __metaclass__ = abc.ABCMeta
 
     def __init__(self, checks=all_checks.ALL_CHECKS,
-                 pool_size=FLAGS.reporter_workers,
```
eranmes: Can you confirm there is no code that creates a Reporter class and passes in these parameters that would be broken by this change?

eranmes:
LGTM assuming Travis is happy with this change and you address the comment about potential breakage.

RJPercival:
That's correct, the constructor is only called from subclasses and none of
those accept that parameter, nor do they provide a value for it.

On Thu, 6 Aug 2015 10:23 Eran Messeri <notifications@github.com> wrote:

> LGTM assuming Travis is happy with this change and you address the comment
> about potential breakage.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/certificate-transparency/pull/863#issuecomment-128303462>
> .
>
-- 
Google UK Ltd | 123 Buckingham Palace Road | London SW1W 9SH | Registered
in England Number: 3977902

PR #865, opened 2015-08-12
Osxportability

Martin2112:


Martin2112:
Sorry, struggling with github. This seems to include stuff from other branches that I wasn't expecting. Going to try again.

PR #866, opened 2015-08-12
Improve Portability, Specifically for OSX

Martin2112:
Fixes a number of issues on OSX and addresses test failures. Includes renaming death tests so they run first, network byte ordering, assumptions that /etc/certs exists, use of thread_local qualifier. Added new checks to configure.ac to support test fixes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -109,6 +109,21 @@ AC_DEFUN([CT_CHECK_TLS],
      fi
    fi])
 
+AC_DEFUN([CT_CHECK_INADDR_LOOPBACK],
+  [AC_CACHE_CHECK(
+     [whether the platform defines INADDR_LOOPBACK],
+     [ct_cv_inaddr_loopback],
+     [AC_LINK_IFELSE(
+        [AC_LANG_PROGRAM([ #include <netinet/in.h>
+           static int var = INADDR_LOOPBACK;], [return var;])],
+        [ct_cv_inaddr_loopback=yes],
+        [ct_cv_inaddr_loopback=no])])
+   if test $ct_cv_inaddr_loopback = yes; then
+     AC_DEFINE([HAVE_INADDR_LOOPBACK], [1],
+               [Define to 1 if the platform defines INADDR_LOOPBACK constant
+                in <netinet/in.h>.])
+   fi])
+
 AC_MSG_CHECKING([checking for gflags library])
 LIBS="-lgflags $LIBS"
 AC_LINK_IFELSE([AC_LANG_PROGRAM([#include <gflags/gflags.h>], [google::ParseCommandLineFlags(NULL, NULL, true)])], [have_gflags=yes], [have_gflags=no])
@@ -173,6 +188,7 @@ AC_TYPE_UINT64_T
 AC_TYPE_UINT8_T
 
 CT_CHECK_TLS
+CT_CHECK_INADDR_LOOPBACK
 
 AC_MSG_CHECKING([whether pthread_t is a pointer])
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([
--- a/configure.ac
+++ b/configure.ac
@@ -72,6 +72,43 @@ AC_SEARCH_LIBS([SSL_CTX_new], [ssl],, [missing_openssl=1], [$save_LIBS])
 AS_IF([test -n "$missing_openssl"],
       [AC_MSG_ERROR([could not find the OpenSSL libraries])])
 
+dnl Checks for thread-local storage support.
+dnl
+dnl Taken from the openvswitch config code (Apache 2.0 License)
+dnl with some local modifications. Does not include <threads.h>
+dnl as this does not currently exist on GCC.
+dnl Checks whether the compiler and linker support the C11
+dnl thread_local macro from <threads.h>, and if so defines
+dnl HAVE_THREAD_LOCAL.  If not, checks whether the compiler and linker
+dnl support the GCC __thread extension, and if so defines
+dnl HAVE___THREAD.
+AC_DEFUN([CT_CHECK_TLS],
+  [AC_CACHE_CHECK(
+     [whether $CC has <threads.h> that supports thread_local],
+     [ct_cv_thread_local],
+     [AC_LINK_IFELSE(
+        [AC_LANG_PROGRAM([ static thread_local int var;], [return var;])],
+        [ct_cv_thread_local=yes],
+        [ct_cv_thread_local=no])])
+   if test $ct_cv_thread_local = yes; then
+     AC_DEFINE([HAVE_THREAD_LOCAL], [1],
+               [Define to 1 if the C compiler and linker supports the C11
+                thread_local macro defined in <threads.h>.])
+   else
+     AC_CACHE_CHECK(
+       [whether $CC supports __thread],
+       [ct_cv___thread],
+       [AC_LINK_IFELSE(
+          [AC_LANG_PROGRAM([static __thread int var;], [return var;])],
+          [ct_cv___thread=yes],
+          [ct_cv___thread=no])])
+     if test $ct_cv___thread = yes; then
+       AC_DEFINE([HAVE___THREAD], [1],
+                 [Define to 1 if the C compiler and linker supports the
+                  GCC __thread extensions.])
+     fi
+   fi])
+
 AC_MSG_CHECKING([checking for gflags library])
 LIBS="-lgflags $LIBS"
 AC_LINK_IFELSE([AC_LANG_PROGRAM([#include <gflags/gflags.h>], [google::ParseCommandLineFlags(NULL, NULL, true)])], [have_gflags=yes], [have_gflags=no])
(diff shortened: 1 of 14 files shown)


cpp/log/etcd_consistent_store_test.cc
```diff
@@ -497,9 +497,9 @@ TEST_F(EtcdConsistentStoreDeathTest,
                "sequence_number\\(\\) < mapping");
 }
 
-
-TEST_F(EtcdConsistentStoreDeathTest,
-       TestUpdateSequenceMappingBarfsWithMappingNonContiguousToServingTree) {
+TEST_F(
+    EtcdConsistentStoreDeathTest,
```
benlaurie: Slightly eccentric formatting. Leave first line alone and outdent second?

cpp/merkletree/compact_merkle_tree.cc
```diff
@@ -21,7 +21,7 @@ CompactMerkleTree::CompactMerkleTree(SerialHasher* hasher)
 
 CompactMerkleTree::CompactMerkleTree(MerkleTree& model, SerialHasher* hasher)
     : MerkleTreeInterface(),
-      tree_(std::max(0L, int64_t(model.LevelCount()) - 1)),
+      tree_(std::max(int64_t(0), int64_t(model.LevelCount()) - 1)),
```
benlaurie: Won't just 0 do?

Martin2112: Sadly not. It produces the same type mismatch error as 0L.

cpp/log/etcd_consistent_store_test.cc
Martin2112: Will do.

cpp/log/etcd_consistent_store_test.cc
```diff
@@ -289,8 +289,8 @@ TEST_F(EtcdConsistentStoreTest, TestSetServingSTHWontOverwriteWithOlder) {
   EXPECT_EQ(util::error::OUT_OF_RANGE, status.CanonicalCode()) << status;
 }
 
-
-TEST_F(EtcdConsistentStoreDeathTest, TestSetServingSTHChecksInconsistentSize) {
+TEST_F(EtcdConsistentStoreDeathTest,
+       TestSetServingSTHChecksInconsistentSizeDeathTest) {
```
pphaneuf: I'm finding this very mysterious. Hold on a sec before merging in...

pphaneuf: I'm pretty sure that the test case, rather than individual tests, are supposed to be called `*DeathTest`? The [How to Write a Death Test](https://code.google.com/p/googletest/wiki/V1_7_AdvancedGuide#How_to_Write_a_Death_Test) section says this:

> *Important*: We strongly recommend you to follow the convention of naming your test case (not test) `*DeathTest` when it contains a death test, as demonstrated in the above example. The [Death Tests And Threads](https://code.google.com/p/googletest/wiki/V1_7_AdvancedGuide#Death_Tests_And_Threads) section below explains why.

This change adds `DeathTest` to the tests (it was already there for the test cases), in direct opposition of this "important" recommendation. Maybe that's the thing to do, but I'm finding it suspicious that we're going directly against the recommendation here...

The FAQ later talks a bit more on this:

https://code.google.com/p/googletest/wiki/FAQ#Why_does_Google_Test_require_the_entire_test_case,_instead_of_in

Martin2112: OK but I had consistent failures of the death tests OSX until I made this change. I wonder if I was reading out of date docs because the impression I got was that it applied to the tests. Also we might have some cases that mix non death tests?

I also note one test failed in the travis build but they all passed locally. Sigh.

I'll investigate all of this more tomorrow.

pphaneuf: From what I can tell here, it runs all the tests in `FooDeathTest`, *then* all the tests in `FooTest`. For example:

```
[==========] Running 5 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 1 test from ThreadPoolDeathTest
[ RUN      ] ThreadPoolDeathTest.AddingMoreTasksAfterClosedGoesBang

[WARNING] /usr/src/gmock/gtest/src/gtest-death-test.cc:825:: Death tests use fork(), which is unsafe particularly in a threaded context. For this test, Google Test couldn't detect the number of threads.
[       OK ] ThreadPoolDeathTest.AddingMoreTasksAfterClosedGoesBang (200 ms)
[----------] 1 test from ThreadPoolDeathTest (200 ms total)

[----------] 4 tests from ThreadPoolTest
[ RUN      ] ThreadPoolTest.Delay
[       OK ] ThreadPoolTest.Delay (200 ms)
[ RUN      ] ThreadPoolTest.DelayDoesNotBlockAThread
[       OK ] ThreadPoolTest.DelayDoesNotBlockAThread (200 ms)
[ RUN      ] ThreadPoolTest.NaturalOrderingPreserved
[       OK ] ThreadPoolTest.NaturalOrderingPreserved (201 ms)
[ RUN      ] ThreadPoolTest.CancelsDelayTasks
[       OK ] ThreadPoolTest.CancelsDelayTasks (0 ms)
[----------] 4 tests from ThreadPoolTest (601 ms total)

[----------] Global test environment tear-down
[==========] 5 tests from 2 test cases ran. (802 ms total)
[  PASSED  ] 5 tests.
```

The documentation seems to imply that there's something forcing you to put `EXPECT_DEATH` checks only in test cases named `*DeathTest`, but I just tried it, and it works (there's not even a warning).

So it's possible that we might have a stray `EXPECT_DEATH` in a test that doesn't conform to the naming, and that could very well throw things out of whack...

Martin2112:
I've rolled back the test name changes and fixed the test breakage (dependencies). I think it's about ready to merge but it would be good if someone could sanity check my configure.ac changes.

configure.ac
```diff
@@ -72,6 +72,58 @@ AC_SEARCH_LIBS([SSL_CTX_new], [ssl],, [missing_openssl=1], [$save_LIBS])
...
+          [ct_cv___thread=yes],
+          [ct_cv___thread=no])])
+     if test $ct_cv___thread = yes; then
+       AC_DEFINE([HAVE___THREAD], [1],
+                 [Define to 1 if the C compiler and linker supports the
+                  GCC __thread extensions.])
+     fi
+   fi])
+
+AC_DEFUN([CT_CHECK_INADDR_LOOPBACK],
```
pphaneuf: This one is overkill, I think? You could do it with just this (make sure to blow away `cpp/config.h` and `cpp/config.h.in` when trying it out):

    AC_CHECK_DECLS([INADDR_LOOPBACK], [], [], [#include <netinet/in.h>])

This will define `HAVE_DECL_INADDR_LOOPBACK` when it's present in the header specified.

configure.ac
```diff
@@ -72,6 +72,58 @@ AC_SEARCH_LIBS([SSL_CTX_new], [ssl],, [missing_openssl=1], [$save_LIBS])
...
+dnl
+dnl Taken from the openvswitch config code (Apache 2.0 License)
+dnl with some local modifications. Does not include <threads.h>
+dnl as this does not currently exist on GCC.
+dnl Checks whether the compiler and linker support the C11
+dnl thread_local macro from <threads.h>, and if so defines
+dnl HAVE_THREAD_LOCAL.  If not, checks whether the compiler and linker
+dnl support the GCC __thread extension, and if so defines
+dnl HAVE___THREAD.
+AC_DEFUN([CT_CHECK_TLS],
```
pphaneuf: I *think* this does what we need, but could you put the function definition in its own file in the `m4` directory?

cpp/merkletree/compact_merkle_tree.cc
pphaneuf: I think you could avoid this (and the C-style rando-casts) by doing this?

    std::max<int64_t>(0, model.LevelCount() - 1)

pphaneuf:
I'm not sure I like this pull request as a whole... I can understand having a branch and many commits to work out everything that needs to be worked out, but this branch is a bit of a mess (even including merge commits unto itself!).

Now that everything is there, it shouldn't be hard to split this up in easy to digest chunks?

cpp/merkletree/compact_merkle_tree.cc
Martin2112: I can't do that as it causes one of the tree tests to fail with a
std::bad_alloc.

Martin

Martin2112:
Split out the m4 file and used AC_CHECK_DECLS.

pphaneuf:
For example, you could do this:

```shell
git checkout -b one_fix osxportability
git reset --soft upstream/master
git reset
git add -p <specific files for just one fix>
git commit
git reset --hard
```

This would create a new branch, letting you pick out just a subset of the changes.

Martin2112:
I've had such a joyful week with Git already, what could go wrong?

pphaneuf:
Ha!

The first command makes everything safe (enough), since it creates a new branch, thus leaving this one in peace. :wink: 

The only potentially tricky bit is the interactive `git add -p` command (which is one of my favourite!), but there's a "help" option in the interactive mode, shouldn't be a problem. It's just a question of staging, though, so again, non-destructive.

pphaneuf:
That last commit, for example, doesn't do everything that it claims to do in its description! :stuck_out_tongue:

Martin2112:
That's because I missed adding the other file.

Martin2112:
I'll try to split it up into multiple PRs. Stand by.

Martin2112:
Closing this pull request as it has been split into four new ones.

PR #867, opened 2015-08-13
Try to mitigate the race condition in the cluster controller test.

Martin2112:
This is not a perfect solution (which looks hard) but I got this test case to pass 300 times in succession, which doesn't happen without this fix.

### DIFF
--- a/cpp/log/cluster_state_controller_test.cc
+++ b/cpp/log/cluster_state_controller_test.cc
@@ -501,6 +501,16 @@ TEST_F(ClusterStateControllerTest, TestGetLocalNodeState) {
 
 
 TEST_F(ClusterStateControllerTest, TestNodeHostPort) {
+  // Allow some time for our view of the node state to stabilize and then
+  // check that it has. Ideally we would wait for this to definitely happen
+  // but there seems to be no easy way to arrange this.
+  sleep(1);
+  const ClusterNodeState orig_node_state(GetNodeStateView(kNodeId1));
+  EXPECT_EQ(kNodeId1, orig_node_state.hostname());
+  EXPECT_EQ(9001, orig_node_state.log_port());
+
+  // Now try to change the state and again allow some time for our view
+  // to update
   const string kHost("myhostname");
   const int kPort(9999);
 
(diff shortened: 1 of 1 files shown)


cpp/log/cluster_state_controller_test.cc
```diff
@@ -501,6 +501,16 @@ TEST_F(ClusterStateControllerTest, TestGetLocalNodeState) {
 
 
 TEST_F(ClusterStateControllerTest, TestNodeHostPort) {
+  // Allow some time for our view of the node state to stabilize and then
+  // check that it has. Ideally we would wait for this to definitely happen
+  // but there seems to be no easy way to arrange this.
+  sleep(1);
+  const ClusterNodeState orig_node_state(GetNodeStateView(kNodeId1));
+  EXPECT_EQ(kNodeId1, orig_node_state.hostname());
+  EXPECT_EQ(9001, orig_node_state.log_port());
```
pphaneuf: Maybe those two should be `ASSERT_EQ` instead? As in, if these fails, the rest of the test is kind of meaningless, right?

I guess you could make it a little loop like this, to try a bit harder, but not *too* hard either?

```C++
for (int i = 0; i < 10; ++i) {
  const ClusterNodeState orig_node_state(GetNodeStateView(kNodeId1));
  if (orig_node_state.hostname() == kNodeId1 && orig_node_state.log_port() == 9091) {
    break;
  }
  sleep(1);
}
```

pphaneuf: Would be nicer if it counted *down* instead, and had a `ASSERT_NE(0, i) << "too many attempts";`, actually.

Martin2112: You're technically correct .. but I can't do this kind of loop because GetNodeStateView will fail a CHECK if the node isn't found in the peer list yet. Similarly you're right it should probably be ASSERT_EQ for clarity but in reality it will have exited before that for the same reason.

pphaneuf: Ah, I see...

Would be nicer to have `GetNodeStateView` return a `util::StatusOr<ClusterNodeState>` instead, in the future. Maybe add a TODO?

pphaneuf:
LGTM

cpp/log/cluster_state_controller_test.cc
Martin2112: Done. Also added TODO that it's best not to use sleep() in tests and this could be improved.

PR #868, opened 2015-08-14
Output the error message when there is a problem.

pphaneuf:
This happens more often than we like, at least give a hint...


Martin2112:
LGTM

PR #869, opened 2015-08-14
Update tests that relied on /etc/ssl/certs to use our test certs

Martin2112:
These tests don't need particular ones but fail if we can't initialize the store so they now use our generated test certs instead.


pphaneuf:
LGTM

PR #870, opened 2015-08-14
Some portability improvements, mostly for tests but also thread local

Martin2112:
Add support to configure.ac for checking thread local support
and also check IPADDR_LOOPBACK is defined. In test use
loopback if we can to prevent firewall issues and use correct
thread local qualifier in sever. Plus ensure server.h qualifies
calls to std::bind to prevent clash with C socket API.

### DIFF
--- a/cpp/net/url_fetcher_test.cc
+++ b/cpp/net/url_fetcher_test.cc
@@ -1,3 +1,5 @@
+#include "config.h"
+
 #include <csignal>
 #include <fcntl.h>
 #include <gflags/gflags.h>
@@ -7,6 +9,9 @@
 #include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#ifdef HAVE_VFORK_H
+#include <vfork.h>
+#endif
 
 #include "net/connection_pool.h"
 #include "net/url_fetcher.h"
@@ -57,7 +62,12 @@ class LocalhostResolver : public libevent::Base::Resolver {
 pid_t RunOpenSSLServer(uint16_t port, const std::string& cert_file,
                        const std::string& key_file,
                        const std::string& mode = "-www") {
+#ifdef HAVE_WORKING_VFORK
+  pid_t pid(vfork());
+#else
   pid_t pid(fork());
+#endif
+
   if (pid == -1) {
     LOG(INFO) << "fork() failed: " << pid;
   } else if (pid == 0) {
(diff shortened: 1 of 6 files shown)


m4/ct_check_tls.m4
```diff
@@ -0,0 +1,37 @@
...
+dnl Checks whether the compiler and linker support the C11
+dnl thread_local macro from <threads.h>, and if so defines
+dnl HAVE_THREAD_LOCAL.  If not, checks whether the compiler and linker
+dnl support the GCC __thread extension, and if so defines
+dnl HAVE___THREAD.
+AC_DEFUN([CT_CHECK_TLS],
+  [AC_CACHE_CHECK(
+     [whether $CC has <threads.h> that supports thread_local],
+     [ct_cv_thread_local],
+     [AC_LINK_IFELSE(
```
pphaneuf: This *might* work as a `AC_COMPILE_IFELSE` instead of an `AC_LINK_IFELSE`, but if it works like this, I'd be tempted to leave it well alone. :stuck_out_tongue:

cpp/net/url_fetcher_test.cc
```diff
@@ -7,6 +9,9 @@
 #include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#ifdef HAVE_VFORK_H
```
pphaneuf: Since there's a separate `HAVE_VFORK_H` and `HAVE_WORKING_VFORK`, I wonder where `vfork()` is when you don't have `vfork.h`?!? Or maybe there's never a working `vfork()` when there isn't a `vfork.h`?

Anyway, works on Linux and Mac OS X, good enough for me. :wink:

cpp/server/event.cc
```diff
@@ -251,7 +251,7 @@ bool Services::InitServer(int* sock, int port, const char* ip, int type) {
   server.sin_family = AF_INET;
   server.sin_port = htons((unsigned short)port);
   if (ip == NULL)
-    server.sin_addr.s_addr = INADDR_ANY;
+    server.sin_addr.s_addr = htonl(INADDR_ANY);
```
pphaneuf: Considering `INADDR_ANY` is usually 0, how did you find this?!?

pphaneuf:
LGTM, I just have random questions, for my own edification. :smiley:

cpp/server/event.cc
Martin2112: Another change where I switched to INADDR_LOOPBACK broke so I changed all occurrences for safety.

cpp/net/url_fetcher_test.cc
Martin2112: I think vfork.h might not exist and on some platforms the vfork API exists but does not work correctly. These days it shouldn't be necessary to use vfork but it fixed this test on my laptop.

PR #871, opened 2015-08-14
Some fixes to tests, flakiness improvement and test ordering

Martin2112:
Trying to reduce cluster state test flakiness and making sure death tests are in the correctly
named test suite.

### DIFF
--- a/cpp/log/database_test.cc
+++ b/cpp/log/database_test.cc
@@ -47,7 +47,12 @@ typedef testing::Types<FileDB<cert_trans::LoggedCertificate>,
 
 typedef Database<cert_trans::LoggedCertificate> DB;
 
+template <class T>
+class DBTestDeathTest : public DBTest<T> {
+};
+
 TYPED_TEST_CASE(DBTest, Databases);
+TYPED_TEST_CASE(DBTestDeathTest, Databases);
 
 
 TYPED_TEST(DBTest, CreateSequenced) {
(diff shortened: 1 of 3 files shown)


pphaneuf:
Could probably fix up this little mistake with `git rebase -i HEAD^^`, and changing the `pick` on the second line to `fixup` (or just `f`, same thing).

Not requiring it, but it'd be neater. :wink:

cpp/log/database_test.cc
```diff
@@ -47,7 +47,12 @@ typedef testing::Types<FileDB<cert_trans::LoggedCertificate>,
 
 typedef Database<cert_trans::LoggedCertificate> DB;
 
+template <class T>
+class DBTestDeathTest : public DBTest<T> {
```
pphaneuf: Could use the newfangled way here, if you wanted (but this is reasonable too, I think):

```C++
template <typename T>
using DBDeathTest = DBTest<T>;
```

pphaneuf:
LGTM

cpp/log/database_test.cc
Martin2112: That doesn't seem to work. Possibly because TYPED_TEST_CASE is a macro? It produces lots of:

In file included from ../gmock-1.7.0/gtest/include/gtest/gtest.h:66:0,
                 from cpp/log/database_test.cc:2:
../gmock-1.7.0/gtest/include/gtest/gtest-typed-test.h:172:24: error: expected template-name before ‘<’ token
       : public CaseName<gtest_TypeParam_> { \
                        ^
cpp/log/database_test.cc:369:1: note: in expansion of macro ‘TYPED_TEST’
 TYPED_TEST(DBTestDeathTest, CannotOverwriteNodeId) {
 ^

pphaneuf: I tried it here, it works for me? Had to replace the whole templated class definition with the snippet I gave, but was fine. Oh well, no big deal, keep it like this, then. :smiley:

pphaneuf:
Oooh, those "merge branch" commits that were added, I'm not so sure about those... :wink:

Martin2112:
It wouldn't let me push unless I merged. I couldn't see why this was
happening.

Martin

Martin2112:
What's the best way to fix this up?

benlaurie:
On Wed, 19 Aug 2015 at 09:31 Martin2112 <notifications@github.com> wrote:

> What's the best way to fix this up?
>

git rebase -i master
git push -f

pphaneuf:
Except `git rebase -i upstream/master`, in his case. And might need a `git fetch upstream`.

Martin2112:
Thanks Ben + Pierre. I think the branch is now correct.

pphaneuf:
Ah, the "files changed" got a lot neater! :smiley: 

LGTM

PR #872, opened 2015-08-14
Portability tweak to compact merkle tree code.

Martin2112:
Constructor tweak to CompactMerkleTree. Alternatives tried either did not compile or caused tests to fail at runtime with std::bad_alloc.

### DIFF
--- a/cpp/merkletree/compact_merkle_tree.cc
+++ b/cpp/merkletree/compact_merkle_tree.cc
@@ -21,7 +21,7 @@ CompactMerkleTree::CompactMerkleTree(SerialHasher* hasher)
 
 CompactMerkleTree::CompactMerkleTree(MerkleTree& model, SerialHasher* hasher)
     : MerkleTreeInterface(),
-      tree_(std::max(int64_t(0), int64_t(model.LevelCount()) - 1)),
+      tree_(std::max<int64_t>(0, model.LevelCount() - 1)),
       treehasher_(hasher),
       leaf_count_(model.LeafCount()),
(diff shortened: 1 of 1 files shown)


cpp/merkletree/compact_merkle_tree.cc
```diff
@@ -21,7 +21,7 @@ CompactMerkleTree::CompactMerkleTree(SerialHasher* hasher)
 
 CompactMerkleTree::CompactMerkleTree(MerkleTree& model, SerialHasher* hasher)
     : MerkleTreeInterface(),
-      tree_(std::max(0L, int64_t(model.LevelCount()) - 1)),
+      tree_(std::max(int64_t(0), int64_t(model.LevelCount()) - 1)),
```
pphaneuf: You said that this didn't work?

```C++
tree_(std::max<int64_t>(0, model.LevelCount() - 1)),
```

Martin2112: It didn't when I tried it on Friday but I tried again and it looks like it did. Maybe I missed off a bracket before.

pphaneuf: I'd like that better, because C-style casts make me hiss, due to the occasional silent promotion to `reinterpret_cast` (where the C++ ones are a bit less dangerous): http://en.cppreference.com/w/cpp/language/explicit_cast

Martin2112:
I assume I'm good to merge this?

pphaneuf:
LGTM

PR #874, opened 2015-08-20
Fix some client memory leaks, found by cppcheck.

Martin2112:
I ran cppcheck on our code and this should hopefully fix the three memory leaks it found.

### DIFF
--- a/cpp/client/client.cc
+++ b/cpp/client/client.cc
@@ -84,6 +84,7 @@ bool Client::Read(size_t length, string* result) {
       PCHECK(errno == EPIPE) << "Read failed";
       LOG(ERROR) << "Remote server closed the connection.";
       Disconnect();
+      delete[] buf;
       return false;
     }
 
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -254,7 +254,7 @@ static bool VerifySCTAndPopulateSSLClientCTData(
     const SignedCertificateTimestamp& sct, SSLClientCTData* ct_data) {
   SSLClientCTData::SCTInfo* sct_info = ct_data->add_attached_sct_info();
   sct_info->mutable_sct()->CopyFrom(sct);
-  unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
+  const unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
   string merkle_leaf;
   LogVerifier::VerifyResult result =
       verifier->VerifySignedCertificateTimestamp(
@@ -755,7 +755,7 @@ static void DiagnoseCertChain() {
       continue;
     }
     LOG(INFO) << "SCT number " << i + 1 << ":\n" << sct.DebugString();
-    if (verifier.get()) {
+    if (verifier) {
       if (sct.id().key_id() != verifier->KeyID()) {
         LOG(WARNING) << "SCT key ID does not match verifier's ID, skipping";
         continue;
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -913,7 +913,7 @@ int GetSTH() {
 
   // Allow for 10 seconds of clock skew
   uint64_t latest = ((uint64_t)time(NULL) + 10) * 1000;
-  LogVerifier::VerifyResult result =
+  const LogVerifier::VerifyResult result =
       verifier->VerifySignedTreeHead(sth, 0, latest);
 
   LOG(INFO) << "STH is " << sth.DebugString();
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -5,6 +5,7 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 #include <iostream>
+#include <memory>
 #include <openssl/asn1.h>
 #include <openssl/bio.h>
 #include <openssl/bn.h>
@@ -136,6 +137,7 @@ using ct::SignedCertificateTimestamp;
 using ct::SignedCertificateTimestampList;
 using std::shared_ptr;
 using std::string;
+using std::unique_ptr;
 using std::vector;
 using util::StatusOr;
 
@@ -252,7 +254,7 @@ static bool VerifySCTAndPopulateSSLClientCTData(
     const SignedCertificateTimestamp& sct, SSLClientCTData* ct_data) {
   SSLClientCTData::SCTInfo* sct_info = ct_data->add_attached_sct_info();
   sct_info->mutable_sct()->CopyFrom(sct);
-  LogVerifier* verifier = GetLogVerifierFromFlags();
+  unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
   string merkle_leaf;
   LogVerifier::VerifyResult result =
       verifier->VerifySignedCertificateTimestamp(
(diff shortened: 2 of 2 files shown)


cpp/client/ct.cc
```diff
@@ -753,7 +755,7 @@ static void DiagnoseCertChain() {
       continue;
     }
     LOG(INFO) << "SCT number " << i + 1 << ":\n" << sct.DebugString();
-    if (verifier != NULL) {
+    if (verifier.get()) {
```
pphaneuf: I think `unique_ptr` can be used directly in a conditional, no need to convert to pointer.

cpp/client/ct.cc
```diff
@@ -252,7 +254,7 @@ static bool VerifySCTAndPopulateSSLClientCTData(
     const SignedCertificateTimestamp& sct, SSLClientCTData* ct_data) {
   SSLClientCTData::SCTInfo* sct_info = ct_data->add_attached_sct_info();
   sct_info->mutable_sct()->CopyFrom(sct);
-  LogVerifier* verifier = GetLogVerifierFromFlags();
+  unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
```
pphaneuf: Could you make it `const`, while you're there?

cpp/client/ct.cc
```diff
@@ -907,7 +909,7 @@ int GetSTH() {
   ct::SignedTreeHead sth;
   CHECK_EQ(AsyncLogClient::OK, client.GetSTH(&sth));
 
-  LogVerifier* verifier = GetLogVerifierFromFlags();
+  unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
```
pphaneuf: And make this one `const` too?

cpp/client/client.cc
```diff
@@ -84,6 +84,7 @@ bool Client::Read(size_t length, string* result) {
       PCHECK(errno == EPIPE) << "Read failed";
       LOG(ERROR) << "Remote server closed the connection.";
       Disconnect();
+      delete[] buf;
```
pphaneuf: You could also convert this to `unique_ptr` as well (there's an array variant):

```C++
unique_ptr<char[]> buf(new char[length]);
```

pphaneuf:
Looks reasonable, just a few small tweaks... You could skip the `unique_ptr` in `Client::Read`, if you want, your call.

cpp/client/ct.cc
benlaurie: Tch! Pointers are not booleans.

cpp/client/ct.cc
Martin2112: That one can't be const alas. I have to call reset() as not all paths use the verifier.

cpp/client/client.cc
Martin2112: Not sure that works because the call to recv below does pointer arithmetic.

cpp/client/ct.cc
Martin2112: I thought our code style was "if (wibble)", should it be "if (wibble != nullptr)"?

cpp/client/ct.cc
Martin2112: Sorry, was looking at wrong line. This one can be const.

cpp/client/client.cc
pphaneuf: I think `buf.get() + offset` should work?

cpp/client/ct.cc
```diff
@@ -907,11 +909,11 @@ int GetSTH() {
   ct::SignedTreeHead sth;
   CHECK_EQ(AsyncLogClient::OK, client.GetSTH(&sth));
 
-  LogVerifier* verifier = GetLogVerifierFromFlags();
+  const unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
 
   // Allow for 10 seconds of clock skew
   uint64_t latest = ((uint64_t)time(NULL) + 10) * 1000;
-  LogVerifier::VerifyResult result =
+  const LogVerifier::VerifyResult result =
```
pphaneuf: Yay! Bonus `const`! :wink:

pphaneuf:
LGTM

cpp/client/client.cc
Martin2112: It got me about 10 pages of unreadable template-related compilation errors.

pphaneuf: Yech.

cpp/client/ct.cc
pphaneuf: No, @benlaurie is just remembering "C from the late 70's", due to the CDC 6000 he keeps around to work on OpenSSL... :wink:

This was explicitly legislated as "proper C++" a long time ago (and proper C as well, from at least C89, I think?):

https://github.com/cplusplus/draft/blob/093c1eebd3e2a1682331e8b1bd16677cbdb79d6f/source/conversions.tex#L549

benlaurie: On Thu, 20 Aug 2015 at 15:38 Pierre Phaneuf <notifications@github.com>
wrote:

> In cpp/client/ct.cc
> <https://github.com/google/certificate-transparency/pull/874#discussion_r37536309>
> :
>
> > @@ -753,7 +755,7 @@ static void DiagnoseCertChain() {
> >        continue;
> >      }
> >      LOG(INFO) << "SCT number " << i + 1 << ":\n" << sct.DebugString();
> > -    if (verifier != NULL) {
> > +    if (verifier.get()) {
>
> No, @benlaurie <https://github.com/benlaurie> is just remembering "C from
> the late 70's", due to the CDC 6000 he keeps around to work on OpenSSL... [image:
> :wink:]
>

Fun fact: second machine I ever coded for and first assembler I learnt was,
in fact, a CDC 6400.

This was explicitly legislated as "proper C++" a long time ago (and proper
> C as well, from at least C89, I think?):
>
>
> https://github.com/cplusplus/draft/blob/093c1eebd3e2a1682331e8b1bd16677cbdb79d6f/source/conversions.tex#L549
>

Actually, its not to do with legality, its to do with:

a) preference,

b) static analysis (one thing I've been playing with recently is ensuring
that non-bool values are correctly tested).

pphaneuf: I don't know why, I had a vague feeling you might have had some Control Data in your past life... :smiley:

http://c-faq.com/null/machexamp.html

If you're looking to go against this `if (p)` style, you've got a solid uphill battle ahead of you... For example, because `bool` is an integer (and so, will auto-convert to `int` and friends), objects that want to be "testable" aren't wise to implement an `operator bool`, but instead have a conversion operator to some non-void pointer (usually a private inner class that's never defined, or a method pointer, since those won't bloody convert even to things they ought to!).

I guess those objects would then be comparable with `nullptr`, say, but you're going to have a big pile of existing code that does this (and is perfectly correct to do it).

benlaurie: On Fri, 21 Aug 2015 at 22:36 Pierre Phaneuf <notifications@github.com>
wrote:

> In cpp/client/ct.cc
> <https://github.com/google/certificate-transparency/pull/874#discussion_r37680673>
> :
>
> > @@ -753,7 +755,7 @@ static void DiagnoseCertChain() {
> >        continue;
> >      }
> >      LOG(INFO) << "SCT number " << i + 1 << ":\n" << sct.DebugString();
> > -    if (verifier != NULL) {
> > +    if (verifier.get()) {
>
> I don't know why, I had a vague feeling you might have had some Control
> Data in your past life... [image: :smiley:]
>
> http://c-faq.com/null/machexamp.html
>

Sorry to disappoint, but the CDC 6400 had 18 bit pointers. And I'm pretty
sure null would've been zero.


> If you're looking to go against this if (p) style, you've got a solid
> uphill battle ahead of you... For example, because bool is an integer
> (and so, will auto-convert to int and friends), objects that want to be
> "testable" aren't wise to implement an operator bool, but instead have a
> conversion operator to some non-void pointer (usually a private inner class
> that's never defined, or a method pointer, since those won't bloody convert
> even to things they ought to!).
>
Hmm. That seems like a different thing, though.

PR #878, opened 2015-08-24
Refactoring to support multiple handler versions + server binaries

Martin2112:
This is my straw man proposal for supporting v1 and v2 handlers. There are other possible approaches, which would isolate handler implementations more.

Pulls out a 'server' library to share between ct-server and ct-mirror. Builds two binary versions with a different set of handlers, which have an abstract base and share all but pre-certificate submit at the moment. Avoids conditional compilation or configure time switches.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,6 +155,20 @@ AS_IF([test -n "$missing_libevhtp"],
       [AC_MSG_ERROR([could not find the evhtp library])])
 LIBS="$save_LIBS"
 
+# CT protocol version support. Only one version may be enabled
+# at compile time. Defaults to v1 support (RFC 6962).
+AC_ARG_WITH(
+  ctversion,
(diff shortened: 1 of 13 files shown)


configure.ac
```diff
@@ -64,8 +64,8 @@ AS_IF([test -n "$missing_gmock"],
 AC_SEARCH_LIBS([__b64_ntop], [resolv])
 AC_SEARCH_LIBS([pthread_create], [pthread])
 
-dnl We're pretty crypto-centric, having the OpenSSL libraries in LIBS
-dnl is fine.
+# We're pretty crypto-centric, having the OpenSSL libraries in LIBS
+# is fine.
```
pphaneuf: Could you slice out the changes to this file to another branch? Keeping the PR to just one thing, and as small as possible...

Martin2112:
Closing this because the server was changed underneath me. Going to try again using the new code.

PR #879, opened 2015-08-26
Make GCM exporter a bit more resilient to GCM hiccups.

AlCutter:


cpp/monitoring/gcm/exporter.cc
```diff
@@ -79,29 +84,35 @@ GCMExporter::~GCMExporter() {
...
+    task.Wait();
 
-  // TODO(alcutter): Handle this better
-  CHECK_EQ(util::Status::OK, task.status());
-  CHECK_EQ(200, resp.status_code);
-  token_refreshed_at_ = system_clock::now();
+    if (!task.status().ok() || resp.status_code != 200) {
+      LOG(WARNING) << "Failed to refresh GCM credentials, status: "
+                   << task.status() << ", response code: " << resp.status_code;
+      sleep(FLAGS_google_compute_monitoring_retry_delay_seconds);
```
pphaneuf: All this stuff runs on some thread pool, I'm presuming, but it's not of unlimited size... "One more thing" isn't going to stuff us right there and then, but seeing the `SyncTask` above made me think it could be split up (for example, `RefreshCredentials` taking a `util::Task`, doing its thing async with child tasks, and `PushMetrics` being split at the point it calls `RefreshCredentials`, either setting the second part as the callback for the child task, or calling the second part directly).

Could be a separate PR, but think about it? Maybe file an issue, to remember?

cpp/monitoring/gcm/exporter.cc
```diff
@@ -79,29 +84,35 @@ GCMExporter::~GCMExporter() {
...
+      continue;
```
pphaneuf: Might be good to have a metric for this as well?

cpp/monitoring/gcm/exporter.cc
AlCutter: Added a TODO.

cpp/monitoring/gcm/exporter.cc
```diff
@@ -157,25 +171,34 @@ void GCMExporter::CreateMetrics() const {
...
+      SyncTask task(executor_);
+      VLOG(1) << "Creating metric m.Name()...";
+      VLOG(2) << req.body;
+      fetcher_->Fetch(req, &resp, task.task());
+      task.Wait();
+      if (!task.status().ok() || resp.status_code != 200) {
+        LOG(WARNING) << "Failed to create/update metric metadata; status: "
+                     << task.status()
+                     << ", response_code: " << resp.status_code;
+        sleep(FLAGS_google_compute_monitoring_retry_delay_seconds);
```
pphaneuf: Same kind of story here, both for the `SyncTask`, the `sleep`, and incrementing a counter.

pphaneuf:
Done with this review, sort of no major change needed (as in, I think some *are* needed, but can be in a different PR). :wink:

cpp/monitoring/gcm/exporter.cc
AlCutter: Added

cpp/monitoring/gcm/exporter.cc
AlCutter: done*2

pphaneuf:
LGTM

PR #885, opened 2015-08-27
Start to migrate away from Cert::Status, this is being done incrementally

Martin2112:
Cert::Status is being deprecated in favour of util::Status and util::StatusOr. This converts 3 methods to StatusOr and adds some helpers to support the change.

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -38,6 +38,35 @@ static int X509_get_signature_nid(const X509* x) {
 
 namespace cert_trans {
 
+// Useful because util::Status clashes with Cert::Status so we have to
+// refer to it within the namespace while Cert::Status exists.
+static const util::Status& ERROR_STATUS = util::Status::UNKNOWN;
+
+// TODO. These helpers can be removed when Cert::Status has been removed.
+// They make it easier to do an incremental removal
+
+static Cert::Status StatusOrBoolToCertStatus(StatusOr<bool> status) {
+  if (status.ok()) {
+    return status.ValueOrDie() ? Cert::TRUE : Cert::FALSE;
+  } else {
+    return Cert::ERROR;
+  }
+}
+
+
+static StatusOr<bool> CertStatusToStatusOrBool(Cert::Status status) {
+  if (status == Cert::FALSE) {
+    return StatusOr<bool>(false);
+  } else if (status == Cert::TRUE) {
+    return StatusOr<bool>(true);
+  } else {
+    return StatusOr<bool>(ERROR_STATUS);
+  }
+}
+
+// END of section to be cleaned up
+
+
 // Convert string from ASN1 and check it doesn't contain nul characters
 string ASN1ToStringAndCheckForNulls(ASN1_STRING* asn1_string,
                                     const string& tag,
@@ -289,10 +318,10 @@ Cert::Status Cert::HasBasicConstraintCATrue() const {
 }
 
 
-Cert::Status Cert::HasExtendedKeyUsage(int key_usage_nid) const {
+StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return ERROR_STATUS;
   }
 
   const ASN1_OBJECT* key_usage_obj = OBJ_nid2obj(key_usage_nid);
@@ -300,7 +329,7 @@ Cert::Status Cert::HasExtendedKeyUsage(int key_usage_nid) const {
     LOG(ERROR) << "OpenSSL OBJ_nid2obj returned NULL for NID " << key_usage_nid
                << ". Is the NID not recognised?";
     LOG_OPENSSL_ERRORS(WARNING);
-    return ERROR;
+    return ERROR_STATUS;
   }
 
   void* ext_struct;
@@ -311,8 +340,9 @@ Cert::Status Cert::HasExtendedKeyUsage(int key_usage_nid) const {
     LOG(ERROR) << "Failed to check ExtendedKeyUsage extension";
   }
 
-  if (status != TRUE)
-    return status;
+  if (status != TRUE) {
+    return CertStatusToStatusOrBool(status);
+  }
 
   // |eku| is never NULL upon success.
   EXTENDED_KEY_USAGE* eku = static_cast<EXTENDED_KEY_USAGE*>(ext_struct);
@@ -325,7 +355,7 @@ Cert::Status Cert::HasExtendedKeyUsage(int key_usage_nid) const {
   }
 
   EXTENDED_KEY_USAGE_free(eku);
-  return ext_key_usage_found ? TRUE : FALSE;
+  return StatusOr<bool>(ext_key_usage_found ? true : false);
 }
 
 
@@ -1308,7 +1338,10 @@ Cert::Status PreCertChain::UsesPrecertSigningCertificate() const {
     return Cert::FALSE;
   }
 
-  return issuer->HasExtendedKeyUsage(cert_trans::NID_ctPrecertificateSigning);
+  StatusOr<bool> issuer_eku_status =
+      issuer->HasExtendedKeyUsage(cert_trans::NID_ctPrecertificateSigning);
+
+  return StatusOrBoolToCertStatus(issuer_eku_status);
 }
 
 
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -56,11 +56,11 @@ static Cert::Status StatusOrBoolToCertStatus(StatusOr<bool> status) {
 
 static StatusOr<bool> CertStatusToStatusOrBool(Cert::Status status) {
   if (status == Cert::FALSE) {
-    return StatusOr<bool>(false);
+    return false;
   } else if (status == Cert::TRUE) {
-    return StatusOr<bool>(true);
+    return true;
   } else {
-    return StatusOr<bool>(ERROR_STATUS);
+    return ERROR_STATUS;
   }
 }
 
@@ -296,7 +296,7 @@ Cert::Status Cert::HasCriticalExtension(int extension_nid) const {
 StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return StatusOr<bool>(ERROR_STATUS);
+    return ERROR_STATUS;
   }
 
   void* ext_struct;
@@ -315,14 +315,14 @@ StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
   BASIC_CONSTRAINTS* constraints = static_cast<BASIC_CONSTRAINTS*>(ext_struct);
   bool is_ca = constraints->ca;
   BASIC_CONSTRAINTS_free(constraints);
-  return StatusOr<bool>(is_ca ? true : false);
+  return is_ca ? true : false;
 }
 
 
 StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return StatusOr<bool>(ERROR_STATUS);
+    return ERROR_STATUS;
   }
 
   const ASN1_OBJECT* key_usage_obj = OBJ_nid2obj(key_usage_nid);
@@ -330,7 +330,7 @@ StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
     LOG(ERROR) << "OpenSSL OBJ_nid2obj returned NULL for NID " << key_usage_nid
                << ". Is the NID not recognised?";
     LOG_OPENSSL_ERRORS(WARNING);
-    return StatusOr<bool>(ERROR_STATUS);
+    return ERROR_STATUS;
   }
 
   void* ext_struct;
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -315,7 +315,7 @@ StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
   BASIC_CONSTRAINTS* constraints = static_cast<BASIC_CONSTRAINTS*>(ext_struct);
   bool is_ca = constraints->ca;
   BASIC_CONSTRAINTS_free(constraints);
-  return is_ca ? true : false;
+  return is_ca;
 }
 
 
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -356,7 +356,7 @@ StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
   }
 
   EXTENDED_KEY_USAGE_free(eku);
-  return ext_key_usage_found ? true : false;
+  return ext_key_usage_found;
 }
 
 
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -293,10 +293,10 @@ Cert::Status Cert::HasCriticalExtension(int extension_nid) const {
 }
 
 
-Cert::Status Cert::HasBasicConstraintCATrue() const {
+StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return StatusOr<bool>(ERROR_STATUS);
   }
 
   void* ext_struct;
@@ -307,21 +307,22 @@ Cert::Status Cert::HasBasicConstraintCATrue() const {
     LOG(ERROR) << "Failed to check BasicConstraints extension";
   }
 
-  if (status != TRUE)
-    return status;
+  if (status != TRUE) {
+    return CertStatusToStatusOrBool(status);
+  }
 
   // |constraints| is never NULL upon success.
   BASIC_CONSTRAINTS* constraints = static_cast<BASIC_CONSTRAINTS*>(ext_struct);
   bool is_ca = constraints->ca;
   BASIC_CONSTRAINTS_free(constraints);
-  return is_ca ? TRUE : FALSE;
+  return StatusOr<bool>(is_ca ? true : false);
 }
 
 
 StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR_STATUS;
+    return StatusOr<bool>(ERROR_STATUS);
   }
 
   const ASN1_OBJECT* key_usage_obj = OBJ_nid2obj(key_usage_nid);
@@ -329,7 +330,7 @@ StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
     LOG(ERROR) << "OpenSSL OBJ_nid2obj returned NULL for NID " << key_usage_nid
                << ". Is the NID not recognised?";
     LOG_OPENSSL_ERRORS(WARNING);
-    return ERROR_STATUS;
+    return StatusOr<bool>(ERROR_STATUS);
   }
 
   void* ext_struct;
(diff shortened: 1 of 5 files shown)


cpp/log/cert.cc
```diff
@@ -233,8 +262,8 @@ string Cert::PrintTime(ASN1_TIME* when) {
 }
 
 
-Cert::Status Cert::IsIdenticalTo(const Cert& other) const {
-  return X509_cmp(x509_, other.x509_) == 0 ? TRUE : FALSE;
+StatusOr<bool> Cert::IsIdenticalTo(const Cert& other) const {
+  return StatusOr<bool>(X509_cmp(x509_, other.x509_) == 0 ? true : false);
```
pphaneuf: You normally don't need to construct a temporary, `StatusOr<bool>` has two non-explicit constructors that takes a `util::Status` or `bool`. Stripping out that "stutter" should make things more readable (here, and for every other `return` statements in the file).

Martin2112: Bah! I originally did that then thought "let's make it all explicit". I'll change it.

The main thing that worries me about StatusOr is that you can return StatusOr<int> from a function declared to return StatusOr<bool>, which means I'm going to have to be careful.

cpp/log/cert.cc
```diff
@@ -325,7 +356,7 @@ Cert::Status Cert::HasExtendedKeyUsage(int key_usage_nid) const {
   }
 
   EXTENDED_KEY_USAGE_free(eku);
-  return ext_key_usage_found ? TRUE : FALSE;
+  return ext_key_usage_found ? true : false;
```
pphaneuf: `ext_key_usage_found` is already a `bool`, you can just `return` it. :smiley:

cpp/log/cert.cc
pphaneuf: You're not safe from conversion between integer types, unfortunately. The following all compile, in a function with a `StatusOr<bool>`:

```C++
return true;
return 42;
return StatusOr<bool>(23);
return StatusOr<int>(77);
```

So there's no gain to be made?

Martin2112: I think it would be safer if I don't introduce StatusOr<int> returns. It
would be cleaner to use this as we could get rid of result pointer
parameters but if we don't use multiple integer return types with StatusOr
there's much less chance of accidentally returning the wrong thing. For
these it can just return util::Status.

pphaneuf: It's nothing specific to `StatusOr`, though, if you have a function that returns `bool`, you can `return 42;` just fine (that's "true"), and we can't really say that we can avoid `int` returns. :wink: 

In other words, returning `StatusOr<Foo>` behaves pretty closely to how it would if you had defined the function as returning `Foo`, whatever that may be. And integer types in C++ convert rather merrily between each others (lest too much C compatibility was lost), so that's something we have to live with.

Martin2112: Yes, I'm thinking specifically about the cert.cc code where some operations
call several other methods that could return a mix of StatusOr types it
would be easier to make an error in what then gets returned. I don't know
what's best yet and am still thinking it through.

cpp/log/cert.cc
```diff
@@ -278,21 +307,22 @@ Cert::Status Cert::HasBasicConstraintCATrue() const {
...
+  if (status != TRUE) {
+    return CertStatusToStatusOrBool(status);
+  }
 
   // |constraints| is never NULL upon success.
   BASIC_CONSTRAINTS* constraints = static_cast<BASIC_CONSTRAINTS*>(ext_struct);
   bool is_ca = constraints->ca;
   BASIC_CONSTRAINTS_free(constraints);
-  return is_ca ? TRUE : FALSE;
+  return is_ca ? true : false;
```
pphaneuf: No need for this ternary expression.

cpp/log/cert.cc
```diff
@@ -233,8 +262,8 @@ string Cert::PrintTime(ASN1_TIME* when) {
 }
 
 
-Cert::Status Cert::IsIdenticalTo(const Cert& other) const {
-  return X509_cmp(x509_, other.x509_) == 0 ? TRUE : FALSE;
+StatusOr<bool> Cert::IsIdenticalTo(const Cert& other) const {
+  return X509_cmp(x509_, other.x509_) == 0 ? true : false;
```
pphaneuf: No need for the ternary expression here.

pphaneuf: Also, this could be a proper `bool` instead of a `StatusOr<bool>`, could make other call sites simpler?

Martin2112: Yes, that one doesn't return any errors. Might be better if I change it
just to make that clear.

Martin

pphaneuf: Oh, what the heck, see #893. :smiley:

cpp/log/ct_extensions_test.cc
```diff
@@ -166,16 +166,20 @@ TEST_F(CtExtensionsTest, TestPoisonExtension) {
 TEST_F(CtExtensionsTest, TestPrecertSigning) {
   // Sanity check
   Cert simple_ca_cert(simple_ca_cert_);
-  EXPECT_EQ(Cert::FALSE, simple_ca_cert.HasExtendedKeyUsage(
-                             cert_trans::NID_ctPrecertificateSigning));
+  StatusOr<bool> simple_ca_eku_status = simple_ca_cert.HasExtendedKeyUsage(
+      cert_trans::NID_ctPrecertificateSigning);
+  EXPECT_TRUE(simple_ca_eku_status.ok() &&
+              simple_ca_eku_status.ValueOrDie() == false);
```
pphaneuf: In tests, could probably *use* the implicit `OrDie`, and just go `EXPECT_FALSE(simple_ca_eku_status.ValueOrDie());`?

cpp/log/ct_extensions_test.cc
```diff
@@ -166,16 +166,20 @@ TEST_F(CtExtensionsTest, TestPoisonExtension) {
...
   ASSERT_TRUE(pre_signing_cert.IsLoaded());
   // Check we can find the key usage by its advertised NID.
   // We should really be checking that the OID matches the expected OID but
   // what other key usage could this cert be having that the other one doesn't?
-  ASSERT_EQ(Cert::TRUE, pre_signing_cert.HasExtendedKeyUsage(
-                            cert_trans::NID_ctPrecertificateSigning));
+  StatusOr<bool> pre_signing_eku_status = pre_signing_cert.HasExtendedKeyUsage(
+      cert_trans::NID_ctPrecertificateSigning);
+  ASSERT_TRUE(pre_signing_eku_status.ok() &&
+              pre_signing_eku_status.ValueOrDie());
```
pphaneuf: Same kind of thing here.

cpp/log/cert_test.cc
```diff
@@ -380,11 +380,14 @@ TEST_F(CertTest, Extensions) {
 
   EXPECT_EQ(Cert::TRUE, pre.HasCriticalExtension(cert_trans::NID_ctPoison));
 
-  EXPECT_EQ(Cert::FALSE, leaf.HasBasicConstraintCATrue());
-  EXPECT_EQ(Cert::TRUE, ca.HasBasicConstraintCATrue());
-
-  EXPECT_EQ(Cert::TRUE, ca_pre.HasExtendedKeyUsage(
-                            cert_trans::NID_ctPrecertificateSigning));
+  StatusOr<bool> leaf_ca_status = leaf.HasBasicConstraintCATrue();
+  EXPECT_TRUE(leaf_ca_status.ok() && leaf_ca_status.ValueOrDie() == false);
```
pphaneuf: Same as for my `cpp/log/ct_extensions_test.cc` comment, could let the `OrDie` work its magic in tests, and just `EXPECT_FALSE(leaf_ca_status.ValueOrDie());`? And if it's referenced only once, could use it as it used to be:

```C++
EXPECT_FALSE(leaf.HasBasicConstraintCATrue().ValueOrDie);
```

Similar thing for the checks below...

cpp/log/cert_checker_test.cc
```diff
@@ -336,7 +336,9 @@ TEST_F(CertCheckerTest, AcceptNoBasicConstraintsAndMd2) {
   Cert ca(ca_pem);
   // Verify testdata properties: CA is legacy root.
   ASSERT_EQ("md2WithRSAEncryption", ca.PrintSignatureAlgorithm());
-  ASSERT_EQ(Cert::FALSE, ca.HasBasicConstraintCATrue());
+  StatusOr<bool> has_ca_constraint = ca.HasBasicConstraintCATrue();
+  ASSERT_TRUE(has_ca_constraint.ok() &&
```
pphaneuf: More of the same as my `cpp/log/cert_test.cc` comment here (and the other bit below).

cpp/log/cert_test.cc
Martin2112: Yes wasn't sure about that one. It's easier to read but annoying when tests exit before completion.

cpp/log/cert.cc
```diff
@@ -38,6 +38,35 @@ static int X509_get_signature_nid(const X509* x) {
...
 namespace cert_trans {
 
+// Useful because util::Status clashes with Cert::Status so we have to
+// refer to it within the namespace while Cert::Status exists.
+static const util::Status& ERROR_STATUS = util::Status::UNKNOWN;
+
+// TODO. These helpers can be removed when Cert::Status has been removed.
+// They make it easier to do an incremental removal
+
+static Cert::Status StatusOrBoolToCertStatus(StatusOr<bool> status) {
```
pphaneuf: It's not that bad (just more work for you!), but the existence of this helper makes me notice that you've refactored "from the middle", instead of from the top or the bottom...

It's usually easier to start converting from a leaf function/method (say, `Cert::ExtensionIndex`), and work your way up to its callers, and their callers, etc. Or you can start from the top (as I went in my PRs, since I assumed you were at the bottom, figured we'd meet in the middle!), especially when there are few call sites.

cpp/log/cert.cc
```diff
@@ -38,6 +38,35 @@ static int X509_get_signature_nid(const X509* x) {
 
 namespace cert_trans {
 
+// Useful because util::Status clashes with Cert::Status so we have to
+// refer to it within the namespace while Cert::Status exists.
+static const util::Status& ERROR_STATUS = util::Status::UNKNOWN;
```
pphaneuf: Instead of using `static` for this (and the ones below), use an anonymous namespace. The linker will have some symbols and debug info to work with (which is handier in crashes, since `static` often get you `???` function names), but you get all the benefits (such as the compiler yelling at you when you've cleaned up the last call site).

cpp/log/cert.cc
Martin2112: I did start from the top - the top of the class! I thought I was going to be doing all of it so it wouldn't really matter what order it was done.

cpp/log/cert.cc
```diff
@@ -264,10 +293,10 @@ Cert::Status Cert::HasCriticalExtension(int extension_nid) const {
 }
 
 
-Cert::Status Cert::HasBasicConstraintCATrue() const {
+StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return ERROR_STATUS;
```
pphaneuf: Returning `Status::UNKNOWN`? That's a bit unfriendly, I try to avoid it, as it has no message string... You can also get the benefit of `util::Status` better, letting you have the message be something slightly more specific for human beings, like "certificate not loaded", or "NID not found".

In `CertStatusToStatusOrBool`, you won't have much of a chance, though, but since it might be the only use of `ERROR_STATUS`, you could just use `util::Status::UNKNOWN` directly there?

cpp/log/cert.cc
```diff
@@ -278,21 +307,22 @@ Cert::Status Cert::HasBasicConstraintCATrue() const {
...
+  return is_ca;
 }
 
 
-Cert::Status Cert::HasExtendedKeyUsage(int key_usage_nid) const {
+StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return ERROR_STATUS;
```
pphaneuf: Return a literal `Status()` here (with a message).

cpp/log/cert.cc
```diff
@@ -300,7 +330,7 @@ Cert::Status Cert::HasExtendedKeyUsage(int key_usage_nid) const {
     LOG(ERROR) << "OpenSSL OBJ_nid2obj returned NULL for NID " << key_usage_nid
                << ". Is the NID not recognised?";
     LOG_OPENSSL_ERRORS(WARNING);
-    return ERROR;
+    return ERROR_STATUS;
```
pphaneuf: Return a literal `Status()` here (with a message).

cpp/log/cert.cc
```diff
@@ -278,21 +307,22 @@ Cert::Status Cert::HasBasicConstraintCATrue() const {
     LOG(ERROR) << "Failed to check BasicConstraints extension";
   }
 
-  if (status != TRUE)
-    return status;
+  if (status != TRUE) {
+    return CertStatusToStatusOrBool(status);
```
pphaneuf: Not your fault, but the structure of this function frightens me.

cpp/log/cert.cc
```diff
@@ -883,7 +914,8 @@ Cert::Status Cert::IsValidNameConstrainedIntermediateCa() const {
 
   // If it's not a CA cert or there is no name constraint extension then we
   // don't need to apply the rules any further
-  if (HasBasicConstraintCATrue() != Cert::TRUE
+  StatusOr<bool> has_ca_constraint_status = HasBasicConstraintCATrue();
+  if (StatusOrBoolToCertStatus(has_ca_constraint_status) != Cert::TRUE
```
pphaneuf: You're using your temporary helper just to avoid doing `(!has_ca_constraint_status.ok() || !has_ca_constraint_status.ValueOrDie())` here, but we want to delete that temporary helper, so might as well bite the bullet?

cpp/log/cert.cc
```diff
@@ -883,7 +914,8 @@ Cert::Status Cert::IsValidNameConstrainedIntermediateCa() const {
 
   // If it's not a CA cert or there is no name constraint extension then we
   // don't need to apply the rules any further
-  if (HasBasicConstraintCATrue() != Cert::TRUE
+  StatusOr<bool> has_ca_constraint_status = HasBasicConstraintCATrue();
```
pphaneuf: I found it less messy over time to name `StatusOr<Foo>` variables as if they were just a `Foo` (so `has_ca_constraint` here). Let the type system steady your hand... :smiley:

cpp/log/cert.cc
```diff
@@ -1258,9 +1290,10 @@ Cert::Status CertChain::IsValidCaIssuerChainMaybeLegacyRoot() const {
     // The root cert may not have CA:True
     status = issuer->IsSelfSigned();
     if (status == Cert::FALSE) {
-      Cert::Status s2 = issuer->HasBasicConstraintCATrue();
-      if (s2 != Cert::TRUE)
-        return s2;
+      StatusOr<bool> s2 = issuer->HasBasicConstraintCATrue();
+      if (StatusOrBoolToCertStatus(s2) != Cert::TRUE) {
```
pphaneuf: Again, using the temporary helper isn't really helping?

cpp/log/cert.cc
```diff
@@ -1258,9 +1290,10 @@ Cert::Status CertChain::IsValidCaIssuerChainMaybeLegacyRoot() const {
...
+        return StatusOrBoolToCertStatus(s2);
```
pphaneuf: Especially doing it twice in a row, could have done this instead (styled with my beloved `const` and constructor-style initialisation):

```C++
const Cert::Status s2(StatusOrBoolToCertStatus(issuer->HasBasicConstraintCATrue()));
```

cpp/log/cert.cc
```diff
@@ -1308,7 +1341,10 @@ Cert::Status PreCertChain::UsesPrecertSigningCertificate() const {
     return Cert::FALSE;
   }
 
-  return issuer->HasExtendedKeyUsage(cert_trans::NID_ctPrecertificateSigning);
+  StatusOr<bool> issuer_eku_status =
```
pphaneuf: Can drop the `_status` from the name here, or, even, eliminate the variable (just pass it straight in)?

cpp/log/cert.h
```diff
@@ -8,6 +8,9 @@
 #include <vector>
 
 #include "base/macros.h"
+#include "util/statusor.h"
+
+using util::StatusOr;
```
pphaneuf: Shouldn't use `using` in headers top-levels:

https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Namespaces

pphaneuf:
Ok, FULLY reviewed this now, instead of the stop-and-restart I had been doing! So once all this is addressed, we should be good.

cpp/log/cert.cc
pphaneuf: Haha!

If you start from a function that doesn't return a `Cert::Status`, `StatusOrBoolToCertStatus` wouldn't have needed to exist, and if you start from a function that doesn't call anything returning a `Cert::Status`, you wouldn't have needed `CertStatusToStatusOrBool` (but that can be harder, for example in #892, where I'm having to carry around `GetVerifyError` a bit longer, but fewer and fewer code paths are needing it).

pphaneuf:
BTW, useful bit of info, this here is where the error codes *really* matter, as they get sent to the user:

https://github.com/google/certificate-transparency/blob/master/cpp/server/handler.cc#L122

cpp/log/cert.cc
Martin2112: Well I'm having to work without the tools I'd normally have available to
automate some of this but I did do an analysis before I started.

Because some of the logic is complex and the danger of returning a
StatusOr<bool> instead of int or vice-versa I used the helpers to make the
intermediate state read more like what was there currently.

cpp/log/cert.cc
Martin2112: Yes, the Cert::Status isn't helping either. There are various OpenSSL errors on the way to getting the extension. Might be worth another pass after this is done as we probably don't care exactly how it went wrong in OpenSSL.

cpp/log/cert.cc
Martin2112: Just trying to preserve the structure of the logic for the intermediate state so it reads similarly. I can update it now but thought it would be easier to review.

cpp/log/cert.cc
Martin2112: Noted. BTW I'm used to a type system where bool is not eligible for implicit integer conversion, which steadies hands a good deal more!

cpp/log/cert.cc
pphaneuf: It seems to map directly enough, from "not `Cert::TRUE` to "ok and not `true`" (the checking for `ok()` is really because of the `OrDie` part of `ValueOrDie`, but if you ignore that, it's kind of exactly what it used to say).

cpp/log/cert.cc
Martin2112: It's all going away though fairly shortly.

cpp/log/cert.cc
pphaneuf: Yeah, and for extra fun, any `enum` is also an integral type, so HAHAHAHA! :wink: 

This problem is fixed by using the C++11 `enum class`:

http://en.cppreference.com/w/cpp/language/enum

cpp/log/cert.cc
pphaneuf: Fair enough.

Martin2112:
Is this PR good now? It's really hard to tell with this review system.

pphaneuf:
LGTM

Martin2112:
Thanks. I even rebased it onto master without breaking the universe.

pphaneuf:
Oh, you've never broken the universe that badly... I once rebased the master branch itself by mistake, pushed it, and Al started working in an "alternate universe". :smiley: 

Thanks!

PR #890, opened 2015-08-27
Add conditional compilation for BoringSSL compatibility.

SAHF:
- |Cert::IsSignedBy| tries to infer, from OpenSSL |X509_verify|
  errors, (a) which errors count as |ERROR|, as opposed to mere
  failures to verify, and (b) which errors count as
  |UNSUPPORTED_ALGORITHM|.  This needs conditional compilation, since
  BoringSSL's error reporting is slightly different.

- BoringSSL doesn't define |i2d_re_X509_tbs|, so it should get the
  same substitute function that OpenSSL does.

Along the way, #include <openssl/crypto.h>, for the definition of
|OPENSSL_free|.

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -8,6 +8,7 @@
 #include <glog/logging.h>
 #include <openssl/asn1.h>
 #include <openssl/bio.h>
+#include <openssl/crypto.h>
 #include <openssl/err.h>
 #include <openssl/evp.h>
 #include <openssl/objects.h>
@@ -27,13 +28,15 @@ using util::ClearOpenSSLErrors;
 using util::StatusOr;
 using util::error::Code;
 
-#if OPENSSL_VERSION_NUMBER < 0x10002000L
+#if OPENSSL_VERSION_NUMBER < 0x10002000L || defined(OPENSSL_IS_BORINGSSL)
 // Backport from 1.0.2-beta3.
 static int i2d_re_X509_tbs(X509* x, unsigned char** pp) {
   x->cert_info->enc.modified = 1;
   return i2d_X509_CINF(x->cert_info, pp);
 }
+#endif
 
+#if OPENSSL_VERSION_NUMBER < 0x10002000L
 static int X509_get_signature_nid(const X509* x) {
   return OBJ_obj2nid(x->sig_alg->algorithm);
 }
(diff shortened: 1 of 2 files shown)


cpp/log/cert.cc
```diff
@@ -355,22 +358,36 @@ Cert::Status Cert::IsSignedBy(const Cert& issuer) const {
...
-    int reason = ERR_GET_REASON(err);
-    if (ERR_GET_LIB(err) == ERR_LIB_ASN1 &&
+  if (ret == 1) {
+    return TRUE;
+  }
+  unsigned long err = ERR_peek_last_error();
+  if (!err) {
+    return FALSE;
+  }
+  int reason = ERR_GET_REASON(err);
```
AlCutter: Could you make `err`, `reason`, `lib`, & `unsupported_algorithm` all be `const` while you're here?

SAHF: Done.

cpp/log/cert.cc
```diff
@@ -355,22 +358,36 @@ Cert::Status Cert::IsSignedBy(const Cert& issuer) const {
...
   EVP_PKEY_free(issuer_key);
-  if (ret < 0) {
-    unsigned long err = ERR_peek_last_error();
-    int reason = ERR_GET_REASON(err);
-    if (ERR_GET_LIB(err) == ERR_LIB_ASN1 &&
+  if (ret == 1) {
+    return TRUE;
+  }
+  const unsigned long err = ERR_peek_last_error();
+  if (!err) {
```
eranmes: This is the case where ret != 1 and err is zero, meaning X509_verify "simply" failed to verify without any error, right?
Adding a comment to that effect would be useful.

cpp/log/cert.cc
```diff
@@ -355,22 +358,36 @@ Cert::Status Cert::IsSignedBy(const Cert& issuer) const {
...
+    return FALSE;
+  }
+  const int reason = ERR_GET_REASON(err);
```
eranmes: If I understand correctly at this point you could DCHECK that ret is 0 if OPENSSL_IS_BORINGSSL or ret < 0 for "regular" openssl?

cpp/log/cert.cc
```diff
@@ -355,22 +358,36 @@ Cert::Status Cert::IsSignedBy(const Cert& issuer) const {
...
-    }
+         reason == ASN1_R_UNKNOWN_SIGNATURE_ALGORITHM);
+#endif
+  if (unsupported_algorithm) {
+    LOG(WARNING) << "Unsupported algorithm: " << PrintSignatureAlgorithm();
+    ClearOpenSSLErrors();
+    return UNSUPPORTED_ALGORITHM;
+  }
+  if (lib == ERR_LIB_ASN1 || lib == ERR_LIB_X509) {
+    ClearOpenSSLErrors();  // "Normal" error, not worth logging.
```
eranmes: Nit: Clarify the comment to explain what are "normal" errors in this context.

eranmes:
LGTM with a few minor comments about comments.

pphaneuf:
And it will need a rebase on top of the latest master.

cpp/log/cert.cc
SAHF: Done.

cpp/log/cert.cc
SAHF: I'd like to keep it easy to read across the conditional compilation, so I've made the conditional larger rather than having two of them.  I think it's clearer this way; let me know what you think.

cpp/log/cert.cc
SAHF: Done.

eranmes:
Latest revision LGTM, will merge as soon as travis is happy.

PR #891, opened 2015-08-28
Replace CertSubmissionHandler::SubmitResult with util::Status.

pphaneuf:
From the top, allowing to make sure the `util::Status` that comes out is as callers received in the same situation more easily.

### DIFF
--- a/cpp/log/cert_submission_handler.cc
+++ b/cpp/log/cert_submission_handler.cc
@@ -18,6 +18,30 @@ using ct::LogEntry;
 using ct::PrecertChainEntry;
 using ct::X509ChainEntry;
 using std::string;
+using util::Status;
+
+namespace {
+
+Status GetVerifyError(CertChecker::CertVerifyResult result) {
+  switch (result) {
+    case CertChecker::INVALID_CERTIFICATE_CHAIN:
+    case CertChecker::PRECERT_EXTENSION_IN_CERT_CHAIN:
+    case CertChecker::UNSUPPORTED_ALGORITHM_IN_CERT_CHAIN:
+      return Status(util::error::INVALID_ARGUMENT,
+                    "invalid certificate chain");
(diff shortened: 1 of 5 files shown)


cpp/log/cert_submission_handler.cc
```diff
@@ -18,6 +18,30 @@ using ct::LogEntry;
...
 using std::string;
+using util::Status;
+
+namespace {
+
+Status GetVerifyError(CertChecker::CertVerifyResult result) {
+  switch (result) {
+    case CertChecker::INVALID_CERTIFICATE_CHAIN:
+    case CertChecker::PRECERT_EXTENSION_IN_CERT_CHAIN:
+    case CertChecker::UNSUPPORTED_ALGORITHM_IN_CERT_CHAIN:
```
Martin2112: I wondered if we needed to break out UNSUPPORTED_ALGORITHM separately, possibly add it to util::Status? A couple of places check for this explicitly, though it's possible they then return a generic error anyway.

cpp/log/cert_submission_handler.cc
```diff
@@ -112,15 +134,15 @@ CertSubmissionHandler::ProcessPreCertSubmission(PreCertChain* chain,
   string der_cert;
   // Nothing should fail anymore as we have validated the chain.
   if (chain->LeafCert()->DerEncoding(&der_cert) != Cert::TRUE)
-    return INTERNAL_ERROR;
+    return Status(util::error::INTERNAL, "could not DER-encode the chain");
```
Martin2112: Could pull this error out to a constant as it's used multiple times.

pphaneuf: I wouldn't really bother, since `DerEncoding` is going to return a `util::Status` soon, so it'll just be a "classic" `Status status(thing()); if (!status.ok) return status;`.

PR #892, opened 2015-08-28
Migrate CertChecker::Check{Pre,}CertChain to util::Status.

pphaneuf:
Keep pushing down from the top!

Keeping the `util::Status` as "near literals", to keep things clear and explicit (ideally, the messages would all be different, but this should improve as we convert more). There's some repetition, but as more is converted, it'll go away (just returning the underlying status in case of error).

### DIFF
--- a/cpp/log/cert_checker.cc
+++ b/cpp/log/cert_checker.cc
@@ -20,8 +20,32 @@
 using std::string;
 using std::vector;
 using util::ClearOpenSSLErrors;
+using util::Status;
 
 namespace cert_trans {
+
+namespace {
+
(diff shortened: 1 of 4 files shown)


pphaneuf:
Not sure if those shouldn't be using `StatusOr<bool>` instead, to return `false` when a chain isn't good due to error on the submitter's part, and a non-OK status when there's some sort of internal error that might be worth logging?

But I didn't want to change the code too much (I've left all the logging, for example), this could be improved upon later on, as necessary?

Martin2112:
Yes it's not always obvious. The Cert class was using pretty much tri-state logic with Cert::Status TRUE / FALSE / ERROR so it made sense to me to switch to Status and a bool.

PR #895, opened 2015-08-28
Use unique_ptr to manage a temporary reference.

pphaneuf:
This isn't C! There are rules!

### DIFF
--- a/cpp/log/cert_checker.cc
+++ b/cpp/log/cert_checker.cc
@@ -2,6 +2,7 @@
 #include "log/cert_checker.h"
 
 #include <glog/logging.h>
+#include <memory>
 #include <openssl/asn1.h>
 #include <openssl/bio.h>
 #include <openssl/pem.h>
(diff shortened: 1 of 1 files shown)


cpp/log/cert_checker.cc
```diff
@@ -100,20 +102,17 @@ bool CertChecker::LoadTrustedCertificatesFromBIO(BIO* bio_in) {
...
         error = true;
         break;
       }
 
       ++cert_count;
-      if (is_trusted == OK) {
-        delete cert;
-      } else {
-        certs_to_add.push_back(make_pair(subject_name, cert));
+      if (is_trusted != OK) {
```
Martin2112: The code in this function isn't easy to read and reason about. I don't think it's worth restructuring it now though.

pphaneuf: Yeah, the whole thing is a bit wacky. I didn't even change the `x509 != NULL` to a `!x509`, nor add any `const`, or all the rest of the things I usually do, but I was most frightened by the local heap allocation and all the branches...

PR #896, opened 2015-08-28
EV Whitelist: Less restrictive generation.

eranmes:
Some certificates that should have been included in the whitelist
were excluded.
This is fixed by including in the whitelist every certificate with
an EV OID, regardless of the root it chains to.

See https://code.google.com/p/chromium/issues/detail?id=524635 for
more details.


eranmes:
FYI I've re-generated the whitelist and verified the certificate reported missing in the Chromium issue is present.

pphaneuf:
Merged as 0019a23c4a400bb400da1e09dbedbe973ef88994.

I'm having a straight line going, I'm keeping it a bit longer! :stuck_out_tongue:

PR #897, opened 2015-08-28
Migrate CertChecker::GetTrustedCa to util::Status.

pphaneuf:


cpp/log/cert_checker.cc
```diff
@@ -283,18 +283,18 @@ Status CertChecker::CheckPreCertChain(PreCertChain* chain,
   return Status::OK;
 }
 
-CertChecker::CertVerifyResult CertChecker::GetTrustedCa(
-    CertChain* chain) const {
+Status CertChecker::GetTrustedCa(CertChain* chain) const {
   const Cert* subject = chain->LastCert();
   if (subject == NULL || !subject->IsLoaded()) {
```
Martin2112: Might as well be nullptr

pphaneuf: I was leaving it alone, but since you mentioned it, I went to town on the use of `NULL`. :smiley:

PR #898, opened 2015-08-28
Fix an incorrect dependency.

pphaneuf:
Got annoyed at the spewing, was going to fix it, but at least, if it only happens once in a while...


pphaneuf:
Explicit and complete dependencies here might be better, though, but I'm too lazy to fix this now. :wink:

PR #899, opened 2015-08-29
In-progress blob server.

virgo075:


pphaneuf:
I think this is a spam PR?

PR #901, opened 2015-09-01
EV whitelist: Update CRX URL

eranmes:
Point to the latest version (7) of the EV whitelist for the code.
Useful if we'll ever need to re-generate the whitelist.


pphaneuf:
LGTM

PR #902, opened 2015-09-02
Log client retries

eranmes:


eranmes:
PTAL, addressed your comments.

PR #903, opened 2015-09-02
Migrate CertChecker::IsTrusted to util::StatusOr<bool>.

pphaneuf:


Martin2112:
Looks simpler. We're losing the exact reason the chain didn't verify but I don't think we do anything with that anyway.

pphaneuf:
Programmatically, yes, but the error messages are a bit more detailed. Also, I was surprised that there's a bunch of values for the `enum` that simply weren't used? Like `UNSUPPORTED_ALGORITHM_IN_CERT_CHAIN`, for example...

Martin2112:
The comments in the file implied that the different return values were
planned to be used but this never happened. I think it's better to clean
this stuff up because people then don't have to read it and work that out.

pphaneuf:
Agreed, this should have been a case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), having unused values like that is just setting up someone for disappointment.

PR #904, opened 2015-09-02
Migrate Cert::ExtensionIndex to util::Status.

pphaneuf:


cpp/log/cert.cc
```diff
@@ -274,8 +274,16 @@ Cert::Status Cert::HasExtension(int extension_nid) const {
     return ERROR;
   }
 
-  int ignored;
-  return ExtensionIndex(extension_nid, &ignored);
+  const StatusOr<int> index(ExtensionIndex(extension_nid));
+  if (index.ok()) {
+    return TRUE;
```
Martin2112: If this is intentionally returning a Cert::Status as int can you add a comment?

cpp/log/cert.cc
```diff
@@ -606,27 +613,27 @@ Cert::Status Cert::ExtensionIndex(int extension_nid,
...
 
 Cert::Status Cert::GetExtension(int extension_nid,
                                 X509_EXTENSION** ext) const {
-  int extension_index;
-  Status status = ExtensionIndex(extension_nid, &extension_index);
-  if (status != TRUE) {
-    return status;
+  const StatusOr<int> extension_index(ExtensionIndex(extension_nid));
+  if (!extension_index.ok()) {
+    return extension_index.status().CanonicalCode() == util::error::NOT_FOUND
```
Martin2112: I've used error_code() in places. Should I have used this?

pphaneuf: `CanonicalCode()` is preferable, as it returns the `enum` (`error_code()` simply returns `int`). This can make a difference in a `switch` statement (where compiler can catch unhandled `enum` values), but otherwise, not that much.

cpp/log/cert.cc
pphaneuf: Huh? This method has a `Cert::Status` return type, and is returning `Cert::TRUE`, I'm not sure I see where the `int` comes in?

Martin2112: Sorry, I misread the diff as a declaration. Nothing to see here.

PR #905, opened 2015-09-02
Update isIssuedBy and isSignedBy to use util::StatusOr

Martin2112:
And corresponding change to isSelfSigned + update affected tests.

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -370,31 +370,30 @@ StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
 }
 
 
-Cert::Status Cert::IsIssuedBy(const Cert& issuer) const {
+StatusOr<bool> Cert::IsIssuedBy(const Cert& issuer) const {
   if (!IsLoaded() || !issuer.IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
   }
-  int ret = X509_check_issued(const_cast<X509*>(issuer.x509_), x509_);
-  // Seemingly no negative "real" error codes are returned from here.
-  return ret == X509_V_OK ? TRUE : FALSE;
+  // Seemingly no negative "real" error codes are returned from openssl api.
+  return X509_check_issued(const_cast<X509*>(issuer.x509_), x509_) == X509_V_OK;
 }
 
 
-Cert::Status Cert::IsSignedBy(const Cert& issuer) const {
+StatusOr<bool> Cert::IsSignedBy(const Cert& issuer) const {
   if (!IsLoaded() || !issuer.IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
   }
 
-  EVP_PKEY* issuer_key = X509_get_pubkey(issuer.x509_);
-  if (!issuer_key) {
+  EVP_PKEY* const issuer_key = X509_get_pubkey(issuer.x509_);
+  if (issuer_key == NULL) {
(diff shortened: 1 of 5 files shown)


cpp/log/cert.cc
```diff
@@ -402,14 +401,14 @@ Cert::Status Cert::IsSignedBy(const Cert& issuer) const {
          reason == ASN1_R_UNKNOWN_SIGNATURE_ALGORITHM)) {
       LOG(WARNING) << "Unsupported algorithm: " << PrintSignatureAlgorithm();
       ClearOpenSSLErrors();
-      return UNSUPPORTED_ALGORITHM;
+      return util::Status(Code::UNAVAILABLE, "Unsupported algorithm");
```
pphaneuf: Remember to make it `UNIMPLEMENTED` (and the other few spots).

cpp/log/cert_checker.cc
```diff
@@ -306,26 +307,27 @@ Status CertChecker::GetTrustedCa(CertChain* chain) const {
        it != issuer_range.second; ++it) {
     const Cert* issuer_cand = it->second;
 
-    Cert::Status ok = subject->IsSignedBy(*issuer_cand);
-    if (ok == Cert::UNSUPPORTED_ALGORITHM) {
+    StatusOr<bool> signed_by_issuer = subject->IsSignedBy(*issuer_cand);
+    if (signed_by_issuer.status().error_code() == Code::UNAVAILABLE) {
```
pphaneuf: Use `CanonicalCode()` instead of `error_code()`?

cpp/log/cert_test.cc
```diff
@@ -351,7 +352,7 @@ TEST_F(CertTest, TestUnsupportedAlgorithm) {
 // see anything else. Make the test fail if this is not the case to notify the
 // user that their setup is insecure.
 #ifdef OPENSSL_NO_MD2
-  EXPECT_EQ(Cert::UNSUPPORTED_ALGORITHM, legacy.IsSignedBy(legacy));
+  EXPECT_EQ(Code::UNAVAILABLE, legacy.IsSignedBy(legacy).status().error_code());
```
pphaneuf: For `util::Status` tests, you can use `EXPECT_THAT`, with the `StatusIs` helper (from `cpp/util/status_test_util.h`, something like this:

```C++
EXPECT_THAT(legacy.IsSignedBy(legacy).status(), StatusIs(Code::UNAVAILABLE));
```

Yeah, they flip the expected/actual with `EXPECT_THAT`. Gotta keep it exciting! :smiley:

pphaneuf:
All reviewed, mostly nits, really.

Martin2112:
I think I've fixed everything. Also rebased and fixed up conflicts with other refactorings. It's starting to pay off a bit e.g look at CertChain::IsValidSignatureChain.

cpp/log/cert_checker.cc
```diff
@@ -292,26 +293,27 @@ Status CertChecker::GetTrustedCa(CertChain* chain) const {
        it != issuer_range.second; ++it) {
     const Cert* issuer_cand = it->second;
 
-    Cert::Status ok = subject->IsSignedBy(*issuer_cand);
-    if (ok == Cert::UNSUPPORTED_ALGORITHM) {
+    StatusOr<bool> signed_by_issuer = subject->IsSignedBy(*issuer_cand);
+    if (signed_by_issuer.status().CanonicalCode() == Code::UNIMPLEMENTED) {
```
pphaneuf: I wonder if we shouldn't just pass back the underlying status, or if in fact, this sort of translation is good (to control better what the end user see in the HTTP responses)?

pphaneuf:
LGTM

cpp/log/cert_checker.cc
Martin2112: It's not an easy one to answer. In some cases the low level error can be
translated to something more meaningful at a higher level. OTOH you can end
up losing the actual cause of the problem by returning something generic.

Issue #900, opened 2015-09-01
SCTs retrieved by the cpp/fetcher/... code should be verified.


AlCutter:
When fetching entries in either mirror- or log-mode, we don't verify the SCT signature.
Aside from anything else, this is quite important in log-mode in particular as the internal replication traffic is sent over HTTP and so this signature check provides a guard against a packet corruption which may evade TCP's 16bit CRC.

PR #907, opened 2015-09-02
Verify SCTs we get from other nodes, and verify the STH we get from the DB.

AlCutter:
Addresses #900

### DIFF
--- a/cpp/fetcher/continuous_fetcher.cc
+++ b/cpp/fetcher/continuous_fetcher.cc
@@ -5,6 +5,7 @@
 
 #include "fetcher/fetcher.h"
 #include "fetcher/peer_group.h"
+#include "log/log_verifier.h"
 
 using std::bind;
 using std::chrono::seconds;
@@ -30,7 +31,8 @@ namespace {
 class ContinuousFetcherImpl : public ContinuousFetcher {
  public:
   ContinuousFetcherImpl(libevent::Base* base, Executor* executor,
-                        Database<LoggedCertificate>* db, bool fetch_scts);
+                        Database<LoggedCertificate>* db,
+                        const LogVerifier* log_verifier, bool fetch_scts);
 
   void AddPeer(const string& node_id, const shared_ptr<Peer>& peer) override;
   void RemovePeer(const string& node_id);
@@ -43,6 +45,7 @@ class ContinuousFetcherImpl : public ContinuousFetcher {
   libevent::Base* const base_;
   Executor* const executor_;
   Database<LoggedCertificate>* const db_;
+  const LogVerifier* const log_verifier_;
   const bool fetch_scts_;
 
   mutex lock_;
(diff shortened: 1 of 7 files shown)


cpp/fetcher/continuous_fetcher.cc
```diff
@@ -30,7 +31,8 @@ namespace {
 class ContinuousFetcherImpl : public ContinuousFetcher {
  public:
   ContinuousFetcherImpl(libevent::Base* base, Executor* executor,
-                        Database<LoggedCertificate>* db, bool fetch_scts);
+                        Database<LoggedCertificate>* db,
+                        const LogVerifier* log_verifier, bool fetch_scts);
```
eranmes: What are the lifetime requirements of this pointer?

pphaneuf: Based on it not being a `unique_ptr<const LogVerifier>`, I'd say the caller keeps ownership. :smiley:

cpp/fetcher/fetcher.cc
```diff
@@ -74,9 +88,11 @@ struct FetchState {
 
 
 FetchState::FetchState(Database<LoggedCertificate>* db,
-                       unique_ptr<PeerGroup>&& peer_group, Task* task)
+                       unique_ptr<PeerGroup>&& peer_group,
```
pphaneuf: That reminds me... I've read about "how is a `unique_ptr` passed in" recently, and the consensus seems to be that you should prefer taking it by value (which is fine, because `unique_ptr` itself is move-constructible), because that guarantees that it's moved.

Taking by rvalue-ref is suggested for the cases where the ownership *might* be taken, depending on what code path is taken (doing `std::move(foo)` by itself does *nothing*, as you may remember), but this is generally thought of to be slightly weird (after the call, the pointer may or may not be valid), and to be avoided.

cpp/fetcher/fetcher.cc
```diff
@@ -226,10 +242,28 @@ void FetchState::WriteToDatabase(int64_t index, Range* range,
...
       LOG(WARNING) << "could not convert entry to a LoggedCertificate";
+      num_invalid_entries_fetched->Increment("format");
       break;
     }
     if (entry.sct) {
       *cert.mutable_sct() = *entry.sct;
+      // If we have the full SCT (because this LogEntry came from another
+      // internal node which supports our private "give me the SCT too"
+      // option), then verify that the signature is good.
+      const LogVerifier::VerifyResult verify_result(
```
pphaneuf: Note to self: `util::Status` this!

cpp/fetcher/fetcher.cc
```diff
@@ -274,13 +308,13 @@ void FetchState::WriteToDatabase(int64_t index, Range* range,
 
 }  // namespace
 
-namespace cert_trans {
-
 
 void FetchLogEntries(Database<LoggedCertificate>* db,
-                     unique_ptr<PeerGroup>&& peer_group, Task* task) {
+                     unique_ptr<PeerGroup>&& peer_group,
```
pphaneuf: Same thing as above, should drop the `&&` here.

pphaneuf:
Other than the slight tweaking of `unique_ptr` constructor parameter, LGTM.

cpp/fetcher/fetcher.cc
AlCutter: You're talking about the `unique_ptr<PeerGroup>`? That one's yours  :)

I'd rather keep this PR about SCT/STH verification and do cleanup in another...

cpp/fetcher/fetcher.cc
AlCutter: Note to you added ;)

cpp/fetcher/fetcher.cc
pphaneuf: And reviewing #911 shall be yours, then!

cpp/fetcher/fetcher.cc
```diff
@@ -226,10 +242,29 @@ void FetchState::WriteToDatabase(int64_t index, Range* range,
...
       LOG(WARNING) << "could not convert entry to a LoggedCertificate";
+      num_invalid_entries_fetched->Increment("format");
       break;
     }
     if (entry.sct) {
       *cert.mutable_sct() = *entry.sct;
+      // If we have the full SCT (because this LogEntry came from another
+      // internal node which supports our private "give me the SCT too"
+      // option), then verify that the signature is good.
+      // TODO(pphaneuf): Note to self: util::Status this!
```
pphaneuf: That's not what I meant. :stuck_out_tongue:

AlCutter: I know ;)

PR #909, opened 2015-09-02
Migrate Cert::GetExtension to util::Status.

pphaneuf:
Keep on iteratin'...

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -20,6 +20,7 @@
 #include <vector>
 
 using std::string;
+using std::to_string;
 using std::unique_ptr;
 using std::vector;
 using util::ClearOpenSSLErrors;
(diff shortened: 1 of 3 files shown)


cpp/log/cert.cc
```diff
@@ -860,17 +862,19 @@ Cert::Status Cert::IsValidWildcardRedaction() const {
 
   // If we reach here then the RFC says the CT redaction count extension
   // MUST BE present.
-  X509_EXTENSION* exty;
-  if (GetExtension(NID_ctPrecertificateRedactedLabelCount,
-                   &exty) != Cert::TRUE) {
+  const StatusOr<X509_EXTENSION*> exty(
+      GetExtension(NID_ctPrecertificateRedactedLabelCount));
+  if (!exty.ok()) {
     LOG(WARNING) << "Required CT redaction count extension missing from cert";
```
Martin2112: Have we combined the absent / error case here? If so the error message may not be accurate.

pphaneuf: Temporarily so. Here, the existing condition was `!= Cert::TRUE`, so the behaviour should be exactly the same.

In a follow-up PR, line 869 will become `return exty.status();`, or something of that style, passing on whatever error information we might have upstream.

Maybe the log message could be better worded as "extension could not be found in cert"? As in, maybe it's there, but something prevented us from getting it?

As I've mentioned before, I don't really like those log messages anyway, it's just completely unactionable blabbering, caused by data in requests that we don't give in the log, and are probably 100% not our fault.

pphaneuf: See 39216bd0ea38cc066700179a103b5a8e6b816cbf for a preview. :smiley:

Martin2112: OK. I more or less agree on the usefulness logs. Maybe reword that message but I don't think it's essential.

PR #910, opened 2015-09-02
Fix use-after-free in remote_peer.

AlCutter:
The d'tor tries to send a `CANCELLED` status, but `impl_` is already deleted if a `Status` has been returned elsewhere.


pphaneuf:
While reviewing #975, I finally remembered why this was wrong! And the good news is that you fixed it *correctly* in #975, so thanks! :smiley:

PR #912, opened 2015-09-03
Convert a bunch of accessors to use util::Status and StatusOr.

Martin2112:
Update all affected code and tests. Please double check the CHECK conditionals in ssl_client that were using tristate logic before. Thanks!

### DIFF
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -221,7 +221,7 @@ static bool PrecertChainToEntry(const cert_trans::PreCertChain& chain,
 
   entry->set_type(ct::PRECERT_ENTRY);
   string key_hash;
-  if (chain.CertAt(1)->SPKISha256Digest(&key_hash) != util::Status::OK) {
+  if (!chain.CertAt(1)->SPKISha256Digest(&key_hash).ok()) {
     LOG(ERROR) << "Failed to get SPKISha256.";
     return false;
   }
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -816,10 +816,10 @@ void WrapEmbedded() {
                cert_trans::NID_ctEmbeddedSignedCertificateTimestampList));
 
   string serialized_scts;
-  CHECK_EQ(true,
+  CHECK_EQ(::util::Status::OK,
            chain.LeafCert()->OctetStringExtensionData(
                cert_trans::NID_ctEmbeddedSignedCertificateTimestampList,
-               &serialized_scts).ok());
+               &serialized_scts));
   SignedCertificateTimestampList sct_list;
   CHECK_EQ(Deserializer::OK,
            Deserializer::DeserializeSCTList(serialized_scts, &sct_list));
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -729,11 +729,11 @@ static void DiagnoseCertChain() {
   }
 
   string serialized_scts;
-  StatusOr<bool> status = chain.LeafCert()->OctetStringExtensionData(
+  util::Status status = chain.LeafCert()->OctetStringExtensionData(
       cert_trans::NID_ctEmbeddedSignedCertificateTimestampList,
       &serialized_scts);
-  if (!status.ok() || !status.ValueOrDie()) {
-    LOG(ERROR) << "SCT extension data is invalid.";
+  if (!status.ok()) {
+    LOG(ERROR) << "SCT extension data is missing / invalid.";
     return;
   }
 
@@ -817,11 +817,9 @@ void WrapEmbedded() {
 
   string serialized_scts;
   CHECK_EQ(true,
-           chain.LeafCert()
-               ->OctetStringExtensionData(
-                   cert_trans::NID_ctEmbeddedSignedCertificateTimestampList,
-                   &serialized_scts)
-               .ValueOrDie());
+           chain.LeafCert()->OctetStringExtensionData(
+               cert_trans::NID_ctEmbeddedSignedCertificateTimestampList,
+               &serialized_scts).ok());
   SignedCertificateTimestampList sct_list;
   CHECK_EQ(Deserializer::OK,
            Deserializer::DeserializeSCTList(serialized_scts, &sct_list));
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -221,7 +221,7 @@ static bool PrecertChainToEntry(const cert_trans::PreCertChain& chain,
 
   entry->set_type(ct::PRECERT_ENTRY);
   string key_hash;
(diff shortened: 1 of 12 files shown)


cpp/client/ct.cc
```diff
@@ -221,7 +221,7 @@ static bool PrecertChainToEntry(const cert_trans::PreCertChain& chain,
 
   entry->set_type(ct::PRECERT_ENTRY);
   string key_hash;
-  if (chain.CertAt(1)->SPKISha256Digest(&key_hash) != Cert::TRUE) {
+  if (chain.CertAt(1)->SPKISha256Digest(&key_hash) != util::Status::OK) {
```
pphaneuf: In a `CHECK`, comparing against `OK` is useful, because when it's not true, `CHECK_EQ` will dump what it got instead, but here, you could just use `chain.CertAt(1)->SPKISha256Digest(&key_hash).ok()`.

cpp/log/cert.cc
```diff
@@ -550,17 +550,17 @@ Cert::Status Cert::PublicKeySha256Digest(string* result) const {
...
+    return util::Status(Code::INVALID_ARGUMENT, "SHA256 digest failed");
   }
   result->assign(reinterpret_cast<char*>(digest), len);
-  return TRUE;
+  return util::Status::OK;
 }
 
 
-Cert::Status Cert::SPKISha256Digest(string* result) const {
+util::Status Cert::SPKISha256Digest(string* result) const {
```
pphaneuf: Reading that method signature, it looks to me like it really wants to be `StatusOr<string> Cert::SPKISha256Digest() const`, wouldn't you say?

cpp/log/cert.cc
```diff
@@ -578,28 +578,28 @@ Cert::Status Cert::SPKISha256Digest(string* result) const {
...
   result->assign(sha256_digest);
   OPENSSL_free(der_buf);
-  return TRUE;
+  return util::Status::OK;
 }
 
-
-Cert::Status Cert::OctetStringExtensionData(int extension_nid,
-                                            string* result) const {
+StatusOr<bool> Cert::OctetStringExtensionData(int extension_nid,
```
pphaneuf: For this one, `false` isn't really used, `true` really means `OK`, and we might want to return `StatusOr<string>`?

cpp/log/cert.cc
Martin2112: I considered it. I think we should refactor all the similar patterns separately after Cert::Status has gone as the changes are already quite large. I'm open to changing my mind.

pphaneuf:
I was skimming this PR more or less from the top, but when I hit `cert.h`, I saw all those `Status GetThing(string* result) const`, and thought, um, well... :smiley: 

Those should be `StatusOr<string> GetThing() const`, unless there's a good reason not to?

Martin2112:
I agree we should make that change and it will improve the code. I'm suggesting deferring it as it's not necessary to get rid of Cert::Status, which is the short term aim. Still open to reconsidering.

Martin2112:
Is this OK or did we decide to do the further refactoring now?

cpp/log/cert_checker_test.cc
```diff
@@ -273,7 +273,8 @@ TEST_F(CertCheckerTest, PreCert) {
   checker_.LoadTrustedCertificates(cert_dir_ + "/" + kCaCert);
   EXPECT_OK(checker_.CheckPreCertChain(&chain, &issuer_key_hash, &tbs));
   string expected_key_hash;
-  ASSERT_EQ(Cert::TRUE, chain.CertAt(1)->SPKISha256Digest(&expected_key_hash));
+  ASSERT_EQ(util::Status::OK,
```
pphaneuf: There's `ASSERT_OK` and `EXPECT_OK` in `util/status_test_util.h`, could you use those instead?

pphaneuf:
For the bulk of them with the `Status GetFoo(string* result)` pattern, we could do this for now, they're still quite recognisable for later refactoring.

For `OctetStringExtensionData`, it's been done slightly different from the others, and maybe it could be changed to match, using `NOT_FOUND` for the case where it returns `Cert::Status::FALSE`?

Martin2112:
All fixed.

cpp/client/ct.cc
```diff
@@ -815,10 +816,10 @@ void WrapEmbedded() {
                cert_trans::NID_ctEmbeddedSignedCertificateTimestampList));
 
   string serialized_scts;
-  CHECK_EQ(Cert::TRUE,
+  CHECK_EQ(true,
```
pphaneuf: Should be a `util::Status::OK`.

cpp/log/cert_test.cc
```diff
@@ -304,7 +304,7 @@ TEST_F(CertTest, LoadInvalid) {
 TEST_F(CertTest, LoadValidFromDer) {
   Cert leaf(leaf_pem_);
   string der;
-  ASSERT_EQ(Cert::TRUE, leaf.DerEncoding(&der));
+  ASSERT_EQ(util::Status::OK, leaf.DerEncoding(&der));
```
pphaneuf: There's plenty of potential `ASSERT_OK` in this file...

cpp/log/frontend_test.cc
```diff
@@ -165,7 +165,7 @@ TYPED_TEST(FrontendTest, TestSubmitValid) {
   Cert cert(this->leaf_pem_);
 
   string sha256_digest;
-  ASSERT_EQ(Cert::TRUE, cert.Sha256Digest(&sha256_digest));
+  ASSERT_EQ(util::Status::OK, cert.Sha256Digest(&sha256_digest));
```
pphaneuf: There's plenty of potential `ASSERT_OK` in this file too...

cpp/client/ct.cc
Martin2112: I'm checking .ok() though, not the status value.

pphaneuf:
LGTM, assuming you fix those little things, rebase it to master (to make sure there's no weird interaction, as some things changed in the last few commits) and see it pass on Travis.

cpp/client/ct.cc
pphaneuf: Yeah, saw that, but this looks like a conversion from having been `ValueOrDie()`? It would be equivalent and more friendly if it was `CHECK_EQ(util::Status::OK, ...)` instead.

pphaneuf:
I'm not that worried about those last few changes, though... None of them touch the actual `ct-server`/`ct-mirror` binaries, which are my biggest worries, in terms of correctness. The change in `ct.cc` is just for being generally helpful, and the `ASSERT_OK` stuff is merely code readability, with no effect whatsoever.

I'd live if this was pushed as-is. :wink:

Martin2112:
Don't worry, it's all being fixed. Just waiting for tests to run.

PR #913, opened 2015-09-03
Add STH signature verification

AlCutter:
(and tidy up the existing tests a touch)

### DIFF
--- a/go/serialization_test.go
+++ b/go/serialization_test.go
@@ -158,15 +158,15 @@ func TestCheckExtensionsFormatTooBig(t *testing.T) {
 }
 
 const (
-	DefaultSCTLogIDString          string = "iamapublickeyshatwofivesixdigest"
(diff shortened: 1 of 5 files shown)


go/serialization_test.go
```diff
@@ -203,112 +203,151 @@ const (
...
+		// tree size, 8 bytes
+		"0000000000000006" +
+		// root hash, 32 bytes
+		"696d757374626565786163746c7974686972747974776f62797465736c6f6e67"
 )
 
-func DefaultSCTLogID() []byte {
-	return []byte(DefaultSCTLogIDString)
+func defaultSCTLogID() []byte {
+	return []byte(defaultSCTLogIDString)
```
benlaurie: Is it not possible to do the cast while defining the const? (Asked from a position of complete ignorance)

AlCutter: Sadly not; `[]byte` isn't a valid constant type.

PR #914, opened 2015-09-03
Pass in all known etcd servers to log/mirror binaries.

AlCutter:


pphaneuf:
LGTM.

Did you hit an issue in production, or is this preventative?

AlCutter:
Just finishing off wiring up the flag since the binaries now support multiple etcds.

PR #915, opened 2015-09-03
Migrate CertChain::IsValidSignatureChain to util::Status.

pphaneuf:


cpp/log/cert.cc
```diff
@@ -1319,22 +1319,42 @@ Cert::Status CertChain::IsValidCaIssuerChainMaybeLegacyRoot() const {
...
+                            "invalid certificate chain");
+      case Cert::UNSUPPORTED_ALGORITHM:
+        // UNSUPPORTED_ALGORITHM can happen when a weak algorithm
+        // (such as MD2) is intentionally not accepted in which case
+        // it's correct to say that the chain is invalid.
+        // It can also happen when EVP is not properly initialized, in
+        // which case it's more of an INTERNAL_ERROR. However a bust
+        // setup would manifest itself in many other ways, including
+        // failing tests, so we assume the failure is intentional.
+        return util::Status(util::error::UNIMPLEMENTED,
```
pphaneuf: Note that this used to be converted to `INVALID_ARGUMENT`, but since none of the current callers actually cared (as long as it's not `ALREADY_EXISTS`), I figured I might as well?

Martin2112: I think I used UNAVAILABLE for this status in my PR. Either works but we should use the same everywhere.

pphaneuf: With `UNAVAILABLE`, there's an implication of being able to be retried:

https://github.com/google/certificate-transparency/blob/master/cpp/util/status.h#L70

I kind of liked `FAILED_PRECONDITION`, but it's used above. :stuck_out_tongue:

Martin2112: I'm OK with unimplemented, I think it's closer to the right meaning.

Martin

PR #916, opened 2015-09-03
Convert Cert::Status to an "enum class".

pphaneuf:
Simple `enum`s are just too dangerous... While they were just sitting there, presumably well reviewed, it was okay, but now that we're poking the wasp nest, I think this hack is a good idea.

### DIFF
--- a/cpp/log/cert.h
+++ b/cpp/log/cert.h
@@ -31,7 +31,7 @@ class Cert {
   }
   ~Cert();
 
-  enum Status {
+  enum class Status {
     TRUE,
     // OpenSSL makes it very hard to distinguish between input errors
     // (cert somehow corrupt) and internal library errors (malloc errors,
@@ -50,6 +50,11 @@ class Cert {
     // unconditionally not accepted.
     UNSUPPORTED_ALGORITHM,
   };
+
+  static const Status TRUE = Status::TRUE;
+  static const Status FALSE = Status::FALSE;
+  static const Status ERROR = Status::ERROR;
+  static const Status UNSUPPORTED_ALGORITHM = Status::UNSUPPORTED_ALGORITHM;
 
   bool IsLoaded() const {
     return x509_ != NULL;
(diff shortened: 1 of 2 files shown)


Martin2112:
Good idea. I thought that this would need more work.

pphaneuf:
I thought so as well, then a lightbulb came up with the constants. :smiley:

cpp/log/cert.h
```diff
@@ -272,6 +277,26 @@ class TbsCertificate {
...
+      break;
+    case Cert::Status::FALSE:
+      out << "FALSE";
+      break;
+    case Cert::Status::ERROR:
+      out << "ERROR";
+      break;
+    case Cert::Status::UNSUPPORTED_ALGORITHM:
+      out << "UNSUPPORTED_ALGORITHM";
+      break;
```
RJPercival: I'd add a default case that sets the fail bit on the ostream.

pphaneuf: Hmm, never thought of that... I did it different from my usual, which is to `return` in every case (instead of `break`), and have a default behaviour (output something like "invalid Cert::Status", say) after the `switch` statement (I avoid `default` to get the compiler to warn us if new values get added to the `enum`).

RJPercival: That sounds like the best way of doing it. I really must get used to all these post-2006 compiler innovations I get to enjoy now!

pphaneuf:
Tweaked the `operator<<` a bit.

cpp/log/cert.h
```diff
@@ -272,6 +277,27 @@ class TbsCertificate {
...
+      return out;
+    case Cert::Status::ERROR:
+      out << "ERROR";
+      return out;
+    case Cert::Status::UNSUPPORTED_ALGORITHM:
+      out << "UNSUPPORTED_ALGORITHM";
+      return out;
+  }
+
+  out << "<unknown Cert::Status: " << static_cast<int>(status) << ")";
```
Martin2112: Surprised that you can cast an enum class value directly to int. But I'm often surprised by C++.

Martin2112:
Still LGTM.

cpp/log/cert.h
```diff
@@ -272,6 +277,27 @@ class TbsCertificate {
   DISALLOW_COPY_AND_ASSIGN(TbsCertificate);
 };
 
+inline std::ostream& operator<<(std::ostream& out,
+                                const Cert::Status& status) {
+  switch (status) {
+    case Cert::Status::TRUE:
+      out << "TRUE";
+      return out;
```
RJPercival: Each of these pairs of lines could be merged into one.

cpp/log/cert.h
pphaneuf: A normal `enum` *is* an `int` (so you can assign to `int`, or from `int`, *without any cast*). The "innovation" with `enum class` (new in C++11) is that you now need to do something explicitly to get the conversion (the cast).

cpp/log/cert.h
pphaneuf: You mean `return out << "TRUE";`, for example? Hmm, sure, why not!

PR #917, opened 2015-09-03
Convert a few accessors from Cert::Status to bool/void.

pphaneuf:
This change is dangerous, but I've tested it on top of #916 (with which it is safe). Rebased on top of master, since it's actually independent (but I'd understand if you wouldn't want to merge this before #916!).

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -1233,33 +1233,32 @@ CertChain::CertChain(const string& pem_string) {
 }
 
 
-Cert::Status CertChain::AddCert(Cert* cert) {
+bool CertChain::AddCert(Cert* cert) {
   if (!cert || !cert->IsLoaded()) {
     LOG(ERROR) << "Attempting to add an invalid cert";
     if (cert)
       delete cert;
-    return Cert::ERROR;
+    return false;
   }
   chain_.push_back(cert);
-  return Cert::TRUE;
+  return true;
 }
 
 
-Cert::Status CertChain::RemoveCert() {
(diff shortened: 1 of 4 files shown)


cpp/log/cert.cc
```diff
@@ -1228,33 +1228,32 @@ CertChain::CertChain(const string& pem_string) {
...
   chain_.push_back(cert);
-  return Cert::TRUE;
+  return true;
 }
 
 
-Cert::Status CertChain::RemoveCert() {
-  if (!IsLoaded()) {
+void CertChain::RemoveCert() {
+  if (IsLoaded()) {
```
Martin2112: We're losing the status, is it never checked by callers?

cpp/log/cert.h
```diff
@@ -285,21 +285,21 @@ class CertChain {
...
+  // Else returns false.
+  bool AddCert(Cert* cert);
 
   // Remove a cert from the end of the chain.
-  // If successful, returns TRUE.
-  // If the chain is empty, returns ERROR.
-  Cert::Status RemoveCert();
+  // If successful, returns true.
+  // If the chain is empty, returns false.
+  void RemoveCert();
```
Martin2112: Comment doesn't match void function.

cpp/log/cert.cc
pphaneuf: Nope!

cpp/log/cert.h
pphaneuf: Updated the comment.

PR #918, opened 2015-09-04
Use unique_ptr rather than shared_ptr for Cert vectors.

pphaneuf:
That's what we wanted all along, but didn't have C++11 back then (as you can also tell by the `> >`).

I'm having some evening project of managing the damned memory, and `unique_ptr` is better suited for that...

### DIFF
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -885,11 +885,11 @@ void GetEntries() {
 int GetRoots() {
   HTTPLogClient client(FLAGS_ct_server);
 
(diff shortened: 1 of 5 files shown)


cpp/client/ct.cc
```diff
@@ -885,11 +885,11 @@ void GetEntries() {
 int GetRoots() {
```
Martin2112: Not your issue but not sure why this is an int method.

pphaneuf: This function is a sub-command, it becomes the exit status, in the `main()`.

PR #919, opened 2015-09-04
Add STH verification to the gossip server.

AlCutter:


AlCutter:
ping

go/client/logclient.go
```diff
@@ -230,9 +230,11 @@ func (c *LogClient) addChainWithRetry(path string, chain []ct.ASN1Cert) (*ct.Sig
 	if err != nil {
 		return nil, err
 	}
+    var logID ct.SHA256Hash
+    copy(logID[:], rawLogID)
 	return &ct.SignedCertificateTimestamp{
 		SCTVersion: resp.SCTVersion,
-		LogID:      rawLogID,
+        LogID:      logID,
```
pphaneuf: This is some... intriguing indentation?

AlCutter: *cough*
fixed.

go/gossip/handler.go
```diff
@@ -87,6 +122,20 @@ func (h *Handler) HandleSTHPollination(rw http.ResponseWriter, req *http.Request
 
 // NewHandler creates a new Handler object, taking a pointer a Storage object to
 // use for storing and retrieving feedback and pollination data.
-func NewHandler(s *Storage) Handler {
-	return Handler{storage: s}
+func NewHandler(s *Storage, v SignatureVerifierMap) Handler {
```
pphaneuf: No mention of the `SignatureVerifierMap` in the comment?

go/gossip/handler.go
```diff
@@ -87,6 +122,20 @@ func (h *Handler) HandleSTHPollination(rw http.ResponseWriter, req *http.Request
...
+	return Handler{
+		storage:   s,
+		verifiers: v,
+		clock:     realClock{},
+	}
+}
+
+// NewHandler creates a new Handler object, taking a pointer a Storage object to
+// use for storing and retrieving feedback and pollination data.
```
pphaneuf: Same here.

go/gossip/handler.go
```diff
@@ -66,13 +85,29 @@ func (h *Handler) HandleSTHPollination(rw http.ResponseWriter, req *http.Request
...
+			log.Printf("Pollination entry for unknown logID: %s", sth.LogID.Base64String())
+			continue
+		}
+		if err := v.VerifySTHSignature(sth); err != nil {
+			log.Printf("Failed to verify STH, dropping: %v", err)
+			continue
+		}
+		sthToKeep = append(sthToKeep, sth)
+	}
+	p.STHs = sthToKeep
```
pphaneuf: Why the `sthToKeep` temporary, rather than appending directly?

go/gossip/main/gossip_server.go
```diff
@@ -1,18 +1,56 @@
 package main
 
 import (
+    "errors"
 	"flag"
+    "fmt"
```
pphaneuf: Wacky formatting strikes again?

go/gossip/main/gossip_server.go
```diff
@@ -1,18 +1,56 @@
...
 	"net/http"
+    "strings"
 
+	ct "github.com/google/certificate-transparency/go"
 	"github.com/google/certificate-transparency/go/gossip"
 )
 
 var dbPath = flag.String("database", "/tmp/gossip.sq3", "Path to database.")
 var listenAddress = flag.String("listen", ":8080", "Listen address:port for HTTP server.")
+var logKeys = flag.String("log_public_keys", "", "Comma separated list of files containing trusted Logs' public keys in PEM format")
```
pphaneuf: Single file with all the keys might be nicer? Not a big deal, can be addressed later...

go/gossip/main/gossip_server.go
```diff
@@ -1,18 +1,56 @@
...
+        }
+        return &m, nil
+}
 
 func main() {
 	flag.Parse()
+    verifierMap, err := createVerifiers()
+    if err != nil {
+            log.Fatalf("Failed to load log public keys: %v", err)
+    }
```
pphaneuf: Get a real editor? :wink:

go/types.go
```diff
@@ -196,22 +239,58 @@ type LogEntry struct {
...
+func (s SHA256Hash) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + s.Base64String() + `"`), nil
+}
+
+// UnmarshalJSON implements the json.Unmarshaller interface.
+func (s *SHA256Hash) UnmarshalJSON(b []byte) error {
+	if b[0] != '"' || b[len(b)-1] != '"' {
+		return fmt.Errorf("expected a quoted string, but got '%s'", string(b))
+	}
+	content := string(b[1 : len(b)-1])
```
pphaneuf: This seems a bit iffy, with regard to JSON (quoting and other wackiness)... Would it be possible to leverage `UnmarshalJSON` for `string`, instead? So it would just be something like `content.UnmarshalJSON(b)`? Presumably, there's handling in there for the cases where it's not a string, and so on...

pphaneuf:
Okay, review done, just a few niggles and a question.

go/gossip/handler.go
AlCutter: Added.

go/gossip/handler.go
AlCutter: Same here.

go/gossip/handler.go
AlCutter: 'cos I'm actually, in effect, *removing* things from` p.STHs` here, and using the temporary this way so it's `O(n)` and not `(n*n)` in element copies. Could do that in-place by swapping elements and truncating of course, but that seems to be extra complexity for no real gain.

go/gossip/main/gossip_server.go
AlCutter: Escaped the clutches of `go fmt`.
No longer, though.

go/gossip/main/gossip_server.go
AlCutter: The code actually already supports that use case, see the `for len(pem) ... ` bit.

go/gossip/main/gossip_server.go
AlCutter: I use cat & Ctrl-D.

go/types.go
AlCutter: There is no `string.UnmarshalJSON()`, but I can do it via `json.Marshal()`.

AlCutter:
PTAL

go/gossip/handler.go
pphaneuf: Ah, I missed the source, I thought it was coming from somewhere else, for some reason. Erasing while iterating is always a bit iffy, fine like this.

go/gossip/main/gossip_server.go
pphaneuf: Am I right, or it might not skip junk between PEMs? Again, can be fixed up later, I was already thinking that the bit about the `clock` interface could have easily been in a separate PR... :wink:

pphaneuf:
LGTM.

PR #920, opened 2015-09-07
Add compile-time hardening flags, and enable by default.

AlCutter:


.travis.yml
```diff
@@ -42,6 +42,7 @@ install:
 
 script:
  - ./configure && make -j$(getconf _NPROCESSORS_ONLN) check VERBOSE=1
+ - cat config.log
```
pphaneuf: Remove that line. :smiley:

configure.ac
```diff
@@ -25,6 +28,16 @@ AC_CHECK_PROGS([ANT], [ant])
 
 PKG_CHECK_MODULES([json_c], [json-c])
 
+if test "x${enable_hardening}" != "xno"; then
+  AS_VAR_APPEND([CFLAGS], [" -Qunused-arguments -fstack-protector-all -fPIE -Wa,--noexecstack -Wformat -Wformat-security"])
+  AS_VAR_APPEND([CPPFLAGS], [" -D_FORTIFY_SOURCE=2"])
```
pphaneuf: I'd move this line up by one, to have to two "related" `CFLAGS` and `CXXFLAGS` together?

pphaneuf:
LGTM, just a nit.

.travis.yml
AlCutter: Now that we've finished debugging why it didn't build on Travis, I will!

AlCutter:
I've had to modify a bit to support `gcc/g++` too, PTAL.

configure.ac
```diff
@@ -14,8 +17,8 @@ AC_ARG_VAR([GTEST_DIR], [directory containing Google Test])
 
 # Checks for programs.
 AC_PROG_AWK
-AC_PROG_CXX
-AC_PROG_CC
+AC_PROG_CXX([g++])
+AC_PROG_CC([gcc])
```
pphaneuf: Are these really necessary? These will limit detection of compilers... Maybe what would be better would be to move the flag-setting much higher, so that all the tests are done *with* the flags? Possibly even failing at the "detecting C++ compiler" step...

pphaneuf:
This feels like it's slowly turning into the tarpit I was expecting... :disappointed:

configure.ac
AlCutter: Whoops, no that shouldn't have been there, that was me messing around earlier on.

AlCutter: Removed.

pphaneuf:
All right, let's have that, and figure it out better later.

PR #921, opened 2015-09-09
Go: Add SCT serialization

AlCutter:


go/scanner/scanner_test_data.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 const (
-    // TODO(alcutter): this signature is syntactically correct, but invalid. 
+	// TODO(alcutter): this signature is syntactically correct, but invalid.
```
pphaneuf: Ouch. :smiley:

go/serialization.go
```diff
@@ -352,6 +350,77 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+	}
+	if err := binary.Write(&buf, binary.BigEndian, sig); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+// SerializeSCT serializes the passed in sct into the format specified
+// by RFC6962 section 3.2
+func SerializeSCT(sct SignedCertificateTimestamp) ([]byte, error) {
```
pphaneuf: Would it make sense to serialize to an `io.Writer`, instead of to `[]byte`? This is used only for testing at the moment (that I can see), so not a big deal, but if you're going to have more of this...

pphaneuf:
LGTM

go/serialization.go
AlCutter: It may well do (as you saw I did just that with the DS serialisation), I'm planning to go back through here once I've finished with this round of additions and normalise the naming a bit anyway, so I might extend that to normalising the API style a bit too.

pphaneuf: I am 200% in favour of small, well contained, and easy to review PRs. :smiley:

PR #922, opened 2015-09-10
Convert HasExtension, HasCriticialExtension + some others to util::Status

Martin2112:
More refactoring to get rid of Cert::Status usages.

### DIFF
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -202,14 +202,14 @@ static bool PrecertChainToEntry(const cert_trans::PreCertChain& chain,
     return false;
   }
 
-  Cert::Status status =
+  const StatusOr<bool> has_poison =
       chain.LeafCert()->HasExtension(cert_trans::NID_ctPoison);
-  if (status != Cert::TRUE && status != Cert::FALSE) {
+  if (!has_poison.ok()) {
     LOG(ERROR) << "Failed to test for poison extension.";
     return false;
   }
 
-  if (status == Cert::FALSE) {
+  if (!has_poison.ValueOrDie()) {
     LOG(ERROR) << "Leaf cert doesn't seem to be a Precertificate (no Poison).";
     return false;
   }
@@ -712,9 +712,9 @@ static void DiagnoseCertChain() {
                           << " is not a valid PEM-encoded certificate chain";
 
 
-  if (chain.LeafCert()->HasExtension(
-          cert_trans::NID_ctEmbeddedSignedCertificateTimestampList) !=
-      Cert::TRUE) {
+  const StatusOr<bool> has_timestamp_list = chain.LeafCert()->HasExtension(
+      cert_trans::NID_ctEmbeddedSignedCertificateTimestampList);
+  if (!has_timestamp_list.ok() || !has_timestamp_list.ValueOrDie()) {
     LOG(ERROR) << "Certificate has no embedded SCTs";
     return;
   }
@@ -813,9 +813,10 @@ void WrapEmbedded() {
   CertChain chain(pem_chain);
   CHECK(chain.IsLoaded()) << cert_file
                           << " is not a valid PEM-encoded certificate chain";
-  CHECK_EQ(Cert::TRUE,
-           chain.LeafCert()->HasExtension(
-               cert_trans::NID_ctEmbeddedSignedCertificateTimestampList));
+  CHECK(chain.LeafCert()
+            ->HasExtension(
+                cert_trans::NID_ctEmbeddedSignedCertificateTimestampList)
+            .ValueOrDie());
 
   string serialized_scts;
   CHECK_EQ(::util::Status::OK,
--- a/cpp/client/ssl_client.cc
+++ b/cpp/client/ssl_client.cc
@@ -184,9 +184,20 @@ int SSLClient::VerifyCallback(X509_STORE_CTX* ctx, void* arg) {
 
   string serialized_scts;
   // First, see if the cert has an embedded proof.
-  if (chain.LeafCert()->HasExtension(
-          cert_trans::NID_ctEmbeddedSignedCertificateTimestampList) ==
-      Cert::TRUE) {
+  const StatusOr<bool> has_embedded_proof = chain.LeafCert()->HasExtension(
+      cert_trans::NID_ctEmbeddedSignedCertificateTimestampList);
+
+  // Pull out the superfluous cert extension if it exists for use later
+  // Let's assume the superfluous cert is always last in the chain.
+  const StatusOr<bool> superf_has_timestamp_list =
+      input_chain.Length() > 1
+          ? input_chain.LastCert()->HasExtension(
+                cert_trans::NID_ctSignedCertificateTimestampList)
+          : util::Status::UNKNOWN;
+
+  // First check for embedded proof, if not present then look for the proof in a
+  // superfluous cert.
+  if (has_embedded_proof.ok() && has_embedded_proof.ValueOrDie()) {
     LOG(INFO) << "Embedded proof extension found in certificate, "
               << "verifying...";
     util::Status status = chain.LeafCert()->OctetStringExtensionData(
@@ -199,12 +210,8 @@ int SSLClient::VerifyCallback(X509_STORE_CTX* ctx, void* arg) {
       CHECK_EQ(Code::NOT_FOUND, status.CanonicalCode());
       LOG(ERROR) << "Failed to parse extension data: corrupt cert?";
     }
-    // Else look for the proof in a superfluous cert.
-    // Let's assume the superfluous cert is always last in the chain.
-  } else if (input_chain.Length() > 1 &&
-             input_chain.LastCert()->HasExtension(
-                 cert_trans::NID_ctSignedCertificateTimestampList) ==
-                 Cert::TRUE) {
+  } else if (superf_has_timestamp_list.ok() &&
+             superf_has_timestamp_list.ValueOrDie()) {
     LOG(INFO) << "Proof extension found in certificate, verifying...";
     util::Status status = input_chain.LastCert()->OctetStringExtensionData(
         cert_trans::NID_ctSignedCertificateTimestampList, &serialized_scts);
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -278,38 +278,42 @@ bool Cert::IsIdenticalTo(const Cert& other) const {
 }
 
 
-Cert::Status Cert::HasExtension(int extension_nid) const {
+util::StatusOr<bool> Cert::HasExtension(int extension_nid) const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
   }
 
   const StatusOr<int> index(ExtensionIndex(extension_nid));
   if (index.ok()) {
-    return TRUE;
+    return true;
   }
 
   if (index.status().CanonicalCode() == util::error::NOT_FOUND) {
-    return FALSE;
+    return false;
   }
 
-  return ERROR;
+  return util::Status(Code::INTERNAL, "Failed to get extension");
 }
 
 
-Cert::Status Cert::HasCriticalExtension(int extension_nid) const {
+StatusOr<bool> Cert::HasCriticalExtension(int extension_nid) const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
-    return ERROR;
+    return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
   }
 
   const StatusOr<X509_EXTENSION*> ext(GetExtension(extension_nid));
   if (!ext.ok()) {
-    return ext.status().CanonicalCode() == util::error::NOT_FOUND ? FALSE
-                                                                  : ERROR;
+    // The extension may be absent, which is not an error
+    if (ext.status().CanonicalCode() == util::error::NOT_FOUND) {
+      return false;
+    } else {
+      return util::Status(Code::INTERNAL, "Failed to get extension");
+    }
   }
 
-  return X509_EXTENSION_get_critical(ext.ValueOrDie()) > 0 ? TRUE : FALSE;
+  return X509_EXTENSION_get_critical(ext.ValueOrDie()) > 0;
 }
 
 
@@ -683,9 +687,10 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
   // Let's first check if the extension is present. This allows us to
   // distinguish between "NID not recognized" and the more harmless
   // "extension not found, found more than once or corrupt".
-  Cert::Status status = HasExtension(extension_nid);
-  if (status != TRUE)
-    return status;
+  const StatusOr<bool> status = HasExtension(extension_nid);
+  if (!status.ok() || !status.ValueOrDie()) {
+    return StatusOrBoolToCertStatus(status);
+  }
 
   int crit;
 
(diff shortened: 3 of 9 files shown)


Martin2112:
Rebased + conflicts fixed. PTAL.

Martin2112:
Rebased + latest conflicts fixed.

pphaneuf:
-- Gee, Brain, what do you want to do tonight?
-- The same thing we do every night, Pinky - rebase and fix conflicts!

Martin2112:
I know and they're all my own conflicts.

pphaneuf:
That's one reason why I sometimes send in only one change at a time, because I have the others in a branch that's *on top* of the one I sent for review (so it won't conflict, the only conflicts would be for changes when addressing comments, usually more minimal). But it's more git branch wrangling, and it loses a bit of "reviewing concurrency"...

Martin2112:
Yes these are all pretty much separate, the conflicts were minimal.

pphaneuf:
Of note: I started reviewing this PR a number of times now, but I've given up before getting to the end every time... It's a bit too big for my taste?

I might have liked better tackling `HasExtension` (in itself quite simple) and its repercussions (less simple), then another one that's in itself simple and its not-so-simple repercussions, etc.

Martin2112:
I more or less did it by groups of logically related functionality. I think it t made it easier to work out that the logic hadn't changed, if not to review.

pphaneuf:
Yeah, they're related, but for reviewing, they traverse more than one layer, so I need to check all the callers, *and* the callers callers, and at some point, I lose the plot. :stuck_out_tongue:

For example, that little change in behaviour in that other PR, where `status == TRUE` became something else that's *almost* the same, is much easier to fish out in a smaller sized diff. In a bigger diff like this one, I don't feel nearly as confident (which might very well be my own limitation!).

pphaneuf:
Heck, I find #910 complicated and scary! :wink: 

But I'll buckle down and do both, stand by... :smiley:

Martin2112:
I know what you mean but pulling apart this type of logic often doesn't work easily on any obvious axis. It's nearly done now at least.

cpp/client/ct.cc
```diff
@@ -202,14 +202,14 @@ static bool PrecertChainToEntry(const cert_trans::PreCertChain& chain,
     return false;
   }
 
-  Cert::Status status =
+  StatusOr<bool> status =
```
AlCutter: Couple of while-you're-here ones:

Constable, constable?
/status/has_poison/ ?

cpp/client/ct.cc
```diff
@@ -813,9 +813,11 @@ void WrapEmbedded() {
   CertChain chain(pem_chain);
   CHECK(chain.IsLoaded()) << cert_file
                           << " is not a valid PEM-encoded certificate chain";
-  CHECK_EQ(Cert::TRUE,
-           chain.LeafCert()->HasExtension(
-               cert_trans::NID_ctEmbeddedSignedCertificateTimestampList));
+  CHECK_EQ(true,
```
AlCutter: CHECK(chain...)

cpp/client/ssl_client.cc
```diff
@@ -184,9 +184,17 @@ int SSLClient::VerifyCallback(X509_STORE_CTX* ctx, void* arg) {
 
   string serialized_scts;
   // First, see if the cert has an embedded proof.
-  if (chain.LeafCert()->HasExtension(
-          cert_trans::NID_ctEmbeddedSignedCertificateTimestampList) ==
-      Cert::TRUE) {
+  const StatusOr<bool> has_timestamp_list = chain.LeafCert()->HasExtension(
```
AlCutter: `has_embedded_proof` to line up with the comment?

cpp/client/ssl_client.cc
```diff
@@ -201,10 +209,8 @@ int SSLClient::VerifyCallback(X509_STORE_CTX* ctx, void* arg) {
     }
     // Else look for the proof in a superfluous cert.
     // Let's assume the superfluous cert is always last in the chain.
```
AlCutter: Probably need to move this comment with the assumption up to ~L190

cpp/log/cert.cc
```diff
@@ -683,9 +687,10 @@ Cert::Status Cert::ExtensionStructure(int extension_nid,
   // Let's first check if the extension is present. This allows us to
   // distinguish between "NID not recognized" and the more harmless
   // "extension not found, found more than once or corrupt".
-  Cert::Status status = HasExtension(extension_nid);
-  if (status != TRUE)
-    return status;
+  StatusOr<bool> status = HasExtension(extension_nid);
```
AlCutter: const

cpp/log/cert.cc
```diff
@@ -966,15 +971,20 @@ util::Status Cert::IsValidNameConstrainedIntermediateCa() const {
   // If it's not a CA cert or there is no name constraint extension then we
   // don't need to apply the rules any further
   StatusOr<bool> has_ca_constraint = HasBasicConstraintCATrue();
```
AlCutter: const here & next line

cpp/log/cert.cc
```diff
@@ -966,15 +971,20 @@ util::Status Cert::IsValidNameConstrainedIntermediateCa() const {
...
+  StatusOr<bool> has_name_constraints = HasExtension(NID_name_constraints);
+
   if (!has_ca_constraint.ok() || !has_ca_constraint.ValueOrDie()
-      || HasExtension(NID_name_constraints) == Cert::FALSE) {
+      || !has_name_constraints.ok() || !has_name_constraints.ValueOrDie()) {
```
AlCutter: This is different - previously failing to get the `NID_name_constraints` extension would not have returned `OK`, but an `INVALID_ARGUMENT` below.

Arguably that's misleading, but perhaps we could do something like
```C++
if (!has_ca_constraint.ok()) {
  return has_ca_constraint.status();
}
if (!has_name_constraints.ok()) {
  return has_name_constraints.status();
}
if (!has_ca_constraint.ValueOrDie() || !has_name_constraints.ValueOrDie()) {
  return OK;
}
```

cpp/log/cert.cc
```diff
@@ -966,15 +971,20 @@ util::Status Cert::IsValidNameConstrainedIntermediateCa() const {
...
     return util::Status::OK;
   }
 
   // So there now must be a CT extension and the name constraint must not be
   // in error
-  if (HasExtension(NID_name_constraints) != Cert::TRUE
-      || HasExtension(NID_ctNameConstraintNologIntermediateCa) != Cert::TRUE) {
+  StatusOr<bool> has_ct_nolog_intermediate =
```
AlCutter: const

cpp/log/cert.cc
```diff
@@ -966,15 +971,20 @@ util::Status Cert::IsValidNameConstrainedIntermediateCa() const {
...
+      HasExtension(NID_ctNameConstraintNologIntermediateCa);
+
+  if (!has_name_constraints.ValueOrDie() || !has_ct_nolog_intermediate.ok() ||
```
AlCutter: I think that `has_name_constraints` must be `true` here, so maybe pull this out of the `if` and `CHECK(has_name_contraints.ValueOrDie());` ?

cpp/log/cert.cc
```diff
@@ -1419,49 +1429,53 @@ void CertChain::ClearChain() {
...
-    return Cert::ERROR;
+    return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
   }
 
   const Cert* pre = PreCert();
 
   // (1) Check that the leaf contains the critical poison extension.
-  Cert::Status status = pre->HasCriticalExtension(cert_trans::NID_ctPoison);
-  if (status != Cert::TRUE)
+  StatusOr<bool> status = pre->HasCriticalExtension(cert_trans::NID_ctPoison);
```
AlCutter:
```C++
const StatusOr<bool> has_poison(...);
```

cpp/log/cert.cc
```diff
@@ -1419,49 +1429,53 @@ void CertChain::ClearChain() {
...
+  if (!status.ok() || !status.ValueOrDie()) {
     return status;
+  }
 
   // (2) If signed by a Precertificate Signing Certificate, check that
   // the AKID extensions are compatible.
-  status = UsesPrecertSigningCertificate();
-  if (status == Cert::FALSE) {
+  StatusOr<bool> uses_precert_signing = UsesPrecertSigningCertificate();
```
AlCutter: const

cpp/log/cert.cc
```diff
@@ -1419,49 +1429,53 @@ void CertChain::ClearChain() {
...
+  if (uses_precert_signing.ok() && !uses_precert_signing.ValueOrDie()) {
     // If there is no precert signing extendedKeyUsage, no more checks:
     // the cert was issued by a regular CA.
-    return Cert::TRUE;
+    return true;
   }
-  if (status != Cert::TRUE)
+  if (!uses_precert_signing.ok() || !uses_precert_signing.ValueOrDie()) {
```
AlCutter:
```C++
if (!uses_precert_signing.ok()) {
  return uses_precert_signing.status();  // incorrectly returning status currently!
}

CHECK(uses_precert_signing.ValueOrDie());
```
?

cpp/log/cert.cc
```diff
@@ -1419,49 +1429,53 @@ void CertChain::ClearChain() {
...
+  }
 
   const Cert* issuer = PrecertIssuingCert();
   // If pre has the extension set but the issuer doesn't, error.
-  status = pre->HasExtension(NID_authority_key_identifier);
-  if (status == Cert::FALSE)
-    return Cert::TRUE;
-  if (status != Cert::TRUE)
-    return status;
+  StatusOr<bool> has_akid = pre->HasExtension(NID_authority_key_identifier);
```
AlCutter: const

cpp/log/cert.cc
```diff
@@ -1419,49 +1429,53 @@ void CertChain::ClearChain() {
...
+
+  if (has_akid.ok() && !has_akid.ValueOrDie()) {
+    return true;
+  }
+  if (!has_akid.ok()) {
+    return has_akid;
+  }
```
AlCutter: Could add `CHECK(has_akid.ValueOrDie())` here to guard against future modifications to this getting it wrong/document the invariant.

cpp/log/cert_checker.cc
```diff
@@ -141,14 +141,15 @@ Status CertChecker::CheckCertChain(CertChain* chain) const {
     return Status(util::error::INVALID_ARGUMENT, "invalid certificate chain");
 
   // Weed out things that should obviously be precert chains instead.
-  Cert::Status status =
+  StatusOr<bool> status =
```
AlCutter: const
/status/has_poison/

cpp/log/cert_checker.cc
```diff
@@ -178,12 +179,15 @@ Status CertChecker::CheckIssuerChain(CertChain* chain) const {
...
                                       string* issuer_key_hash,
                                       string* tbs_certificate) const {
-  if (!chain || !chain->IsLoaded())
+  if (!chain || !chain->IsLoaded()) {
     return Status(util::error::INVALID_ARGUMENT, "invalid certificate chain");
-  Cert::Status status = chain->IsWellFormed();
-  if (status == Cert::FALSE)
+  }
+
+  StatusOr<bool> status = chain->IsWellFormed();
```
AlCutter:
```C++
const StatusOr<bool> chain_well_formed(chain->…);
```

cpp/log/cert_checker.cc
```diff
@@ -178,12 +179,15 @@ Status CertChecker::CheckIssuerChain(CertChain* chain) const {
...
+  if (status.ok() && !status.ValueOrDie()) {
     return Status(util::error::INVALID_ARGUMENT, "prechain not well formed");
-  if (status != Cert::TRUE) {
+  }
+  if (!status.ok() || !status.ValueOrDie()) {
```
AlCutter:
```C++
if (!chain_well_formed.ok()) {
 …
}

CHECK(chain_well_formed.ValueOrDie());
```

cpp/log/cert_checker.cc
```diff
@@ -208,12 +212,13 @@ Status CertChecker::CheckPreCertChain(PreCertChain* chain,
       return status;
   }
 
-  Cert::Status uses_pre_issuer = chain->UsesPrecertSigningCertificate();
-  if (uses_pre_issuer != Cert::TRUE && uses_pre_issuer != Cert::FALSE)
+  StatusOr<bool> uses_pre_issuer = chain->UsesPrecertSigningCertificate();
```
AlCutter: const

cpp/log/cert_submission_handler.cc
```diff
@@ -32,14 +33,14 @@ bool CertSubmissionHandler::X509ChainToEntry(const CertChain& chain,
   if (!chain.IsLoaded())
     return false;
 
-  Cert::Status status = chain.LeafCert()->HasExtension(
+  StatusOr<bool> status = chain.LeafCert()->HasExtension(
```
AlCutter: const ... has_embedded_proof

cpp/log/cert_submission_handler.cc
```diff
@@ -133,9 +134,9 @@ bool CertSubmissionHandler::SerializedTbs(const Cert& cert, string* result) {
     return false;
   }
 
-  Cert::Status status = cert.HasExtension(
+  StatusOr<bool> status = cert.HasExtension(
```
AlCutter: `const … has_embedded_proof`

AlCutter:
All done, that wasn't so bad :)

cpp/log/cert.cc
Martin2112: I think you're right. While it's slightly ambiguous I don't think it makes sense to return OK if the extension couldn't be looked up as it's only going to fail later when we read the data for it. That shouldn't happen anyway as it was registered at startup. We already checked the cert was loaded, which is the only other error case.

cpp/log/cert_checker.cc
Martin2112: Not sure about that one. ValueOrDie() only dies if status is not OK. We want to return an error status if the chain isn't well formed.

Martin2112:
That hopefully addresses all the comments.

cpp/client/ct.cc
Martin2112: Done.

cpp/client/ct.cc
Martin2112: Done.

cpp/client/ssl_client.cc
Martin2112: Done.

cpp/client/ssl_client.cc
Martin2112: Done. Split it into two parts, one where we get the value and one where the decision occurs.

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc
Martin2112: Fixed up the incorrect return. Didn't add a CHECK as this shouldn't exit if it's not well formed.

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert_checker.cc
Martin2112: Done.

cpp/log/cert_checker.cc
Martin2112: Done.

cpp/log/cert_checker.cc
Martin2112: Didn't do this. It returns a non OK status unless the chain is well formed. I could add the CHECK as an invariant guarantee but I think the if statement is correct.

cpp/log/cert_checker.cc
Martin2112: Done.

cpp/log/cert_submission_handler.cc
Martin2112: Done.

cpp/log/cert_submission_handler.cc
Martin2112: Done.

Martin2112:
Replied to all comments.

cpp/log/cert_checker.cc
AlCutter: I think it's right; if `chain_well_formed` was false (and ok), then we'd have returned at that point, so at 190 we only have to check if there was an error with the `!ok()`.

Martin2112: I think you're right. Fixing it now and running the tests again.

PR #925, opened 2015-09-10
Update remaining TBSCertificate methods to util::Status

Martin2112:
Updates call sites + test, and comments on status returned by the functions.

### DIFF
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -234,13 +234,15 @@ static bool PrecertChainToEntry(const cert_trans::PreCertChain& chain,
     LOG(ERROR) << "Failed to get TbsCertificate.";
     return false;
   }
-  if (tbs.DeleteExtension(cert_trans::NID_ctPoison) != Cert::TRUE) {
+  // DeleteExtension can return NOT_FOUND but we checked the extension exists
+  // above so this is not expected.
+  if (!tbs.DeleteExtension(cert_trans::NID_ctPoison).ok()) {
     LOG(ERROR) << "Failed to delete poison extension.";
     return false;
   }
 
   string tbs_der;
-  if (tbs.DerEncoding(&tbs_der) != Cert::TRUE) {
+  if (!tbs.DerEncoding(&tbs_der).ok()) {
     LOG(ERROR) << "Couldn't serialize TbsCertificate to DER.";
     return false;
   }
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -1108,14 +1108,14 @@ util::Status TbsCertificate::DeleteExtension(int extension_nid) {
     return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded (TBS)");
   }
 
-  StatusOr<int> status = ExtensionIndex(extension_nid);
+  const StatusOr<int> extension_index(ExtensionIndex(extension_nid));
   // If the extension doesn't exist then there is nothing to do and this
   // propagates the NOT_FOUND status.
-  if (!status.ok()) {
-    return status.status();
+  if (!extension_index.ok()) {
+    return extension_index.status();
   }
 
-  X509_EXTENSION* ext = X509_delete_ext(x509_, status.ValueOrDie());
+  X509_EXTENSION* ext = X509_delete_ext(x509_, extension_index.ValueOrDie());
 
   if (!ext) {
     // Truly odd.
@@ -1130,8 +1130,8 @@ util::Status TbsCertificate::DeleteExtension(int extension_nid) {
 
   // ExtensionIndex returns the first matching index - if the extension
   // occurs more than once, just give up.
-  status = ExtensionIndex(extension_nid);
-  if (status.ok() || status.status().CanonicalCode() == Code::INTERNAL) {
+  const StatusOr<int> ignored_index = ExtensionIndex(extension_nid);
+  if (ignored_index.ok()) {
     LOG(WARNING)
         << "Failed to delete the extension. Does the certificate have "
         << "duplicate extensions?";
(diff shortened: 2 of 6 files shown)


cpp/client/ct.cc
```diff
@@ -234,13 +234,15 @@ static bool PrecertChainToEntry(const cert_trans::PreCertChain& chain,
     LOG(ERROR) << "Failed to get TbsCertificate.";
     return false;
   }
-  if (tbs.DeleteExtension(cert_trans::NID_ctPoison) != Cert::TRUE) {
+  // DeleteExtension can return NOT_FOUND but we checked the extension exists
+  // above so this is not expected.
```
pphaneuf: And this function would return `false` in both cases, so no worries.

cpp/log/cert_test.cc
```diff
@@ -597,7 +597,7 @@ TEST_F(TbsCertificateTest, DerEncoding) {
 
   string cert_tbs_der, raw_tbs_der;
   EXPECT_OK(leaf.DerEncodedTbsCertificate(&cert_tbs_der));
-  EXPECT_EQ(Cert::TRUE, tbs.DerEncoding(&raw_tbs_der));
+  EXPECT_TRUE(tbs.DerEncoding(&raw_tbs_der).ok());
```
pphaneuf: Use `EXPECT_OK` instead? Everywhere in this file, of course.

cpp/log/cert.cc
```diff
@@ -1086,32 +1086,34 @@ Cert::Status TbsCertificate::DerEncoding(string* result) const {
...
     LOG(ERROR) << "TBS not loaded";
-    return Cert::ERROR;
+    return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded (TBS)");
   }
 
-  int extension_index;
-  Cert::Status status = ExtensionIndex(extension_nid, &extension_index);
-  if (status != Cert::TRUE)
-    return status;
+  StatusOr<int> status = ExtensionIndex(extension_nid);
```
pphaneuf: Would be better named `extension_index`, I think. Could be made `const`, and use constructor initialisation, while you're there?

cpp/log/cert.cc
```diff
@@ -1120,30 +1122,33 @@ Cert::Status TbsCertificate::DeleteExtension(int extension_nid) {
 
   // ExtensionIndex returns the first matching index - if the extension
   // occurs more than once, just give up.
-  int ignored;
-  status = ExtensionIndex(extension_nid, &ignored);
-  if (status == Cert::TRUE) {
+  status = ExtensionIndex(extension_nid);
```
pphaneuf: Could use a separate (`const`) `ignored` (or `ignored_index`), instead of reusing the one above?

cpp/log/cert.cc
```diff
@@ -1120,30 +1122,33 @@ Cert::Status TbsCertificate::DeleteExtension(int extension_nid) {
...
+  if (status.ok() || status.status().CanonicalCode() == Code::INTERNAL) {
```
pphaneuf: Isn't this changing the logic? `INTERNAL` used to be `Cert::ERROR`, the condition used to be `status == Cert::TRUE`, and now it's the equivalent of `status == Cert::TRUE || status == Cert::ERROR)`?

cpp/log/cert.cc
```diff
@@ -1151,27 +1156,27 @@ Cert::Status TbsCertificate::CopyIssuerFrom(const Cert& from) {
   X509_NAME* ca_name = X509_get_issuer_name(from.x509_);
   if (!ca_name) {
     LOG(WARNING) << "Issuer certificate has NULL name";
-    return Cert::FALSE;
+    return util::Status(Code::FAILED_PRECONDITION, "Issuer cert has NULL name");
```
pphaneuf: This one loses the ability to distinguish that error, but from what I can see, nobody care, this function could even just return a plain `bool`. :stuck_out_tongue:

pphaneuf:
Done with that review.

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert_test.cc
Martin2112: Done.

cpp/log/cert.cc
Martin2112: We shouldn't get INTERNAL here because it's the same NID we looked up once before without seeing this error and that return implies it wasn't known. However, now that I read it again that INTERNAL status will be propagated by the next if condition. I think it's more readable to remove the INTERNAL check and not worry about the log message. I'll do that.

Martin2112:
All fixed + rebased. PTAL.

cpp/log/cert.cc
```diff
@@ -1128,30 +1130,34 @@ Cert::Status TbsCertificate::DeleteExtension(int extension_nid) {
...
-  int ignored;
-  status = ExtensionIndex(extension_nid, &ignored);
-  if (status == Cert::TRUE) {
+  const StatusOr<int> ignored_index = ExtensionIndex(extension_nid);
+  if (ignored_index.ok()) {
     LOG(WARNING)
         << "Failed to delete the extension. Does the certificate have "
         << "duplicate extensions?";
-    return Cert::FALSE;
+    return util::Status(Code::ALREADY_EXISTS, "Multiple extensions in cert");;
```
pphaneuf: There's two semi-colons here.

cpp/log/cert.cc
```diff
@@ -1128,30 +1130,34 @@ Cert::Status TbsCertificate::DeleteExtension(int extension_nid) {
 
   // ExtensionIndex returns the first matching index - if the extension
   // occurs more than once, just give up.
-  int ignored;
-  status = ExtensionIndex(extension_nid, &ignored);
-  if (status == Cert::TRUE) {
+  const StatusOr<int> ignored_index = ExtensionIndex(extension_nid);
```
pphaneuf: Using assignment instead of constructor syntax, but no worries, it's all the same...

I just dislike the idea of "assigning to a `const` object", and how this triggers only the copy constructor (or even constructing in place, if RVO kicks in), short-circuiting any `operator=`.

pphaneuf:
LGTM

cpp/log/cert.cc
Martin2112: Guess who needed two goes to get the regex replace right? Will fix.

cpp/log/cert.cc
Martin2112: I'll fix it.

PR #926, opened 2015-09-10
Print out socket error string when connections fail.

AlCutter:


eranmes:
Drive-by LGTM (I'll point out that Chromium's preferred style is to initialize objects using = rather than c'tor. Do we prefer the opposite in our codebase?).

AlCutter:
Yes, we prefer c'tor initialisation 'round these parts.

PR #928, opened 2015-09-11
Convert some PreCert stuff and Cert loading from DER to util::Status

Martin2112:
Still a few more bits to convert but we're getting near to the end of this now!

### DIFF
--- a/cpp/log/cert_test.cc
+++ b/cpp/log/cert_test.cc
@@ -306,7 +306,7 @@ TEST_F(CertTest, LoadValidFromDer) {
   string der;
   ASSERT_OK(leaf.DerEncoding(&der));
   Cert second;
-  EXPECT_TRUE(second.LoadFromDerString(der).ok());
+  EXPECT_OK(second.LoadFromDerString(der));
   EXPECT_TRUE(second.IsLoaded());
 }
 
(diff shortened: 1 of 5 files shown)


cpp/log/cert_test.cc
```diff
@@ -306,7 +306,7 @@ TEST_F(CertTest, LoadValidFromDer) {
   string der;
   ASSERT_OK(leaf.DerEncoding(&der));
   Cert second;
-  EXPECT_EQ(Cert::TRUE, second.LoadFromDerString(der));
+  EXPECT_TRUE(second.LoadFromDerString(der).ok());
```
pphaneuf: Throughout the test, these `EXPECT_TRUE` should be `EXPECT_OK` instead.

cpp/log/cert_test.cc
```diff
@@ -691,23 +691,23 @@ TEST_F(CertChainTest, RemoveCert) {
 TEST_F(CertChainTest, IssuerChains) {
   // A single certificate.
   CertChain chain(leaf_pem_);
-  EXPECT_EQ(Cert::TRUE, chain.IsValidCaIssuerChainMaybeLegacyRoot());
+  EXPECT_TRUE(chain.IsValidCaIssuerChainMaybeLegacyRoot().ok());
```
pphaneuf: Would be better to have this as `EXPECT_THAT(..., StatusIs(...));`.

pphaneuf:
LGTM for the main code, test could be tweaked a bit, but I could live with it like this too.

Martin2112:
By "tweaked" do you mean something else than using EXPECT_OK? I'm going to fix all those as it's just regex replaces.

Martin2112:
I've rebased and plan to merge if it passes. If there are any other test changes to be done let me know and I'll follow up.

PR #929, opened 2015-09-11
Add a simple db_tool for inspecting log/mirror database contents.

AlCutter:


pphaneuf:
Missing some link-time dependency? Travis failed...

Makefile.am
```diff
@@ -248,6 +249,17 @@ cpp_tools_ct_clustertool_SOURCES = \
 	cpp/util/util.cc \
 	cpp/version.cc
 
+cpp_tools_db_tool_LDADD = \
+	cpp/libcore.a \
+	$(libevent_LIBS) \
+	-lprotobuf -lsqlite3 -lleveldb
```
pphaneuf: Ah, I see the problem. You should be using `$(leveldb_LIBS)` rather than `-lleveldb`.

cpp/tools/db_tool.cc
```diff
@@ -0,0 +1,129 @@
...
+  InitCT(&argc, &argv);
+
+  if (argc != 2) {
+    Usage();
+    return 1;
+  }
+
+  // ------8<----------8<---------8<-----------
+  // TODO(alcutter): Refactor this out into a common CreateDatabase() call
+  // somewhere.
```
pphaneuf: And it shall take a single string, and spit out a single `Database<>`. Yes, even `FileDB<>`.

pphaneuf:
LGTMish, as soon as you fix Travis. :wink:

PR #931, opened 2015-09-11
Allow TLS client to support various TLS versions.

AlCutter:
Defaults to only allowing TLSv1.2, but optionally can support TLSv1 or TLSv1.1 and above via a flag.
Also allows passing of acceptable ciphers for the TLS client through to OpenSSL, with a default of the Mozilla 'modern compatibility' set.

All current known CT Logs support TLSv1.2, except `ct.izenpe.com` which only supports TLSv1,   adjusted config accordingly.

### DIFF
--- a/cpp/net/connection_pool.cc
+++ b/cpp/net/connection_pool.cc
@@ -50,6 +50,24 @@ DEFINE_string(trusted_root_certs, "/etc/ssl/certs/ca-certificates.crt",
 DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
              "maximum number of URL fetcher connections per host:port");
 
+DEFINE_string(tls_client_minimum_protocol, "tlsv12",
+              "Minimum acceptable TLS "
+              "version protocol (tlsv1, tlsv11, tlsv12)");
+
+DEFINE_string(
+    tls_client_ciphers,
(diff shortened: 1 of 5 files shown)


cpp/net/connection_pool.cc
```diff
@@ -177,10 +195,47 @@ void ConnectionPool::Connection::ReleaseConnection() {
...
+               << FLAGS_tls_client_minimum_protocol;
+
+  if (FLAGS_tls_client_minimum_protocol == "tlsv1") {
+    SSL_CTX_set_options(ctx, kSslOpMinVersionTls1);
+  } else if (FLAGS_tls_client_minimum_protocol == "tlsv11") {
+    SSL_CTX_set_options(ctx, kSslOpMinVersionTls11);
+  } else if (FLAGS_tls_client_minimum_protocol == "tlsv12") {
+    SSL_CTX_set_options(ctx, kSslOpMinVersionTls12);
+  } else {
+    LOG(FATAL) << "Unrecognised TLS version "
```
pphaneuf: How 'bout a colon?

pphaneuf:
I'm *not* going to say that this "looks good", but all right, you can merge it. :stuck_out_tongue: 

This, right there, is why we can't have nice things.

PR #933, opened 2015-09-15
Don't dwell on bad STHs.

AlCutter:


pphaneuf:
LGTM

Bad STHs, a bit like bad breakups, shouldn't be dwelled on.

PR #934, opened 2015-09-15
Allow keys in the log list to be different sizes

RJPercival:
The Venafi log has a different sized key to the other logs, so this change is required in order to add it.


eranmes:
That's going to merit a change in Chromium for handling keys of different sizes. Let's have that reviewed first (and manually edit ct_known_logs_static.h) then change the way the code is generated.

eranmes:
Seeing as this was approved in https://codereview.chromium.org/1338393002/, LGTM

PR #935, opened 2015-09-16
Retry failed operations during updates.

AlCutter:


cloud/google/update_log.sh
```diff
@@ -11,21 +11,30 @@ set -e
 GCLOUD="gcloud"
 
 Header "Updating log instances..."
-for i in `seq 0 $((${LOG_NUM_REPLICAS} - 1))`; do
+i=0
+while [ $i -lt ${MIRROR_NUM_REPLICAS} ]; do
   echo "Updating ${LOG_MACHINES[${i}]}"
   echo "${LOG_META[${i}]}" > /tmp/metadata.${i}
```
phad: you'll retry forever, is that desired?

cloud/google/update_mirror.sh
```diff
@@ -7,25 +7,35 @@ fi
...
 source ${DIR}/config.sh $1
 
-set -e
 GCLOUD="gcloud"
 
 Header "Updating mirror instances..."
-for i in `seq 0 $((${MIRROR_NUM_REPLICAS} - 1))`; do
+i=0
+while [ $i -lt ${MIRROR_NUM_REPLICAS} ]; do
+#for i in `seq 0 $((${MIRROR_NUM_REPLICAS} - 1))`; do
```
phad: remove commented-out code

cloud/google/update_log.sh
```diff
@@ -11,21 +11,30 @@ set -e
...
+  fi
 
-  gcloud compute ssh ${LOG_MACHINES[${i}]} \
+  if ! gcloud compute ssh ${LOG_MACHINES[${i}]} \
       --zone ${LOG_ZONES[${i}]} \
       --command \
           'sudo docker pull gcr.io/'${PROJECT}'/super_duper:test &&
-           sudo docker kill $(sudo docker ps | grep super_duper | awk -- "{print \$1}" )'
+           sudo docker kill $(sudo docker ps | grep super_duper | awk -- "{print \$1}" )'; then
+    echo "Retrying"
```
phad: if this retry is needed then on the next iteration it will do the 'gcloud compute instances add-metadata' again, won't it?  is that OK?

cloud/google/update_log.sh
AlCutter: Yeah, I figured that's better than just 'silently' leaving the cluster in an inconsistent state - at least this way the operator will have to do something about it.

cloud/google/update_mirror.sh
AlCutter: Whoops, thanks.

cloud/google/update_log.sh
AlCutter: It will, but it's idempotent so it's ok.

phad: cool

cloud/google/update_log.sh
phad: sounds good.  the alternative is to count the retries and abort the script with an error status if you exceed max retries.

Issue #923, opened 2015-09-10
python/ct/crypto/verify.py does not support RSA public keys


linshunghuang:
Running python/utilities/logobserver/logobserver.py against https://ctlog.api.venafi.com fails with
```
ct.crypto.error.UnsupportedAlgorithmError: Key type 1 not supported
```
python/ct/config/logs.config:
```
ctlog:{
  log_server:'https://ctlog.api.venafi.com'
  log_id:'rDua7X+pZ0dXFZ5tfVdWcvnZgQCUHpve/+yhMTt1eC0='
  public_key_info:{
    type:RSA
    pem_key:'-----BEGIN RSA PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ'
    '8AMIIBCgKCAQEAolpIHxdSlTXLo1s6H1OCdpSj/4DyHDc8wLG9wVmLqy1lk9fz4ATVmm+/'
    '1iN2Nk8jmctUKK2MFUtlWXZBSpym97M7frGlSaQXUWyA3CqQUEuIJOmlEjKTBEiQAvpfDj'
    'CHjlV2Be4qTM6jamkJbiWtgnYPhJL6ONaGTiSPm7Byy57iaz/hbckldSOIoRhYBiMzeNoA'
    '0DiRZ9KmfSeXZ1rB8y8X5urSW+iBzf2SaOfzBvDpcoTuAaWx2DPazoOl28fP1hZ+kHUYvx'
    'bcMjttjauCFx+JII0dmuZNIwjfeG/GBb9frpSX219k1O4Wi6OEbHEr8at/XQ0y7gTikOxB'
    'n/s5wQIDAQAB\n-----END RSA PUBLIC KEY-----'
  }
}
```
Any plans to support it? Thanks!

eranmes:
Not surprising -given that's the first log with RSA keys.

Could you paste a full stack-trace as well as the command-line you used for running the logobserver? Just so we can re-create.
Then I can have a look at the complexity of implementing support for RSA keys.
As you can see from https://github.com/google/certificate-transparency/blob/master/python/ct/crypto/verify.py#L39 ECDSA assumption is somewhat hard-coded.

linshunghuang:
Sure. Using the modified `python/ct/config/logs.config` as pasted above, the command line is:
```
certificate-transparency$ cd python/
python$ export PYTHONPATH=${PYTHONPATH}:$(pwd)
[...]

PR #938, opened 2015-09-17
Support for RSA keys

RJPercival:
Fixes issue #923.

- Introduces a dependency on the pycrypto library.
- Runs the verify_test suite for both an ECDSA key and an RSA key.

### DIFF
--- a/python/ct/crypto/ecdsa_keys.py
+++ b/python/ct/crypto/ecdsa_keys.py
@@ -1,0 +1,76 @@
+from ct.crypto import error
+from ct.crypto import pem
+from ct.proto import client_pb2
+
+import hashlib
+import ecdsa
+
+class EcdsaPublicKey(object):
+    """Wraps an ECDSA public key."""
+
+    # The signature algorithm used for this public key."""
+    SIGNATURE_ALGORITHM = client_pb2.DigitallySigned.ECDSA
+    # The hash algorithm used for this public key."""
+    HASH_ALGORITHM = client_pb2.DigitallySigned.SHA256
+
+    # Markers to look for when reading a PEM-encoded ECDSA public key."""
+    __READ_MARKERS = ("PUBLIC KEY", "ECDSA PUBLIC KEY")
+    # A marker to write when writing a PEM-encoded ECDSA public key."""
+    __WRITE_MARKER = "ECDSA PUBLIC KEY"
+
+    def __init__(self, key_info):
+        """Loads a PEM-encoded ECDSA public key.
+
+        Args:
+        - key_info: KeyInfo protobuf message
+
+        Raises:
+        - PemError: If the key has an invalid encoding
+        """
+        if (key_info.type != client_pb2.KeyInfo.ECDSA):
+            raise error.UnsupportedAlgorithmError(
+                "Expected ECDSA key, but got key type %d" % key_info.type)
+
+        # Will raise a PemError on invalid encoding
+        self.__der, _ = pem.from_pem(key_info.pem_key, self.__READ_MARKERS)
+        try:
+            self.__key = ecdsa.VerifyingKey.from_der(self.__der)
+        except ecdsa.der.UnexpectedDER as e:
+            raise error.EncodingError(e)
+
+    def __repr__(self):
+        return "%r(public key: %r)" % (self.__class__.__name__,
+                                       pem.to_pem(self.__der,
+                                                  self.__WRITE_MARKER))
+
+    def __str__(self):
+        return "%s(public key: %s)" % (self.__class__.__name__,
+                                       pem.to_pem(self.__der,
+                                                  self.__WRITE_MARKER))
+
+    @error.returns_true_or_raises
+    def verify(self, signature_input, signature):
+        """Verifies the signature was created by the owner of this key.
+
+        Args:
+        - signature_input: The data that was originally signed.
+        - signature: An ECDSA SHA256 signature.
+
+        Returns:
+        - True if the signature verifies.
+
+        Raises:
+        - error.EncodingError: If the signature encoding is invalid.
+        - error.SignatureError: If the signature fails verification.
+        """
+        try:
+            return self.__key.verify(signature, signature_input,
+                                     hashfunc=hashlib.sha256,
+                                     sigdecode=ecdsa.util.sigdecode_der)
+        except ecdsa.der.UnexpectedDER:
+            raise error.EncodingError("Invalid DER encoding for signature %s",
+                                      signature.encode("hex"))
+        except ecdsa.keys.BadSignatureError:
+            raise error.SignatureError("Signature did not verify: %s",
+                                       signature.encode("hex"))
+
--- a/python/ct/crypto/verify.py
+++ b/python/ct/crypto/verify.py
@@ -1,23 +1,18 @@
 """Verify CT log statements."""
 
-import hashlib
 import io
 import struct
 
+from ct.crypto import ecdsa_keys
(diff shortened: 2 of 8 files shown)


python/ct/crypto/verify.py
```diff
@@ -13,8 +13,21 @@
...
 from ct.proto import ct_pb2
 from ct.serialization import tls_message
+
+import Crypto.Hash.SHA256
+import Crypto.PublicKey.RSA
+import Crypto.Signature.PKCS1_v1_5
 import ecdsa
 
+SUPPORTED_HASH_ALGORITHMS = (
+    ct_pb2.DigitallySigned.SHA256,
```
eranmes: What else should we be supporting?

python/ct/crypto/verify.py
```diff
@@ -178,49 +191,135 @@ def _create_dst_entry(sct, chain):
 
     return entry
 
-class LogVerifier(object):
-    """CT log verifier."""
-    __ECDSA_READ_MARKERS = ("PUBLIC KEY", "ECDSA PUBLIC KEY")
-    __ECDSA_WRITE_MARKER = "ECDSA PUBLIC KEY"
+class RsaPublicKey(object):
```
eranmes: Extract to a separate file, so only the LogVerifier remains here and each of the RSA/ECDSA specific code lives in a separate module.
Tidier, each file would be shorter and with narrower scope.

python/ct/crypto/verify.py
RJPercival: Good point, I've replaced the tuple with that one constant.

python/ct/crypto/ecdsa_keys.py
```diff
@@ -0,0 +1,76 @@
+from ct.crypto import error
+from ct.crypto import pem
+from ct.proto import client_pb2
+
+import hashlib
+import ecdsa
+
+class EcdsaPublicKey(object):
```
eranmes: Nit: Naming: That's more than a public key representation, that can also verify signatures. Name that accordingly? (EcdsaVerifier, EcdsaValidator, etc)

python/ct/crypto/ecdsa_keys.py
```diff
@@ -0,0 +1,76 @@
...
+            self.__key = ecdsa.VerifyingKey.from_der(self.__der)
+        except ecdsa.der.UnexpectedDER as e:
+            raise error.EncodingError(e)
+
+    def __repr__(self):
+        return "%r(public key: %r)" % (self.__class__.__name__,
+                                       pem.to_pem(self.__der,
+                                                  self.__WRITE_MARKER))
+
+    def __str__(self):
```
eranmes: since __str__ and __repr__ are duplicate, have a private method that both will invoke?

python/ct/crypto/verify.py
```diff
@@ -1,19 +1,23 @@
...
 
-import hashlib
 import io
 import struct
 
+from ct.crypto import ecdsa_keys
 from ct.crypto import error
 from ct.crypto import merkle
-from ct.crypto import pem
+from ct.crypto import rsa_keys
```
eranmes: These module names do not quite capture what the code in them does.

python/ct/crypto/verify_test.py
```diff
@@ -20,33 +20,15 @@ def read_testdata_file(test_file):
...
-        "2c3f087b4130").decode("hex")
-
-    default_key_info = client_pb2.KeyInfo()
-    default_key_info.type = client_pb2.KeyInfo.ECDSA
-    default_key_info.pem_key = (
-        "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAES0AfBk"
-        "jr7b8b19p5Gk8plSAN16wW\nXZyhYsH6FMCEUK60t7pem/ckoPX8hupuaiJzJS0ZQ0SEoJ"
-        "GlFxkUFwft5g==\n-----END PUBLIC KEY-----\n")
 
+class LogVerifierTest(object):
```
eranmes: Some documentation (in form of a comment) for this class is needed.

python/ct/crypto/verify_test.py
```diff
@@ -20,33 +20,15 @@ def read_testdata_file(test_file):
...
     def test_verify_sth(self):
-        verifier = verify.LogVerifier(LogVerifierTest.default_key_info)
-        self.assertTrue(verifier.verify_sth(LogVerifierTest.default_sth))
+        verifier = verify.LogVerifier(self.default_key_info)
```
eranmes: Nit: the 'default_' naming scheme for the key info, STH, etc, is not accurate as it's not really a  default - it is a per-test data.

python/ct/crypto/verify_test.py
```diff
@@ -349,6 +253,138 @@ def test_verify_embedded_scts_with_preca_valid_signature(self):
...
+        "dpSj/4DyHDc8wLG9wVmLqy1lk9fz4ATVmm+/1iN2Nk8jmctUKK2MFUtlWXZBSpym\n"
+        "97M7frGlSaQXUWyA3CqQUEuIJOmlEjKTBEiQAvpfDjCHjlV2Be4qTM6jamkJbiWt\n"
+        "gnYPhJL6ONaGTiSPm7Byy57iaz/hbckldSOIoRhYBiMzeNoA0DiRZ9KmfSeXZ1rB\n"
+        "8y8X5urSW+iBzf2SaOfzBvDpcoTuAaWx2DPazoOl28fP1hZ+kHUYvxbcMjttjauC\n"
+        "Fx+JII0dmuZNIwjfeG/GBb9frpSX219k1O4Wi6OEbHEr8at/XQ0y7gTikOxBn/s5\n"
+        "wQIDAQAB\n"
+        "-----END PUBLIC KEY-----\n")
+
+
+class LogVerifierEcdsaTest(LogVerifierTest, unittest.TestCase):
```
eranmes: Surprised and impressed this is working :)
Let's verify with somebody more knowledgable in Python that this is the right approach to base two separate tests on the same code.

eranmes:
Looks good save for minor comments.

python/ct/crypto/verify_test.py
RJPercival: I settled on this approach after discussing options with @pphaneuf and @plietar, and doing a reasonable amount of searching for other options. It seems parameterised tests aren't something that Python's unittest library really supports.

python/ct/crypto/verify_test.py
RJPercival: That's inherited from the original @ekasper code, but I'll see what better name I can come up with. Perhaps "{sth,key_info}_fixture"?

python/ct/crypto/verify.py
RJPercival: Any suggestions for better names? "{rsa,ecdsa}_keys" was the best I could come up with. Come to think of it, since the classes are now going to be called {Rsa,Ecdsa}Verifier, those module names are even less helpful. How about "verify_{rsa,ecdsa}"?

python/ct/crypto/ecdsa_keys.py
RJPercival: These two methods are subtlety different (one uses %s, the other %r). Nonetheless, I could extract their bulk to a private method. This is also the case in LogVerifier, so I'll do the same there too.

python/ct/crypto/verify.py
eranmes: SGTM.

eranmes:
LGTM.

PR #940, opened 2015-09-18
Fix typo that wouldn't compile on my system.

aeijdenberg:
I haven't tested a full build (ct seems Mac allergic right now) but this fix at least got me past this file.


pphaneuf:
LGTM

PR #941, opened 2015-09-18
Mac ports puts the files protoc need under /opt/local/include.

aeijdenberg:
There's likely a better way to make that portable, but this seems to work, albeit with warnings about missing dirs.


pphaneuf:
Could you file a bug to add some variables on that command-line (that we could override in `configure`, etc)? Thanks!

AlCutter:
Obsoleted by #958

PR #943, opened 2015-09-18
Convert the last (hopefully) Cert API to util::Status and update callers.

Martin2112:
I think this is the last change before cleanup.

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -42,33 +42,6 @@ static int X509_get_signature_nid(const X509* x) {
 }
 #endif
 
-namespace {
-// TODO. These helpers can be removed when Cert::Status has been removed.
-// They make it easier to do an incremental removal
-
-cert_trans::Cert::Status StatusOrBoolToCertStatus(StatusOr<bool> status) {
-  if (status.ok()) {
-    return status.ValueOrDie() ? cert_trans::Cert::TRUE
-                               : cert_trans::Cert::FALSE;
-  } else {
-    return cert_trans::Cert::ERROR;
-  }
-}
-
-
-StatusOr<bool> CertStatusToStatusOrBool(cert_trans::Cert::Status status) {
-  if (status == cert_trans::Cert::FALSE) {
-    return false;
-  } else if (status == cert_trans::Cert::TRUE) {
-    return true;
-  } else {
-    return util::Status(Code::UNKNOWN, "Unknown status");
-  }
-}
-
-// END of section to be cleaned up
-
-}
 
 namespace cert_trans {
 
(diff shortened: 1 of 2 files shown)


AlCutter:
`StatusOrBoolToCertStatus()` and `CertStatusToStatusOrBool()` are now unused, can you ding 'em? (they're causing the compiler to barf because `-Werror`)

cpp/log/cert.cc
```diff
@@ -323,20 +323,22 @@ StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
     return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
   }
 
-  void* ext_struct;
-  Status status = ExtensionStructure(NID_basic_constraints, &ext_struct);
+  const StatusOr<void*> status = ExtensionStructure(NID_basic_constraints);
```
AlCutter: yay const \o/ :)

Martin2112:
Ah. I've got a separate cleanup commit. Looks like I'll have to bring part of it into this then.

cpp/log/cert.cc
AlCutter: /status/basic_constraints/ ?

Martin2112: I'm glad you're pleased.

AlCutter: or maybe `ext_struct`?
(naming `StatusOr` instances for the `Or` thing rather than the `Status` helps the readability I think)

cpp/log/cert.cc
```diff
@@ -357,20 +359,22 @@ StatusOr<bool> Cert::HasExtendedKeyUsage(int key_usage_nid) const {
     return util::Status(Code::INTERNAL, "NID lookup failed");
   }
 
-  void* ext_struct;
-  Status status = ExtensionStructure(NID_ext_key_usage, &ext_struct);
+  const StatusOr<void*> status = ExtensionStructure(NID_ext_key_usage);
```
AlCutter: /status/ext_struct/

cpp/log/cert.cc
Martin2112: Done. Renamed to basic_constraints.

cpp/log/cert.cc
Martin2112: Done. Renamed to ext_status as we're only checking presence at this point. The structure is pulled out below.

cpp/log/cert.cc
```diff
@@ -627,17 +631,17 @@ util::Status Cert::OctetStringExtensionData(int extension_nid,
...
   }
 
-  void* ext_data;
-  Status status = ExtensionStructure(extension_nid, &ext_data);
-  if (status != TRUE) {
-    // TODO(Martin2112): This should return the result of ExtensionStructure()
-    // when it's been converted.
+  // Callers don't care whether extension is missing or invalid as they
+  // usually call this method after confirming it to be present.
+  const StatusOr<void*> status = ExtensionStructure(extension_nid);
```
AlCutter: `ext_struct`

cpp/log/cert.cc
```diff
@@ -682,30 +686,34 @@ StatusOr<X509_EXTENSION*> Cert::GetExtension(int extension_nid) const {
 }
 
 
-Cert::Status Cert::ExtensionStructure(int extension_nid,
-                                      void** ext_struct) const {
+util::StatusOr<void*> Cert::ExtensionStructure(int extension_nid) const {
   // Let's first check if the extension is present. This allows us to
   // distinguish between "NID not recognized" and the more harmless
   // "extension not found, found more than once or corrupt".
   const StatusOr<bool> status = HasExtension(extension_nid);
```
AlCutter: /status/ext/

cpp/log/cert.cc

Martin2112:
I think I fixed all your comments, though I probably should have held off until you'd finished. PTAL.

cpp/log/cert.cc
```diff
@@ -682,30 +660,34 @@ StatusOr<X509_EXTENSION*> Cert::GetExtension(int extension_nid) const {
...
-Cert::Status Cert::ExtensionStructure(int extension_nid,
-                                      void** ext_struct) const {
+util::StatusOr<void*> Cert::ExtensionStructure(int extension_nid) const {
   // Let's first check if the extension is present. This allows us to
   // distinguish between "NID not recognized" and the more harmless
   // "extension not found, found more than once or corrupt".
-  const StatusOr<bool> status = HasExtension(extension_nid);
-  if (!status.ok() || !status.ValueOrDie()) {
-    return StatusOrBoolToCertStatus(status);
+  const StatusOr<bool> ext_status = HasExtension(extension_nid);
```
AlCutter: has_ext?

AlCutter:
Couple more var name change suggestions, but LGTM otherwise.

cpp/log/cert.cc

Martin2112:
Latest renamings done and pushed, both those should now be fixed.

PR #944, opened 2015-09-18
Update cluster howto

AlCutter:


docs/ClusterHowto.md
```diff
@@ -105,27 +96,35 @@ Run the following commands:
    ```
 
 ## Updating the log server binaries
-If you're like to update the running CT Log Server code, then it's just a matter of building new Docker images, pushing them up to GCS, and restarting the containers.
+If you're like to update the running CT Log Server code, then it's just a
```
Martin2112: Pre-existing typo, probably "you'd like"

Martin2112:
Otherwise LGTM.

docs/ClusterHowto.md
AlCutter: Good catch, ta!

PR #947, opened 2015-09-20
Change FLAGS_trusted_root_certs to use SSL_CTX_set_default_verify_paths unless specifically overridden

aeijdenberg:
And change the default to "default" to detect.

Result: I can now start a server on Mac OS X.  Have not tested on Linux.

Unable to get tests to work on Mac before or after fix.

### DIFF
--- a/cpp/net/connection_pool.cc
+++ b/cpp/net/connection_pool.cc
@@ -45,7 +45,7 @@ DEFINE_int32(
     "When there are more than --url_fetcher_max_conn_per_host_port "
     "connections per host:port pair, any unused for at least this long will "
     "be removed.");
-DEFINE_string(trusted_root_certs, "/etc/ssl/certs/ca-certificates.crt",
+DEFINE_string(trusted_root_certs, "",
               "Location of trusted CA root certs for outgoing SSL "
               "connections.");
 DEFINE_int32(url_fetcher_max_conn_per_host_port, 4,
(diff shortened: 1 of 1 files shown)


cpp/net/connection_pool.cc
```diff
@@ -45,7 +45,7 @@ DEFINE_int32(
     "When there are more than --url_fetcher_max_conn_per_host_port "
     "connections per host:port pair, any unused for at least this long will "
     "be removed.");
-DEFINE_string(trusted_root_certs, "/etc/ssl/certs/ca-certificates.crt",
+DEFINE_string(trusted_root_certs, "default",
```
benlaurie: Don't use something that could be a legitimate path. Empty string perhaps?

aeijdenberg:
Made change as suggested.

aeijdenberg:
I'm not sure why Travis is moaning, I think the build was just flaky.  How can I tell it to try again?

aeijdenberg:
Rebased to try to kick travis.

Martin2112:
There is one test that doesn't work on OSX as it exercises undefined C++
behaviour, the others should pass. If not I might have another look at them.

I'm not sure removing the default Linux cert location from this flag is a
good idea, as this default is used in several other places.

Can you explain the problem this is addressing?

Martin

aeijdenberg:
The problem isn't a test (that's a meta-problem!), but rather starting a log server would fail for me on Mac OS X because the file "/etc/ssl/certs/ca-certificates.crt" did not exist (not at Mac right now so I don't have the exact incantation that I used).

openssl already has a notion of a default certificate location in a platform portable manner - it seems strange (wrong) to reinvent this.

On my Ubuntu-like workstation, openssl version -d shows:
OPENSSLDIR: "/usr/lib/ssl"

$ ls -l /usr/lib/ssl/certs
... /usr/lib/ssl/certs -> /etc/ssl/certs

Using SSL_CTX_set_default_verify_paths will actually use all of the certificates in /etc/ssl/certs and per http://manpages.ubuntu.com/manpages/karmic/man8/update-ca-certificates.8.html, ca-certificates.crt should be a duplicate of that data, and as such this should give identical behavior.

One could argue that we drop this flag completely, as if we always use SSL_CTX_set_default_verify_paths then the cert file used can be overridden by setting common openssl environment variables that off the top of my head I cannot recall.

Martin2112:
OK, I now realize this is for outgoing connections. It seems reasonable to
me that this should use the platform defaults.

From reading the docs it appears that your suggestion of removing the flag
and setting default verify paths is a good one.

Martin

aeijdenberg:
@AlCutter - Al, this code is your fault.  What do you think we should do?
(a) Keep the flag, but change the defaults as per this commit (shouldn't change behavior)
(b) Drop the flag entirely and always use SSL_CTX_set_default_verify_paths since I think users can get the same effect by setting SSL_CERT_FILE.

AlCutter:
@aeijdenberg, *everything* is *always* my fault, and I'm cool with that :)

In general I'm less of a fan of environment variables than explicit flags as it's a lot easier to spot the latter when someone comes to you with a problem - but, that said, having a flag whose default value implies using the system roots but then which may or may not use them depending on an environment variable seems like the worst of both worlds, so it seems like the best thing to do would be your option (b): if there's an "officially supported" OpenSSL way of setting this, and it sounds like there is, then we should probably do it that way.

benlaurie:
On Wed, 30 Sep 2015 at 10:36 Al Cutter <notifications@github.com> wrote:

> @aeijdenberg <https://github.com/aeijdenberg>, *everything* is *always*
> my fault, and I'm cool with that :)
>
> In general I'm less of a fan of environment variables than explicit flags
> as it's a lot easier to spot the latter when someone comes to you with a
> problem - but, that said, having a flag whose default value implies using
> the system roots but then which may or may not use them depending on an
> environment variable seems like the worst of both worlds, so it seems like
> the best thing to do would be your option (b): if there's an "officially
> supported" OpenSSL way of setting this, and it sounds like there is, then
> we should probably do it that way.
>

Not sure about that. Setting env variables for things buried deep in other
things can be quite exasperating.

Also, I think that if you use the env variable, you are expecting it to
override the default roots - that is its purpose.

So, disagree. Think we should have a flag whose default means "use
defaults/env".

aeijdenberg:
Ben, I think that's what the current proposed patch does.  By default we use the defaults (which includes env, because that's what openssl default does) or if you specify it, then we use that instead.

So if that's the case, I'll push the big green button as soon as anyone LGTMs it.

PR #949, opened 2015-09-23
Fix some OpenSSL memory leaks: STACK_OF() containers freed but not their contents.

phad:
Also, avoid returning StatusOr<T*>(nullptr) from (private) function calls.  This causes problems and isn't required in the Cert API in any case.

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -45,6 +45,7 @@ static int X509_get_signature_nid(const X509* x) {
 
 namespace cert_trans {
 
+
 // Convert string from ASN1 and check it doesn't contain nul characters
 string ASN1ToStringAndCheckForNulls(ASN1_STRING* asn1_string,
                                     const string& tag,
@@ -292,15 +293,13 @@ StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
 
   const StatusOr<void*> ext_struct(ExtensionStructure(NID_basic_constraints));
 
-  if (!ext_struct.ok()) {
+  if (ext_struct.status().CanonicalCode() == Code::NOT_FOUND) {
+    // No extension found
+    return false;
+  } else if (!ext_struct.ok()) {
     // Truly odd.
     LOG(ERROR) << "Failed to check BasicConstraints extension";
     return ext_struct.status();
-  }
-
-  if (!ext_struct.ValueOrDie()) {
-    // No extension found
-    return false;
   }
 
   // |constraints| is never null upon success.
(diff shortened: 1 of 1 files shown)


cpp/log/cert.cc
```diff
@@ -726,6 +730,16 @@ bool IsValidRedactedHost(const string& hostname) {
 }
 
 
+void freeOpenSslStackOfGeneralName(STACK_OF(GENERAL_NAME)* general_name_stack) {
```
AlCutter: Can these two go in an anonynamespace?

AlCutter: /free/Free/

cpp/log/cert.cc
```diff
@@ -860,6 +874,16 @@ bool Cert::ValidateRedactionSubjectAltNameAndCN(int* dns_alt_name_count,
 }
 
 
+void freeOpenSslStackOfAsn1Integer(STACK_OF(ASN1_INTEGER)* asn1_integer_stack) {
```
AlCutter: anonyspace & /free/Free/

AlCutter: Tears of sadness that you can't do with with a templated method :/

cpp/log/cert.cc
AlCutter: Actually, maybe these could go into `util/openssl_util.{cc,h}` ?

phad: Done the anonymous namespace and the case change.

cpp/log/cert.cc
phad: Done the anonymous namespace and the case change.
ACK re. tears.

cpp/log/cert.cc
phad: Fixed the casing of the validateRedactionSubjectAltNames function as well, and moved that into the anonymous namespace.

cpp/log/cert.cc
```diff
@@ -292,13 +292,14 @@ StatusOr<bool> Cert::HasBasicConstraintCATrue() const {
...
-  if (!ext_struct.ok()) {
+  if (!ext_struct.ok()
+      && ext_struct.status().CanonicalCode() != Code::NOT_FOUND) {
     // Truly odd.
     LOG(ERROR) << "Failed to check BasicConstraints extension";
     return ext_struct.status();
   }
 
-  if (!ext_struct.ValueOrDie()) {
+  if (ext_struct.status().CanonicalCode() == Code::NOT_FOUND) {
```
AlCutter: You could move this check above the one at L295, and then that one could revert to simply `if (!ext_struct.ok())`, might be a little easier to read.
Same for the other instance below at ~L322

phad: Yeah, good call - done.

Issue #950, opened 2015-09-24
Mirroring and ct client requests fail for logs which don't send a Content-Length header.


AlCutter:
The underlying library used for making HTTP requests, libevhtp, appears to assume a zero length body if no `Content-Length` header is received from the target Log, even when the server sends a `Connection: close` header.

pphaneuf:
Do you have an example of a URL (on our server or elsewhere on another log) that caused this? EOF with a chunked encoding is an error (you're supposed to receive a zero-sized chunk to signal the end).

AlCutter:
It's not chunked, just `Connection: close`:

```
$ curl -i https://ct.ws.symantec.com/ct/v1/get-sth
HTTP/1.1 200 OK
Server: nginx/1.6.2
Date: Tue, 06 Oct 2015 15:36:48 GMT
Content-Type: application/json;charset=UTF-8
Connection: close
Set-Cookie: JSESSIONID=222B7BF7004BCC01405644D070269B3A; Path=/ct/; HttpOnly
Expires: Wed, 07 Oct 2015 15:36:48 GMT
Cache-Control: max-age=86400
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Content-Disposition: attachment; filename=f.txt
Alternate-Protocol: 443:quic,p=0.5
Cache-Control: public
status: 200
version: HTTP/1.1

{"tree_size":9222,"timestamp":1444136400307,"sha256_root_hash":"ejr32fKeIGGWyTi4Xr8oDNr0/voWjy7yvKi81aAPgzk=","tree_head_signature":"BAMARjBEAiACWcdaJKZp35cJOSDihRZgFaQ1uVwdHA1J+jUjol+YlwIgMGCjvekMO02kk+AwJJGCHNliNC7UcL01v8PJcuDFR0U="}
```

pphaneuf:
Whew, that's pretty exotic. Note the tasteful inclusion of those `status` and `version` headers... :wink: 

Yeah, technically allowed, although comes with a warning in the HTTP/1.1 RFC (as you can't distinguish between the end of the message or a network failure, this is there mainly for backward compatibility with HTTP/1.0), but an interesting choice, nonetheless. :smiley:

aeijdenberg:
Might be worth asking Symantec to confirm their config is as expected, I
bet there's some config in nginx that could be tweaked to resolve.  I can't
[...]

PR #951, opened 2015-09-24
EOF is not an error, and add some more debug logging.

AlCutter:
Half of a work-around for issue #950 .

### DIFF
--- a/cpp/net/url_fetcher.cc
+++ b/cpp/net/url_fetcher.cc
@@ -157,6 +157,7 @@ void State::RunRequest() {
   VLOG(1) << "evhtp_make_request(" << conn_.get()->connection() << ", "
           << http_req << ", " << verb << ", \"" << request_.url.PathQuery()
           << "\")";
+  VLOG(2) << request_;
   if (evhtp_make_request(conn_->connection(), http_req, verb,
                          request_.url.PathQuery().c_str()) != 0) {
     VLOG(1) << "evhtp_make_request error";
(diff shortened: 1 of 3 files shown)


cpp/net/url_fetcher.cc
```diff
@@ -245,6 +246,8 @@ void State::RequestDone(evhtp_request_t* req) {
               body_length);
   response_->body.swap(body);
 
+  VLOG(2) << *response_;
```
phad: Hopefully this will include logging the response header.  I assume it would as you reference response_->body above.

cpp/util/init.cc
```diff
@@ -18,12 +18,42 @@ using std::string;
...
+}  // namespace
+
+
 void InitCT(int* argc, char** argv[]) {
   google::SetVersionString(cert_trans::kBuildVersion);
   google::ParseCommandLineFlags(argc, argv, true);
   google::InitGoogleLogging(*argv[0]);
   google::InstallFailureSignalHandler();
 
+  event_set_log_callback(&LibEventLog);
```
phad: So I take it you are redirecting libevent logging to glog LOG() macros, and otherwise it'd have been dropped silently?

cpp/net/url_fetcher.cc
AlCutter: Yeah, there's an `operator<<` defined for that so it'll print all the fields nicely formatted.

cpp/util/init.cc
AlCutter: Yes, though I think you need to have a version of the libevent library compiled with debug support to get much out of it.

PR #952, opened 2015-09-24
Generate Java source file for Android from the list of logs.

plietar:
This file will be included in conscrypt for Android


plietar:
Generated output :
https://android-review.googlesource.com/#/c/171932/6/src/main/java/org/conscrypt/ct/KnownLogs.java

DO NOT MERGE YET
I may need to change the format based on reviews on the Android side.

python/utilities/log_list/print_log_list.py
```diff
@@ -131,6 +134,75 @@ def generate_code_for_chromium(json_log_list, output_file):
     f.write("\nconst size_t kNumKnownCTLogs = %d;\n" % len(logs))
     write_cpp_footer(f, include_guard)
 
+def write_android_header(f):
```
eranmes: Break this into a separate module (within this directory).

eranmes: While you're at it, please look into breaking the C++ code generation to a module of its own.

plietar:
I've changed the output format to be more generic rather than just file conscrypt.
The package and class names still default to conscrypt, but can be overridden with a flag.
The copyright notice is still the one from AOSP.

Also split out the code generation is separate modules

python/utilities/log_list/java_generator.py
```diff
@@ -0,0 +1,73 @@
...
+def _write_java_class(f, logs, cls):
+    descriptions = ('        "%s",\n' % log["description"] for log in logs)
+    urls = ('        "%s",\n' % log["url"] for log in logs)
+    keys = (_encode_key(log["description"], base64.decodestring(log["key"]))
+            for log in logs)
+
+    f.write(
+        "public final class %(cls)s {\n"
+        "    public static final int LOG_COUNT = %(count)d;\n"
+        "    public static final String[] LOG_DESCRIPTIONS = new String[] {\n"
```
eranmes: Seems like some lines here are too long, please run it through pylint.

python/utilities/log_list/java_generator.py
```diff
@@ -0,0 +1,73 @@
...
+        "package %(package)s;\n\n" %
+        {"year": year,
+         "package": package})
+
+def _encode_key(description, key):
+    unsigned = (ord(c) for c in key)
+    signed = (u - 256 if u > 127 else u for u in unsigned)
+    return ("        // " + description + "\n"
+            "        new byte[] {\n"
+            + textwrap.fill(", ".join("%d" % s for s in signed), width=85,
```
eranmes: Could you assign the result from this call to a variable you'll include in the string generated here, so the code is slightly more readable?

python/utilities/log_list/java_generator.py
```diff
@@ -0,0 +1,73 @@
...
+        "%(descriptions)s"
+        "    };\n"
+        "    public static final String[] LOG_URLS = new String[] {\n"
+        "%(urls)s"
+        "    };\n"
+        "    public static final byte[][] LOG_KEYS = new byte[][] {\n"
+        "%(keys)s"
+        "    };\n"
+        "}\n" % {
+            "cls": cls,
```
eranmes: nit: Prefer a long name than a shorthand.

eranmes:
A few minor comments, LGTM otherwise - please rebase and let me know so I'll merge and commit.

PR #953, opened 2015-09-24
Switch to using "cryptography" Python lib

RJPercival:
Replaces "ecdsa" and "pycrypto" libraries; the former is considered to be poorly implemented, whilst the latter is poorly maintained.

This library correctly raises exceptions for all forms of invalid ASN.1 in STHs, unlike the "ecdsa" library, so some tests have been modified to account for this and a TODO removed.

See: https://cryptography.io

### DIFF
--- a/python/ct/crypto/verify_ecdsa.py
+++ b/python/ct/crypto/verify_ecdsa.py
@@ -1,16 +1,12 @@
 from ct.crypto import error
 from ct.crypto import pem
-from ct.crypto.asn1 import types
 from ct.proto import client_pb2
 
-import hashlib
-import ecdsa
-
-class _ECDSASignature(types.Sequence):
-    components = (
-            (types.Component("r", types.Integer)),
-            (types.Component("s", types.Integer))
-    )
+import cryptography
+from cryptography.hazmat.backends import default_backend as crypto_backend
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import serialization
 
 class EcdsaVerifier(object):
     """Verifies ECDSA signatures."""
@@ -20,11 +16,6 @@ class EcdsaVerifier(object):
     # The hash algorithm used for this public key."""
     HASH_ALGORITHM = client_pb2.DigitallySigned.SHA256
 
-    # Markers to look for when reading a PEM-encoded ECDSA public key."""
-    __READ_MARKERS = ("PUBLIC KEY", "ECDSA PUBLIC KEY")
-    # A marker to write when writing a PEM-encoded ECDSA public key."""
-    __WRITE_MARKER = "ECDSA PUBLIC KEY"
-
     def __init__(self, key_info):
         """Creates a verifier that uses a PEM-encoded ECDSA public key.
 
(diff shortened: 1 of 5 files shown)


python/ct/crypto/verify_ecdsa.py
```diff
@@ -26,22 +24,27 @@ def __init__(self, key_info):
...
         Raises:
         - PemError: If the key has an invalid encoding
+        - UnsupportedAlgorithmError: If the key uses an unsupported algorithm
         """
         if (key_info.type != client_pb2.KeyInfo.ECDSA):
             raise error.UnsupportedAlgorithmError(
                 "Expected ECDSA key, but got key type %d" % key_info.type)
 
-        # Will raise a PemError on invalid encoding
-        self.__der, _ = pem.from_pem(key_info.pem_key, self.__READ_MARKERS)
```
phad: Is self.__der no longer required?

python/ct/crypto/verify_rsa.py
```diff
@@ -27,22 +24,27 @@ def __init__(self, key_info):
...
         Raises:
         - PemError: If the key has an invalid encoding
+        - UnsupportedAlgorithmError: If the key uses an unsupported algorithm
         """
         if (key_info.type != client_pb2.KeyInfo.RSA):
             raise error.UnsupportedAlgorithmError(
                 "Expected RSA key, but got key type %d" % key_info.type)
 
-        # Will raise a PemError on invalid encoding
-        self.__der, _ = pem.from_pem(key_info.pem_key, self.__READ_MARKERS)
```
phad: Same qn about __der.   Outside here it looks like it was only used in __repr__

python/ct/crypto/verify_test.py
```diff
@@ -333,18 +333,17 @@ def test_verify_sth_for_bad_asn1_length(self):
             sth_fixture.tree_head_signature[:i] +
             chr(ord(sth_fixture.tree_head_signature[i]) - 1) +
             sth_fixture.tree_head_signature[i+1:])
-        self.assertRaises(error.EncodingError, verifier.verify_sth, sth)
+        self.assertRaises(error.SignatureError, verifier.verify_sth, sth)
 
         # Increasing the length means there are not enough ASN.1 bytes left to
-        # decode the sequence, however the ecdsa module silently slices it.
-        # TODO(ekasper): contribute a patch to upstream and make the tests fail
+        # decode the sequence.
```
phad: This is cool.

python/ct/crypto/verify_test.py
```diff
@@ -333,18 +333,17 @@ def test_verify_sth_for_bad_asn1_length(self):
...
         sth = client_pb2.SthResponse()
         sth.CopyFrom(sth_fixture)
         sth.tree_head_signature = (
             sth_fixture.tree_head_signature[:i] +
             chr(ord(sth_fixture.tree_head_signature[i]) + 1) +
             sth_fixture.tree_head_signature[i+1:])
-        self.assertTrue(verifier.verify_sth(sth))
+        self.assertRaises(error.SignatureError, verifier.verify_sth, sth)
```
phad: This, I mean.

eranmes:
Per @ekasper 's comments on cryptography.io , undoing the LGTM until we further discuss how to proceed with it.

python/ct/crypto/verify_ecdsa.py
RJPercival: No, because the cryptography library can load a public key directly from PEM, so there's no need to manually convert to DER first. It can also convert straight back to PEM, so self.__der isn't required in __repr__() either.

RJPercival:
@ekasper has rescinded her reservations about cryptography.io now. However, we probably want to sit on this PR a bit longer until other Googlers have been using cryptography.io successfully for a little while (and possibly worked out any kinks).

RJPercival:
I think we've waited more than long enough to de-risk this now.

eranmes:
Re-adding original LGTM, approval.

PR #956, opened 2015-09-29
Unpack the ASN.1 sequence in our extension via a different method

Martin2112:
The original way used (ASN1_seq_unpack_ASN1_INTEGER) is not supported by BoringSSL. Add some extra type constraints on the sequence data in the extension and some tests to verify them.


cpp/log/cert.cc
```diff
@@ -918,21 +902,35 @@ util::Status Cert::IsValidWildcardRedaction() const {
...
+  }
+
+  // Unpack the extension contents, which should be SEQUENCE OF INTEGER.
+  // For compatibility we unpack any sequence and check integer type as we go.
+  // Don't pass the pointer from the extension directly as it gets incremented
+  // during parsing.
+  const unsigned char* sequence_data(
+      const_cast<const unsigned char*>(exty.ValueOrDie()->value->data));
+  STACK_OF(ASN1_TYPE)*
+      const asn1_types(static_cast<STACK_OF(ASN1_TYPE)*>(d2i_ASN1_SEQUENCE_ANY(
```
phad: might this be easier to read if the d2i_ASN1 ... is folded to the next line?

cpp/log/cert.cc
```diff
@@ -918,21 +902,35 @@ util::Status Cert::IsValidWildcardRedaction() const {
...
-    const int num_integers = sk_ASN1_INTEGER_num(integers);
+  if (asn1_types) {
+    const int num_integers(sk_ASN1_TYPE_num(asn1_types));
 
     // RFC text says there MUST NOT be more integers than there are DNS ids
     if (num_integers > dns_alt_name_count) {
       LOG(WARNING) << "Too many integers in extension: " << num_integers
                    << " but only " << dns_alt_name_count << " DNS names";
-      FreeOpenSslStackOfAsn1Integer(integers);
+      sk_ASN1_TYPE_pop_free(asn1_types, ASN1_TYPE_free);
```
phad: So are you quite sure that sk_FOO_pop_free will free both the container and the contained elements?

Martin2112: I have the appropriate level of sureness anyone using OpenSSL api calls can have. See note in https://wiki.nikhef.nl/grid/How_to_handle_OpenSSL_and_not_get_hurt_and_what_does_that_library_call_really_do

cpp/log/cert.cc
Martin2112: clang_format and Eclipse formatter disagreed on this one. I've now redone it manually, is it OK?

cpp/log/cert.cc
phad: Yes, I see :)  That's a handy page, thanks.

cpp/log/cert.cc
phad: Looks good, thanks

PR #957, opened 2015-09-29
Add a Fedora quickstart README with build instructions for F22

Martin2112:
Tested in a VM from a clean install of F22 Workstation on X64.


benlaurie:
This is why we need the gclient thing. But in the meantime...

Martin2112:
Yep. I'll test F22 with gclient when the current work is submitted.

PR #958, opened 2015-09-29
Gclient for all

AlCutter:
Based on @benlaurie's original gclient PR, this _works_* for FreeBSD 10, OSX 10.10, and Ubuntu 14.04.
I'm sure it's t3h suck in places, but we can iterate.

_* Worked once. For me. Here._

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Build CT server C++ code:
     ./configure GTEST_DIR=../gmock-1.7.0/gtest GMOCK_DIR=../gmock-1.7.0 \
         CPPFLAGS="-I../libevhtp-1.2.10 -I../libevhtp-1.2.10/evthr \
         -I../libevhtp-1.2.10/htparse" LDFLAGS=-L../libevhtp-1.2.10
-    make check 
+    make check
 
 Build and test Java code:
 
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -61,13 +61,47 @@ Build and test Python code:
...
+
+This is the recommended method for all platforms as it gives you a reproducible
+build.
+
+Known to work on FreeBSD 10, OS X (10.10) [tested with XCode + brew installation
+of above deps], and Ubuntu 14.04.
+
+    $ mkdir ct  # or whatever directory you prefer
+    $ cd ct
+    $ gclient config git@github.com:google/certificate-transparency.git
```
benlaurie: Add note (somewhere) that if you want to clone from a branch, for example, then you need to do:

`gclient config --name="certificate-transparency" git@github.com:google/certificate-transparency.git@branch`

README.md
```diff
@@ -61,13 +61,47 @@ Build and test Python code:
...
+    $ gclient sync
+    $ [g[nu]make -C certificate-transparency cpp
```
benlaurie: Strictly this should be

`{.g.gnu}make`.

benlaurie: Also, `cpp` is no longer the right target, right? I used `check`.

build.gclient
```diff
@@ -0,0 +1,41 @@
...
+	make -C openssl -f `pwd`/certificate-transparency/build/Makefile.openssl INSTALL_DIR=$(INSTALL_DIR)
+
+protobuf:
+	certificate-transparency/build/rebuild_protobuf
+
+libevent:
+	certificate-transparency/build/rebuild_libevent
+
+libevhtp:
+	certificate-transparency/build/rebuild_libevhtp
```
benlaurie: Add TODO to make these into makefiles?

README.md
AlCutter: Added a comment about `make` vs. `gmake` vs. `gnumake` instead

AlCutter: Updated to target check

README.md
AlCutter: done.

build.gclient
AlCutter: That's implied by the TODO on L1, but I'll add an explicit TODO too.

AlCutter:
PTAL, some tests fail on OS X and FBSD, but that's a problem with the tests and I've filled an issue for those (#962).

There's a weird bug in gclient on OSX where it hangs at the end once everything is finished; looks like it's not realising that gnumake has exited (it's gone zombie) and just sits there, but killing it and then continuing with the manual `gnumake check` process works as expected (modulo the broken tests.)

benlaurie:
LGTM

Martin2112:
OK I just had a really horribly confusing time until I remembered my firewall rate limits SSH connections.

I think it might be better to sync the repository via HTTPS as this currently doesn't work unless you have set up SSH keys on Github.

Martin2112:
Tried it on Fedora 22 and CentOS and got a similar error on each. Looks like the aclocal.m4 file for glog does not get updated, which should happen when autoreconf -i runs. F22 has automake v15 and Centos has automake v13. The build stops with:

make[2]: Entering directory `/home/martin/ct-gclient/glog'
CDPATH="${ZSH_VERSION+.}:" && cd . && aclocal-1.14 -I m4
/bin/sh: aclocal-1.14: command not found
make[2]: *** [aclocal.m4] Error 127
make[2]: Leaving directory `/home/martin/ct-gclient/glog'
make[1]: *** [all] Error 2

The aclocal.m4 file says it was generated by 14.1, which is the Ubnuntu version:
# generated automatically by aclocal 1.14.1 -*- Autoconf -*-

pphaneuf:
I see the running of `autoreconf -i` is the command for a `configure: configure.ac` rule... But `glog` has the `configure` script pre-generated in their git repo (which isn't recommended, but here we are). So it's quite likely that `make` doesn't run `autoreconf -i` at all.

AlCutter:
These are likely being addressed in the travis_gclient minefie^WPR which I'm currently _enjoying_ putting together.

Martin2112:
Yep. We also discovered that it was picking up the system gflags lib when building glog, which wasn't obvious. It broke on F22 because the /usr/lib64 version is built with a different namespace.

PR #963, opened 2015-10-01
Travis gclient

AlCutter:


.travis.yml
```diff
@@ -1,47 +1,72 @@
-
+sudo: false
```
pphaneuf: Sweet!

.travis.osx.before_install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
pphaneuf: As mentioned in person, might be nicer to have this as a script in `build` (it's not gclient-related, but it *is* build-related!) than as a dot file?

.travis.yml
```diff
@@ -1,47 +1,72 @@
...
+os:
+  - linux
+  - osx
```
pphaneuf: I'd like it better if `language` remained the first one, because for Travis, it's the most important (it sets the defaults for all the other things, picks the base image, etc)?

The `sudo: false` could be immediately after, since it's pretty important as well, but `os` should definitely be together with `compiler` and `env`, because these three are the ones which create the matrix of builds.

And I'd like a newline between top-level things. :smiley:

.travis.yml
```diff
@@ -1,47 +1,72 @@
...
 language: cpp
-
 compiler:
  - clang
 
 env:
- - GOPATH=${TRAVIS_BUILD_DIR}/go CPPFLAGS="-I/usr/src/gmock/include" CXXFLAGS="-Wall -Werror"
+  - CXXFLAGS="-std=c++11"
```
pphaneuf: Huh? I don't understand why some variables have moved to `before_install`, while `CXXFLAGS` remains here (with a rather peculiar value, since the `configure` script should be able to figure that out by itself?!?)...

I agree that I kind of abused `env`, I just wanted to set some environment variables, and `env` is more to add an additional dimension to the build matrix. Move this to an `export CXXFLAGS=...` in `before_install`?

pphaneuf: Actually, you can do this, it seems:

```yaml
env:
  global:
    - CXXFLAGS="-std=c++11"
    - INSTALL_DIR=$TRAVIS_BUILD_DIR/../install
    - INSTALL_DIR=$TRAVIS_BUILD_DIR/../install
```

There's a separate `matrix` subkey to do what `env` normally does otherwise?

http://docs.travis-ci.com/user/environment-variables/#Global-Variables

.travis.yml
```diff
@@ -1,47 +1,72 @@
...
- - unzip -d /tmp /tmp/gmock-1.7.0.zip
- - sudo mv /tmp/gmock-1.7.0 /usr/src/gmock
- - wget https://github.com/ellzey/libevhtp/archive/1.2.10.zip -O /tmp/libevhtp-1.2.10.zip
- - unzip -d /tmp/ /tmp/libevhtp-1.2.10.zip
- - mkdir /tmp/libevhtp-1.2.10/build
- - pushd /tmp/libevhtp-1.2.10/build && cmake -DEVHTP_DISABLE_REGEX:STRING=ON -DCMAKE_C_FLAGS="-fPIC" .. && sudo make install; popd
- - sudo pip install -r python/requirements.txt
- - sudo ln -s protobuf-java.jar /usr/share/java/protobuf.jar
+ - if [ "$CXX" == "clang++" ] && which clang++-3.6; then export CXX="clang++-3.6" CC="clang-3.6"; fi
+ - if [ "$CXX" == "g++" ] && which g++-4.8; then export CXX="g++-4.8" CC="gcc-4.8"; fi
```
pphaneuf: This wackiness is for supporting Mac, I'd guess? If so, maybe put a comment to explain why you don't just use the provided compiler (which could be controlled)?

.travis.yml
```diff
@@ -1,47 +1,72 @@
...
+ - export INSTALL_DIR=$TRAVIS_BUILD_DIR/../install
+ - mkdir -p ${INSTALL_DIR}/include ${INSTALL_DIR}/lib ${INSTALL_DIR}/java
+ - echo "INSTALL_DIR=${INSTALL_DIR}"
+ - export PROTOC="${INSTALL_DIR}/bin/protoc"
+ - export GOPATH=${INSTALL_DIR}/go
  - mkdir -p $GOPATH/src/github.com/google
- - ln -s $PWD $GOPATH/src/github.com/google
+ - ln -sf $PWD $GOPATH/src/github.com/google
+ - echo "jars.protobuf = ${INSTALL_DIR}/java/protobuf.jar" > local.properties
+ - echo "protoc.bin = ${PROTOC}" >> local.properties
```
pphaneuf: Hmm, maybe put a TODO to have `configure` do this? And by "this", I mean have a substitution in `build.properties` (rename it `build.properties.in` and have `configure`fix it up).

.travis.yml
```diff
@@ -1,47 +1,72 @@
...
+
+install:
+ - autoconf --version
+ - pushd ..
```
pphaneuf: I find this a little bit bizarre, because this might be at the top level of a home directory, say, that it wouldn't be completely unreasonable. Subdirectories of the source directory would probably do fine for most of this...

Let's not change that now, though, this was hard enough. :stuck_out_tongue:

.travis.yml
```diff
@@ -1,47 +1,72 @@
...
+ - mkdir -p ct
+ - wget http://central.maven.org/maven2/com/google/protobuf/protobuf-java/2.6.0/protobuf-java-2.6.0.jar -O ${INSTALL_DIR}/java/protobuf.jar
+ - if [ ! -e depot_tools/gclient ]; then git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git; fi
+ - export PATH="$PATH:/usr/local/bin:"`pwd`/depot_tools
+ - cd ct
+ - ln -s ../certificate-transparency certificate-transparency
```
pphaneuf: What have we done to deserve this fate?

DEPS
```diff
@@ -23,10 +22,13 @@ deps_overrides = {
   "freebsd10": {
      "googlemock": 			 "https://github.com/AlCutter/googlemock-fbsd.git@1.7.0",
      "googlemock/gtest": "https://github.com/AlCutter/googletest-fbsd.git@1.7.0",
-     "protobuf/gtest": "https://github.com/benlaurie/googletest.git@1.5.0-fix",
-     "protobuf": "https://github.com/benlaurie/protobuf.git@2.6.1-fix",
-     "glog": "https://github.com/benlaurie/glog.git@0.3.4-fix",
-     "ldns": "https://github.com/benlaurie/ldns.git@1.6.17-fix",
+     #"openssl": 				 "https://github.com/benlaurie/openssl.git@fd5d2ba5e09f86e9ccf797dddd2d09ac8e197e35", # 1.0.2-freebsd
```
pphaneuf: Commented out?

DEPS
```diff
@@ -52,12 +54,72 @@ else:
...
+        "name": "leveldb",
+        "pattern": "^leveldb/",
+        "action": [ make, "-f", os.path.join(here, "certificate-transparency/build.gclient"), "_leveldb" ],
+    },
+    {
+        "name": "json-c",
+        "pattern": "^json-c/",
+        "action": [ make, "-f", os.path.join(here, "certificate-transparency/build.gclient"), "_json-c" ],
+    },
+    # Do this last
```
pphaneuf: Ugh, I "love" that this implies the lack of concurrency...

Makefile.am
```diff
@@ -18,7 +19,10 @@ BUILT_SOURCES = \
 BUILD_VERSION = `git describe --abbrev=0 --dirty --always --tags`
 
 AM_TESTS_ENVIRONMENT = \
-export LD_LIBRARY_PATH=$(LIBRARY_PATH);
+export LD_LIBRARY_PATH=$(INSTALL_DIR)/lib;
+
+export PROTOC
```
pphaneuf: Could you add a TODO to change these to `autoconf` substitutions?

build.gclient
```diff
@@ -1,42 +1,63 @@
...
-tcmalloc:
+_libunwind:
+	if [ -d libunwind ]; then \
+		$(MAKE) -C libunwind -f ../certificate-transparency/build/Makefile.unwind; \
+		(cd libunwind && git checkout --); \
+	fi
+
+_tcmalloc:
 	$(MAKE) -C tcmalloc -f ../certificate-transparency/build/Makefile.tcmalloc
+	(cd tcmalloc && git checkout --)
```
pphaneuf: No need for parenthesis here? They *are* needed for the `_libunwind` command, though, I agree.

build.gclient
```diff
@@ -1,42 +1,63 @@
...
 
-openssl:
-	make -C openssl -f `pwd`/certificate-transparency/build/Makefile.openssl INSTALL_DIR=$(INSTALL_DIR)
+_openssl:
+	$(MAKE) -C openssl -f `pwd`/certificate-transparency/build/Makefile.openssl
+	(cd openssl && git checkout -- apps/progs.h crypto/bn/bn_prime.h)
```
pphaneuf: This, and below, of course...

build.gclient
```diff
@@ -1,42 +1,63 @@
...
-json-c:
-	certificate-transparency/build/rebuild_json-c
+_json-c:
+	$(MAKE) -C json-c -f ../certificate-transparency/build/Makefile.json-c
+	(cd json-c && git checkout --)
 
-configure-ct: tcmalloc openssl protobuf libevent libevhtp gflags glog ldns sqlite3 leveldb json-c
-	certificate-transparency/build/configure-ct INSTALL_DIR=$(INSTALL_DIR)
+_configure-ct:
+	certificate-transparency/build/configure-ct
```
pphaneuf: Could you put a TODO here that this script could be simply inlined?

pphaneuf:
Done with the review, a few niggles, but nothing serious.

.travis.osx.before_install.sh
AlCutter: done

.travis.yml
AlCutter: I'm not entirely sure that "sets default and picks base image" is true when the os is `osx` - I think there is only one `osx` image, so... I don't agree with your ordering reasoning. 
However, I also don't care which order these things are in, and since you do (likely because you haven't spent the last two weeks looking at this stupid file) I'll happily move it :)

.travis.yml
AlCutter: That's not really the Java/ant way - you do that yourself in the not-checked-in `local.properties`.
How will the configure find the right jar if it's not being run on travis where we know its location?

.travis.yml
AlCutter: It's not; it's at `${HOME}/${GIT_OWNER}/${GIT_PROJECT}`

Creating subdirs of the checked-out source *might* work ok, too.

.travis.yml
AlCutter: Well, `gclient` needs to know what it's going to be checking out, and if we don't supply an "unmanaged" `certificate-transparency` directory containing the `DEPS` file, then it'll helpfully check out the repo spec'd by the `gclient config` line below, which, obviously, won't be the particular repo/branch that we're actually trying to test.

I spent some time, with @daviddrysdale's help, trying to coax gclient to check out the correct thing for `certificate-transparency/` to no avail, so we symlink it in, which works just fine.

DEPS
AlCutter: dinged

DEPS
AlCutter: Yeah, that's sad making, and is why I had originally moved all of this into the one `gclient.build` Makefile so that  I could outline the deps.  But, as I told you about offline, that brings its own set of issues (mainly around compile time on Travis).

Probably more can be done in this area later on, though.

Makefile.am
AlCutter: done

build.gclient
AlCutter: Removed

build.gclient
AlCutter: done.

.travis.yml
AlCutter: Changed a bunch of stuff around this, I'm doing an explicit `matrix` now, which makes things easier in the `before_install` section.

Also, the `std=c++11` was a hangover from the gcc versions vs. clang versions vs. stdlib++ versions fight, and since I finally won that particular battle I've removed it.

.travis.yml
pphaneuf: It's possible that they've only made one mega-image for `osx` with *everything* in it, but on `linux`, when `language` is `cpp`, you either have just one or no JVM at all, for example, while the `java` one has a bunch, and the `go` setting gives you a bunch of different golang versions, `cpp` has just one, etc...

You could make any of them compile C++ or do the rest by installing more packages, but since I'm no Java fiend and wasn't going to test over all the available JVMs, I picked the option that looked the smallest and install what it was missing, hoping it'd help with booting up faster. :wink:

.travis.yml
pphaneuf: WHOA.

.travis.yml
pphaneuf: I'm the boorish type that would just say "run ./configure!" even to Java people. :wink: 

And you say "know it's location", but I think you're quite optimistic. I bet `protoc` is in some very "amusing" location on FreeBSD... :wink: 

But I can just keep that TODO in my head, where it might rot, but that might also be correct, so... :smiley:

.travis.yml
```diff
@@ -1,47 +1,92 @@
...
 language: cpp
 
-compiler:
- - clang
-
 env:
- - GOPATH=${TRAVIS_BUILD_DIR}/go CPPFLAGS="-I/usr/src/gmock/include" CXXFLAGS="-Wall -Werror"
+  global:
+    - INSTALL_DIR=${TRAVIS_BUILD_DIR}/../install
+    - PROTOC="${INSTALL_DIR}/bin/protoc"
```
pphaneuf: This is all right (especially as I like make variables like CC, CXX, LD, and, now, PROTOC), but setting `PATH="${PATH}:${INSTALL_DIR}/bin"` might have worked in a slightly more generic way (if more tools that we build and install were to be needed).

Don't change, it's all right! We've waited enough... :wink:

.travis.yml
pphaneuf: They have a specific note that says not to depend on `HOME`, so maybe this will have to do... It should break loudly if it stops working, anyway. Or we could change to subdirs, but no rush.

.travis.yml
pphaneuf: Yeah, that's exactly what I thought was happening, this was more of an existential/rhetorical question. :wink:

.travis.yml
pphaneuf: And you even set `PATH` later on, but whatever, I'm fine with this.

.travis.yml
```diff
@@ -1,47 +1,92 @@
...
- - ./autogen.sh
 
 script:
- - ./configure && make -j$(getconf _NPROCESSORS_ONLN) check VERBOSE=1
+ - make -j$(getconf _NPROCESSORS_ONLN) check VERBOSE=1
  - ant build test
  - make -C python test
- - go test -v ./go/...
+ - pushd ${GOPATH}
+ - LD_LIBRARY_PATH=${INSTALL_DIR}/lib CGO_CPPFLAGS="-I${INSTALL_DIR}/include" CGO_LDFLAGS="-L${INSTALL_DIR}/lib" go test -v ./src/...
```
pphaneuf: An issue for another time: we need to make the content of `${INSTALL_DIR}/lib` only be static libraries, and skip this `LD_LIBRARY_PATH`nonsense. This way lies madness...

Makefile.am
```diff
@@ -18,7 +19,11 @@ BUILT_SOURCES = \
 BUILD_VERSION = `git describe --abbrev=0 --dirty --always --tags`
 
 AM_TESTS_ENVIRONMENT = \
-export LD_LIBRARY_PATH=$(LIBRARY_PATH);
+	LD_LIBRARY_PATH=$(INSTALL_DIR)/lib
+
+# TODO(pphaneuf): change these to autoconf substitutions
```
pphaneuf: Hey! Stop impersonating me! :wink:

Makefile.am
```diff
@@ -18,7 +19,11 @@ BUILT_SOURCES = \
 BUILD_VERSION = `git describe --abbrev=0 --dirty --always --tags`
 
 AM_TESTS_ENVIRONMENT = \
-export LD_LIBRARY_PATH=$(LIBRARY_PATH);
+	LD_LIBRARY_PATH=$(INSTALL_DIR)/lib
```
pphaneuf: Alternatively, this could be set in the caller's environment (where `make check` is called), but I'd say that when we exterminate the dynamic libraries from `$(INSTALL_DIR)/lib`, we'll just ding all the references to `LD_LIBRARY_PATH`, and forget all about that particular bad dream...

Makefile.am
```diff
@@ -48,6 +53,7 @@ noinst_LIBRARIES = \
 	cpp/libtest.a
 
 check_PROGRAMS = \
+	cpp/util/thread_pool_test \
```
pphaneuf: So it's just wholly disabled for now? I'd say adding it conditionally to `TESTS` should be the next priority?

DEPS
pphaneuf: Yes, more can be done, like rewrite it in Go. :wink:

pphaneuf:
LGTM, although I would *particularly* like to see Travis green first. :smiley:

DEPS
AlCutter: :+1: Now we're talking!

.travis.yml
AlCutter: I think I might've mentioned that the other day, but yes that's definitely on the list!

PR #965, opened 2015-10-09
Python: Include Google logs' IDs in generated code

eranmes:
Chromium requires this information to be able to enforce a policy where
SCTs from Google-operated an non-Google operated logs are required.


eranmes:
Fixed.

PR #966, opened 2015-10-09
Fix another sign comparison case ...

Martin2112:
... and control reaching end of non void function warning.

### DIFF
--- a/cpp/server/ct-mirror.cc
+++ b/cpp/server/ct-mirror.cc
@@ -268,13 +268,19 @@ void STHUpdater(
         {
           LoggedCertificate entry;
           CHECK_LE(next_sth.tree_size(), local_size);
-          while (new_tree->LeafCount() < next_sth.tree_size()) {
+          CHECK_GE(next_sth.tree_size(), 0);
+          const uint64_t next_sth_tree_size(
+              static_cast<uint64_t>(next_sth.tree_size()));
+          while (new_tree->LeafCount() < next_sth_tree_size) {
             CHECK(entries->GetNextEntry(&entry));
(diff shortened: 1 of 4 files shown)


cpp/server/ct-mirror.cc
```diff
@@ -268,13 +268,19 @@ void STHUpdater(
         {
           LoggedCertificate entry;
           CHECK_LE(next_sth.tree_size(), local_size);
-          while (new_tree->LeafCount() < next_sth.tree_size()) {
+          CHECK_LE(next_sth.tree_size(), INT64_MAX);
```
AlCutter: I guess this should be `CHECK_GE(..., 0)`?

cpp/server/ct-mirror.cc
```diff
@@ -268,13 +268,19 @@ void STHUpdater(
...
+          const uint64_t next_sth_tree_size(
+              static_cast<uint64_t>(next_sth.tree_size()));
+          while (new_tree->LeafCount() < next_sth_tree_size) {
             CHECK(entries->GetNextEntry(&entry));
             CHECK(entry.has_sequence_number());
-            CHECK_EQ(new_tree->LeafCount(), entry.sequence_number());
+            CHECK_LE(entry.sequence_number(), INT64_MAX);
```
AlCutter: >=0 ?

cpp/server/ct-mirror.cc
Martin2112: Done.

cpp/server/ct-mirror.cc
Martin2112: Done.

PR #968, opened 2015-10-12
Fixes for signed / unsigned comparisons in tests

Martin2112:


cpp/log/tree_signer_test.cc
```diff
@@ -319,7 +319,9 @@ TYPED_TEST(TreeSignerTest, SequenceNewEntriesCleansUpOldSequenceMappings) {
   {
     EntryHandle<SequenceMapping> mapping;
     CHECK_EQ(Status::OK, this->store_->GetSequenceMapping(&mapping));
-    EXPECT_EQ(new_logged_certs.size(), mapping.Entry().mapping_size());
+    CHECK_LE(mapping.Entry().mapping_size(), INT_MAX);
```
AlCutter: Since this is protecting a signed->unsigned cast should it be `>= 0` ?

Martin2112: Yes, done.

PR #972, opened 2015-10-13
Fixes to sign fixes that weren't copied back to main branch.

Martin2112:


cpp/log/log_lookup-inl.h
```diff
@@ -66,7 +66,7 @@ void LogLookup<Logged>::UpdateFromSTH(const ct::SignedTreeHead& sth) {
   auto it(db_->ScanEntries(cert_tree_.LeafCount()));
   // LeafCount() is potentially unsigned here but as this is using memory
   // the count can never get close to overflow in 64 bits.
-  CHECK_LE(cert_tree_.LeafCount(), INT64_MAX);
+  CHECK_LE(cert_tree_.LeafCount(), (uint64_t) INT64_MAX);
```
pphaneuf: Could you switch it to use C++-style casts?

Martin2112: Done.

Martin2112:
Should be fixed now.

cpp/server/event.cc
```diff
@@ -183,7 +183,7 @@ bool EventLoop::EraseCheck(std::deque<FD*>::iterator* pfd) {
 // static
 void EventLoop::Set(int fd, fd_set* fdset, int* max) {
   DCHECK_GE(fd, 0);
-  CHECK_LT((unsigned)fd, FD_SETSIZE);
+  CHECK_LT((unsigned)fd, (unsigned)FD_SETSIZE);
```
pphaneuf: `static_cast` all the things! :smiley:

cpp/util/etcd.cc
```diff
@@ -980,7 +980,8 @@ list<EtcdClient::HostPortPair> SplitHosts(const string& hosts_string) {
   list<EtcdClient::HostPortPair> ret;
   for (const auto& h : hosts) {
     vector<string> hp(util::split(h, ':'));
-    CHECK_EQ(2, hp.size()) << "Invalid host:port string: '" << h << "'";
+    CHECK_EQ(static_cast<size_t>(2), hp.size()) << "Invalid host:port string: '"
```
pphaneuf: That's really necessary to shut it up?!? How about something like `2UL`, maybe that would work?

cpp/util/masterelection.cc
```diff
@@ -419,7 +419,7 @@ void MasterElection::UpdateProposalView(
     } else {
       VLOG(1) << my_proposal_path_
               << ": Proposal deleted: " << update.ToString();
-      CHECK_EQ(1, proposals_.erase(update.key_))
+      CHECK_EQ(static_cast<size_t>(1), proposals_.erase(update.key_))
```
pphaneuf: Same here, if `1UL` works, but if not, LGTM.

cpp/util/masterelection.cc
```diff
@@ -448,7 +448,7 @@ bool MasterElection::DetermineApparentMaster(
 void MasterElection::OnProposalUpdate(
     const vector<EtcdClient::Node>& updates) {
   unique_lock<mutex> lock(mutex_);
-  CHECK_GE(updates.size(), 1);
+  CHECK_GE(updates.size(), static_cast<size_t>(1));
```
pphaneuf: And again, if `1UL` works, but if not, LGTM.

pphaneuf:
LGTM, once you fix up `cpp/server/event.cc`. Feel free to merge after, it's fine. :smiley:

cpp/server/event.cc
Martin2112: I don't think FD_SETSIZE is a fixed type and this file uses non c++ casts elsewhere.

cpp/util/etcd.cc
Martin2112: I didn't want to assume the length of a size_t so I've made everywhere cast to the same type.

pphaneuf: Since this is for a constant, it should be fine (the compiler will do some promotion to make it work), it's the signedness that's a problem (although I'll say that's some crappy static analysis that gives you a warning for comparing a signed with an unsigned, when the "signed" one is a positive constant!).

cpp/server/event.cc
pphaneuf: Just `unsigned` is really `unsigned int` (because `int` is implicit in C! woo!), so that it's not a fixed type is (mostly) okay.

Now that I think of it, I think that if *all* the casts would be dropped from this line, it might be better. :stuck_out_tongue:

pphaneuf:
It's some nit-picking, really, if you think that's good, merge away.

cpp/util/etcd.cc
Martin2112: I know it will get promoted but it will potentially happen unnecessarily. Clock cycles aren't free you know!

pphaneuf: I think a `size_t` is a `ULL`, so you know... :smiley:

cpp/server/event.cc
Martin2112: There's no doc I've seen that says FD_SETSIZE must be unsigned. If it could be signed on some platforms (which makes no real sense) then it won't compile without the casts.

cpp/util/etcd.cc
Martin2112: Is that guaranteed by the standard?

pphaneuf: It's guaranteed to be "pretty big"... Big enough to contain a pointer type, something like that? But then, the `size()` method might already return that.

I wouldn't worry about it either way.

Issue #974, opened 2015-10-13
ElectionTest leaks memory.


AlCutter:

jmhodges:
(This can be closed now)

PR #975, opened 2015-10-13
Do some ASAN in travis too.

AlCutter:
Plus fixes for a couple of obvious leaks found by ASAN (more involved fixes swept under the carpet for the moment, tracked in #974.)


pphaneuf:
LGTM, even though a lot of `unique_ptr` where values would do, and assigning me TODOs again! :wink:

PR #976, opened 2015-10-14
Updates to docs for gclient builds on more platforms

Martin2112:
Update to build with GClient on F22 and CentOS 7 including potential flag changes to get around having -Werror set. Switch git checkouts in docs to https:// for more compatibility.

### DIFF
--- a/README.Fedora
+++ b/README.Fedora
@@ -1,5 +1,13 @@
 ## Quickstart on Fedora 22 ##
 
+Please attempt to use the GClient build as documented in the 
+[main readme](README.md) as this is an easier process and will be
+maintained in future.
+
+If GClient works and tests pass then the following procedures are not
+required.
+
+## Deprecated Manual Build Process ##
 
 Note: This assumes a Workstation install for x64. The additional dependency
 packages that need to be installed may vary if you are starting with a
@@ -8,70 +16,70 @@ different base system.
 
 Install Dependencies:
 
-
-sudo dnf update
+```sudo dnf update
 sudo dnf install cmake gcc-g++ libevent-devel golang autoconf pkgconfig \
     json-c-devel gflags-devel glog-devel protobuf-devel leveldb-devel \
     openssl-devel gperftools-devel protobuf-compiler sqlite-devel ant \
     java-1.8.0-openjdk-devel protobuf-java python-gflags protobuf-python \
-    python-ecdsa python-mock python-httplib2 git ldns-devel automake
-
+    python-ecdsa python-mock python-httplib2 git ldns-devel automake \
+    libtool shtool libunwind-devel
+```
 
 Other Libraries
 
 
-The gflags in Fedora is v2.1 and is using the new default namespace option of
+The `gflags` in Fedora is v2.1 and is using the new default namespace option of
 ‘gflags’ rather than ‘google’ so we need to build our own version. 
 
 
-    git clone https://github.com/gflags/gflags.git
+```git clone https://github.com/gflags/gflags.git
     cd gflags
     cmake -DGFLAGS_NAMESPACE:STRING=google \
         -DCMAKE_CXX_FLAGS:STRING=-fPIC .
     make
     cd ..
-
+```
 
 Next, we need `libevhtp` version `1.2.10` which is not packaged in Fedora, so
 we build from source:
 
-   wget https://github.com/ellzey/libevhtp/archive/1.2.10.zip
+```wget https://github.com/ellzey/libevhtp/archive/1.2.10.zip
    unzip 1.2.10.zip
    cd libevhtp-1.2.10/
    cmake -DEVHTP_DISABLE_REGEX:STRING=ON -DCMAKE_C_FLAGS:STRING=-fPIC .
    make
    cd ..
- 
+```
 
 And let's get our own Google Test / Google Mock as these vary in incompatible
 ways between packaged releases:
 
-   wget https://googlemock.googlecode.com/files/gmock-1.7.0.zip
+```wget https://googlemock.googlecode.com/files/gmock-1.7.0.zip
    unzip gmock-1.7.0.zip
-
+```
 Now, clone the CT repo:
 
-   git clone https://github.com/google/certificate-transparency.git
+```git clone https://github.com/google/certificate-transparency.git
    cd certificate-transparency/
-
+```
 
 One-time setup for Go:
 
-   export GOPATH=$PWD/go
+```export GOPATH=$PWD/go
    mkdir -p $GOPATH/src/github.com/google
    ln -s $PWD $GOPATH/src/github.com/google
    go get -v -d ./...
-
+```
 
 Build CT server C++ code:
 
-   ./autogen.sh
+```./autogen.sh
    ./configure GTEST_DIR=../gmock-1.7.0/gtest GMOCK_DIR=../gmock-1.7.0 \
        CPPFLAGS="-I../libevhtp-1.2.10 -I../libevhtp-1.2.10/evthr \
        -I../libevhtp-1.2.10/htparse -I../gflags/include" \
         LDFLAGS=”-L../libevhtp-1.2.10 -L../gflags/lib”
    make check 
-
+```
 
 The remainder of the Java, Go and Python steps should be very similar to those
-documented for Ubuntu in the main README.
+documented for Ubuntu in the [main readme file](README.md).
--- a/README.MacOS
+++ b/README.MacOS
@@ -1,4 +1,9 @@
-# Trusted root certificates
+## OSX Builds Now Use GClient ##
+
+We recommend that you use GClient to build on OSX. Please follow the 
+instructions in the [main readme](README.md) file.
+
+## Trusted root certificates ##
 
 The CT code requires a set of trusted root certificates in order to:
    1. Validate outbound HTTPS connections
@@ -45,3 +50,8 @@ security find-certificates -a -p /Library/Keychains/System.keychain > certs.pem
 security find-certificates -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> certs.pem
 ```
 
+## Deprecated Build Process ##
+
+This may be out of date and is not guaranteed to work.
+
+gtest: install from source.
--- a/README.md
+++ b/README.md
@@ -1,12 +1,88 @@
 certificate-transparency
 ========================
 
+#Auditing for TLS certificates#
+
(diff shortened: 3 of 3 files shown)


README.CentOS
```diff
@@ -0,0 +1,43 @@
...
+
+Install Dependencies:
+
+sudo yum update
+sudo yum install epel-release
+sudo yum install cmake gcc-c++ libevent-devel golang autoconf pkgconfig \
+    json-c-devel gflags-devel glog-devel protobuf-devel leveldb-devel \
+    openssl-devel gperftools-devel protobuf-compiler sqlite-devel ant \
+    java-1.8.0-openjdk-devel protobuf-java python-gflags protobuf-python \
+    python-ecdsa python-mock python-httplib2 git ldns-devel automake \
```
AlCutter: You shouldn't need the following since they're part of the gclient build:
* libevent-devel 
* json-c-devel 
* gflags-devel
* glog-devel
* protobuf-devel
* leveldb-devel
* openssl-devel
* gperftools-devel
* protobuf-compiler
* sqlite-devel
* ldns-devel
* libunwind-devel
```

File: README.CentOS [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+
+
+At this point you should be able to follow the GClient build procedure as 
+listed in README.md. This is recommended as it avoids the need to manually build
+dependencies.
+
+When you issue the GClient sync command you may need to set compiler options
+in order to build successfully. If the build fails to work try using:
+
+CXXFLAGS="-O2 -Wno-error=unused-variable" gclient sync -R
```
AlCutter: You *may* not need the `-R` any more

README.CentOS
Martin2112: Yep but installing them allows the old style build to be used. Do you think it's worth pulling them out separately? Or even combining the docs into one?

README.CentOS
AlCutter: Could pop these command lines into preformatted block:
```
```bash
CXXFLAGS=... blah -R
``
(actually need 3 backticks, but I can't seem to escape it here)
```

README.CentOS
```diff
@@ -0,0 +1,43 @@
...
+
+If this gives an error about an unused typedef in a glog header file try this:
+
+CXXFLAGS="-O2 -Wno-error=unused-variable -Wno-error=unused-local-typedefs" \
```
AlCutter: this block too

README.CentOS
```diff
@@ -0,0 +1,43 @@
...
+    json-c-devel gflags-devel glog-devel protobuf-devel leveldb-devel \
+    openssl-devel gperftools-devel protobuf-compiler sqlite-devel ant \
+    java-1.8.0-openjdk-devel protobuf-java python-gflags protobuf-python \
+    python-ecdsa python-mock python-httplib2 git ldns-devel automake \
+    libtool shtool libunwind-devel
+
+
+
+At this point you should be able to follow the GClient build procedure as 
+listed in README.md. This is recommended as it avoids the need to manually build
```
AlCutter: `[README.md](../README.md)` or similar?

README.CentOS
```diff
@@ -0,0 +1,43 @@
...
+in order to build successfully. If the build fails to work try using:
+
+CXXFLAGS="-O2 -Wno-error=unused-variable" gclient sync -R
+
+If this gives an error about an unused typedef in a glog header file try this:
+
+CXXFLAGS="-O2 -Wno-error=unused-variable -Wno-error=unused-local-typedefs" \
+gclient sync -R
+
+When changing CXXFLAGS it's safer to remove the existing build directories
```
AlCutter: I'd be tempted to put these `CXXFLAGS` words in single backticks.

README.Fedora
```diff
@@ -14,9 +14,35 @@ sudo dnf install cmake gcc-g++ libevent-devel golang autoconf pkgconfig \
     json-c-devel gflags-devel glog-devel protobuf-devel leveldb-devel \
```
AlCutter: Same comment about the extra unused deps here

README.Fedora
```diff
@@ -14,9 +14,35 @@ sudo dnf install cmake gcc-g++ libevent-devel golang autoconf pkgconfig \
...
     openssl-devel gperftools-devel protobuf-compiler sqlite-devel ant \
     java-1.8.0-openjdk-devel protobuf-java python-gflags protobuf-python \
-    python-ecdsa python-mock python-httplib2 git ldns-devel automake
+    python-ecdsa python-mock python-httplib2 git ldns-devel automake \
+    libtool shtool libunwind-devel
 
 
+At this point you can follow the GClient build procedure as listed in
+README.md. This is recommended as it avoids the need to manually build
```
AlCutter: linky

README.Fedora
```diff
@@ -14,9 +14,35 @@ sudo dnf install cmake gcc-g++ libevent-devel golang autoconf pkgconfig \
...
+dependencies.
+
+When you issue the GClient sync command you may need to set compiler options
+in order to build successfully. If the build fails to work try using:
+
+CXXFLAGS="-O2 -Wno-error=unused-variable" gclient sync -R
```
AlCutter:
```bash block

File: README.Fedora [hunk 9]
```diff
@@ -14,9 +14,35 @@ sudo dnf install cmake gcc-g++ libevent-devel golang autoconf pkgconfig \
...
+
+If this gives an error about an unused typedef in a glog header file try this:
+
+CXXFLAGS="-O2 -Wno-error=unused-variable -Wno-error=unused-local-typedefs" \
```
AlCutter:
```bash block

File: README.Fedora [hunk 10]
```diff
@@ -14,9 +14,35 @@ sudo dnf install cmake gcc-g++ libevent-devel golang autoconf pkgconfig \
...
+gclient sync -R
+
+When changing CXXFLAGS it's safer to remove the existing build directories
```
AlCutter: backtick variable names maybe?

README.MacOS
```diff
@@ -1,3 +1,10 @@
+We recommend that you use GClient to build on OSX. Please follow the 
+instructions in the README.md file.
```
AlCutter: linky

README.md
```diff
@@ -85,12 +85,19 @@ This is the recommended method for all platforms as it gives you a reproducible
 build.
 
 Known to work on FreeBSD 10, OS X (10.10) [tested with XCode + brew installation
-of above deps], and Ubuntu 14.04.
+of above deps], and Ubuntu 14.04. Tested on Fedora 22 but may require manual
+override of compiler options as documented in README.Fedora. Tested on
```
AlCutter: links for the 2 docs

README.CentOS
AlCutter: Ah I see, I think that we should deprecate the old-style build in favour of the GClient one; that way we can rely on there being a known version of OpenSSL etc. rather than relying on the system provided version. WDYT?

Martin2112: Yes, that's probably best. I'm starting to think we should document the GClient required dependencies for each platform and any flag quirks in the main file as the process is basically identical except that.

Martin2112:
I'm going to reorganize the docs so that README.md has the GClient build first followed by any platform specific notes on flags followed by the old build process for Ubuntu. Then point at it from the other platform files, which we can clean up in future, and list other build methods as deprecated.

README.Fedora
Martin2112: This doc now only applies if you're not using recommended gclient build so I think it's now OK.

Martin2112:
Is this better now it's been moved around?

README.MacOS
```diff
@@ -1,3 +1,12 @@
+## OSX Builds Now Use GClient ##
```
AlCutter: I think you might need to merge :)

README.md
```diff
@@ -1,12 +1,97 @@
...
+ - git
+ - GNU make
+ - libtool
+ - shtool
+ - Tcl
+ - pkgconf
+ - python27
+ - [depot_tools](https://www.chromium.org/developers/how-tos/install-depot-tools)
+
+Note: On Fedora, CentOS and possibly others the libunwind-devel package must
```
AlCutter: Does `libunwind` not get built on those?

Might be worth looking at `sys.platform` in `DEPS`, if it's not `linux2` for some reason we might need to duplicate the `deps_overrides` for `linux2` into whatever  the platform string is for those platforms.

README.md
```diff
@@ -1,12 +1,97 @@
...
+of deps listed below], and Ubuntu 14.04. Tested on Fedora 22 but may require
+manual override of compiler options as documented below. Tested on CentOS 7
+with similar caveats.
+
+### Install Dependencies ###
+
+Depending on which platform you have the exact packages required will vary.
+The following tools must be available for the GClient build to succeed:
+
+ - A working C++11 compiler.
```
AlCutter: I have a feeling this is currently more like "use clang" :)

README.md
```diff
@@ -1,12 +1,97 @@
...
+also be installed in addition to the above tools.
+
+### Building with gclient ###
+
+```bash
+mkdir ct  # or whatever directory you prefer
+cd ct
+gclient config --name="certificate-transparency" https://github.com/google/certificate-transparency.git
+gclient sync
+# substitute gmake or gnumake below if that's that your platform calls it:
```
AlCutter: /that's that/that's what/

README.md
```diff
@@ -1,12 +1,97 @@
...
+
+If you're trying to clone from a branch on the CT repo then you'll need to
+substitute the following command for the `gclient config` command above,
+replacing `branch` as appropriate
+
+```bash
+gclient config --name="certificate-transparency" https://github.com/google/certificate-transparency.git@branch
+```
+
+To run any of the built binaries you'll need to:
```
AlCutter: We don't need this LD_LIBRARY_PATH bit anymore since we're semi-static now.

README.MacOS
Martin2112: Conflicts fixed up.

README.md
Martin2112: It didn't build without this when I first tried it. It's possible the later changes you made to deps fixed it. Might be worth retesting it.

README.md
Martin2112: Done.

README.md
Martin2112: Removed the section.

README.md
Martin2112: F22 is linux2. It does now seem to build without having that installed. I'll test on CentOS tomorrow.

Martin2112: I just tested on CentoOS and it built successfully. Removed the note as obsolete.

PR #977, opened 2015-10-14
Improves Python certificate scanning resiliency

RJPercival:
CertificateReport will no longer stop scanning all of the certificates in a batch and discard the results if one certificate results in an unexpected exception being raised. Instead, it will log the exception and continue scanning the rest of the certificates. The reporters have also been made more resilient to exceptions - their writer threads are less likely to die due to unexpected exceptions and may now be restarted.


phad:
This looks much better, thanks.

PR #980, opened 2015-10-16
Python: Small tidying-up of the reporter, client code.

eranmes:
No functional changes.

### DIFF
--- a/python/ct/client/db_reporter.py
+++ b/python/ct/client/db_reporter.py
@@ -19,23 +19,25 @@ class CertDBCertificateReport(reporter.CertificateReport):
         self._writer = None
         super(CertDBCertificateReport, self).__init__(checks=checks)
 
+    def _writer_ready(self):
+        return self._writer and self._writer.is_alive()
+
     def report(self):
         super(CertDBCertificateReport, self).report()
 
-        if self._writer and self._writer.is_alive():
+        if self._writer_ready():
             self._certs_queue.join()
             logging.info("Finished scheduled writing to CertDB")
             self._certs_queue.put(None)
             self.reset()
 
     def reset(self):
-        if self._writer and self._writer.is_alive():
+        if self._writer_ready():
             self._writer.join()
             self._writer = None
 
-
     def _batch_scanned_callback(self, result):
-        if not self._writer or not self._writer.is_alive():
+        if not self._writer_ready():
             self._writer = threading.Thread(target=_process_certs,
                                             args=(self._cert_db, self.log_key,
                                                   self._certs_queue))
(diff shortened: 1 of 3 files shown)


eranmes:
Also @phad

eranmes:
Re the changes in _process_certs, not sure my code ended up being more readable , if you don't think it is, I'll revert and just leave the comment saying the check for certs being None must be inside the try clause because the statement in the finally clause must be called for that case as well.

python/ct/client/reporter.py
```diff
@@ -141,15 +145,15 @@ def scan_der_certs(self, der_certs):
...
-            if scan_results is None:
-                break
-
             report._batch_scanned_callback(scan_results.get())
         except:
             logging.exception("Error occurred during certificate scanning")
         finally:
             queue.task_done()
+        scan_results = queue.get()
+    # Finall task_done call for when certs is None and the body of the
```
phad: Typo: Finall -> Final

phad: C&P:  "when certs is None" :  certs -> scan_results

python/ct/client/db_reporter.py
```diff
@@ -44,15 +46,15 @@ def _batch_scanned_callback(self, result):
...
-            if certs is None:
-                break
-
             db.store_certs_desc(certs, log_key)
         except:
             logging.exception("Failed to store certificate information")
         finally:
             certs_queue.task_done()
+        certs = certs_queue.get()
+    # Finall task_done call for when certs is None and the body of the
```
phad: Typo: Finall->Final

python/ct/client/db_reporter.py
```diff
@@ -44,15 +46,15 @@ def _batch_scanned_callback(self, result):
 
 
 def _process_certs(db, log_key, certs_queue):
-    while True:
-        certs = certs_queue.get()
-
+    certs = certs_queue.get()
```
phad: Personally: I like Rob's approach to the loop better.
He's writing a "forever" loop, with a break, and that's indicated with while(True)
In your version you end up with a pre-loop 'get' call which is then repeated in the loop body so it can decide whether to continue or not.  I prefer the 'forever' construct over the repeated codel

python/ct/client/db_reporter.py
```diff
@@ -19,23 +19,25 @@ def __init__(self, cert_db, log_key, checks=all_checks.ALL_CHECKS):
         self._writer = None
         super(CertDBCertificateReport, self).__init__(checks=checks)
 
+    def _writer_ready(self):
```
phad: This function extraction definitely helps the readability of the code below particular in the negated case.

python/ct/client/db_reporter.py
eranmes: Switched back.

eranmes:
Pushed a new version, @phad  PTAL.

PR #982, opened 2015-10-19
Break out the CMS related code into a new class.

Martin2112:
Break out the CMS related code (v2 related) into a new class and move tests out of cert_checker_test.


Martin2112:
This is almost identical to code submitted last week. Just pulls it out and tweaks the API.

cpp/log/cert_verifier.h
```diff
@@ -0,0 +1,58 @@
...
+#include <openssl/cms.h>
+
+#include "base/macros.h"
+#include "log/cert.h"
+#include "util/openssl_util.h"  // for LOG_OPENSSL_ERRORS
+#include "util/status.h"
+
+namespace cert_trans {
+
+class CertVerifier {
```
AlCutter: Having `CertChecker` and `CertVerifier` could get a bit confusing, how about `CMSVerifier`, (although maybe that's not so good as this doesn't [currently] verify the signature)?

Martin2112:
Naming things is hard! I would maybe go with CMSVerifier. It does verify the CMS signature of the message, but not that the signer cert is part of a valid chain as we use in CT.

cpp/log/cert.h
```diff
@@ -182,6 +182,8 @@ class Cert {
 
   // CertChecker needs access to the x509_ structure directly.
   friend class CertChecker;
+  // CertVerifier needs access to the x509_ structure directly.
```
AlCutter: `CmsVerifier needs ...` :)

cpp/log/cms_verifier.cc
```diff
@@ -0,0 +1,125 @@
+/* -*- indent-tabs-mode: nil -*- */
+#include <log/cms_verifier.h>
```
AlCutter: `#include "log/cms_verifier.h"`

cpp/log/cert.h
Martin2112: So Eclipse refactor didn't work? The world is back to normal now. Will fix it.

cpp/log/cms_verifier_test.cc
```diff
@@ -0,0 +1,211 @@
+/* -*- indent-tabs-mode: nil -*- */
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+#include <gtest/gtest.h>
+#include <log/cms_verifier.h>
```
AlCutter: `"log/cms_verifier.h"` and move down to block below

cpp/log/cms_verifier_test.cc
```diff
@@ -0,0 +1,211 @@
...
+// issuer of the embedded cert
+static const char kCmsSignedDataTest5[] = "cms_test5.der";
+
+// Subject name we expect in our embedded certificate CMS tests
+static const char kCmsTestSubject[] = "CN=?.example.com, C=GB, ST=Wales, "
+    "L=Erw Wen, O=Certificate Transparency";
+
+namespace {
+
+class CertVerifierTest : public ::testing::Test {
```
AlCutter: CmsVerifierTest

cpp/log/cert.h
Martin2112: Done.

cpp/log/cms_verifier.cc
Martin2112: Done.

cpp/log/cms_verifier_test.cc
Martin2112: Done.

cpp/log/cms_verifier_test.cc
Martin2112: Done + other occurrences.

Martin2112:
I think that's all fixed up now.

cpp/log/cms_verifier_test.cc
```diff
@@ -0,0 +1,211 @@
...
+  void SetUp() {
+    CHECK(util::ReadTextFile(cert_dir_ + "/" + kLeafCert, &leaf_pem_))
+        << "Could not read test data from " << cert_dir_
+        << ". Wrong --test_srcdir?";
+    CHECK(util::ReadTextFile(cert_dir_ + "/" + kIntermediateCert,
+                             &intermediate_pem_));
+    CHECK(util::ReadTextFile(cert_dir_ + "/" + kCaCert, &ca_pem_));
+  }
+};
+
```
AlCutter: Last thing: could you pop two lines between top-level items here & elsewhere?

Martin2112: Done. But I'm not sure it's a great idea as it can't be done automatically by formatters.

Issue #979, opened 2015-10-15
#947 doesn't work as intended


AlCutter:
You broke it :)

The "use the default system roots" mode (i.e. the new default) in your PR #947 doesn't work on Linux, at least not on my workstation and not on our GCE instances.

```
I1015 16:52:22.842362 31740 connection_pool.cc:430] new evhtp_connection for ct.googleapis.com:443
I1015 16:52:22.842733 31740 libevent_wrapper.cc:302] Got addr: 216.58.210.10:443
I1015 16:52:22.842994 31735 url_fetcher.cc:157] evhtp_make_request(0x17dcdc0, 0x17e9180, 0, "/testtube/ct/v1/get-sth")
W1015 16:52:22.849531 31735 connection_pool.cc:169] OpenSSL failed to verify cert for /C=US/O=GeoTrust Inc./CN=GeoTrust Global CA: unable to get local issuer certificate
```

aeijdenberg:
Why didn't your tests pick it up? ;)   I'll look into.

aeijdenberg:
In all seriousness, can you give me a set of commands that demonstrate failure?

AlCutter:
Tests didn't pick it up because:
1) the existing tests use the `trusted_root_certs` flag to set the trusted roots to the test roots explicitly, and 
2) you didn't add any tests for your new feature when you changed the code :)

To reproduce:
* Build `cpp/server/ct-mirror`
* cd test
* edit `run_mirror_server.sh` and remove the `--etcd_servers...` line (so that it runs in "standalone" mode with a fake in-memory etcd)
* run: `./run_mirror_server.sh /tmp/test.ldb`
* look for the `OpenSSL failed to verify cert for /C=US/O=GeoTrust Inc./CN=GeoTrust Global CA: unable to get local issuer certificate` errors.

aeijdenberg:
Our working theory is that this issue occurs when using the gclient build which has a custom openssl install.  When using the system libraries, everything is fine.  The default location for certs is inside of OPENSSLDIR which is set a compile time for openssl.

Since we are about to recommend gclient as "the right way" this bears thinking about.

AlCutter:
[...]

PR #983, opened 2015-10-19
Try not to clobber OpenSSL's cert lookup paths.

AlCutter:
This seems to use the system CA certs on Ubuntu and FreeBSD now.

Addresses #979

### DIFF
--- a/build/Makefile.openssl
+++ b/build/Makefile.openssl
@@ -1,3 +1,7 @@
+# Grab the openssldir of the system version of openssl in order to try to have
+# our version re-use the system caroots by default.
+OPENSSLDIR := $(shell openssl version -d | awk '{ print $$2 }')
+
 all: $(INSTALL_DIR)/bin/openssl
 
 $(INSTALL_DIR)/bin/openssl: Makefile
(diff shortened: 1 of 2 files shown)


build/Makefile.openssl
```diff
@@ -7,6 +7,6 @@ $(INSTALL_DIR)/bin/openssl: Makefile
 
 Makefile: config Configure Makefile.org
 	# Force a 64 bit build on MacOS
-	KERNEL_BITS=64 ./config --openssldir=$(INSTALL_DIR) no-shared enable-static-engine -fPIC
+	KERNEL_BITS=64 ./config --install_prefix=$(INSTALL_DIR) --prefix=/ --libdir=lib --openssldir=/etc/ssl no-shared enable-static-engine -fPIC
```
Martin2112: Presumably forcing /etc/ssl mean the certs won't be found on OSX and we're OK with that?

Martin2112: Sorry, in case that wasn't clear. I meant we're OK with people having to specify the path if they want to run a prod log?

AlCutter: Turns out the OSX system version of OpenSSL (0.9.8zg apparently) has some magic Apple pixie dust which allows it to reach into the keychain to retry failed verification attempts with whatever roots it has in there.

Our build of vanilla OpenSSL 1.0.2 can't do that so people are going to have to do it the 'normal' OpenSSL way on OS X too.
Where 'normal' is one of:
   * use the flag, Luke
   * set SSL_CERT_FILE
   * dump some certs in /System/Library/OpenSSL/certs

I added some stuff in the `README.MacOS` doc about it.

AlCutter:
PTAL

Martin2112:
Doc update looks good.

PR #987, opened 2015-10-21
Python: Verify correct signature encoding for ECDSA signatures.

eranmes:
We have witnessed an SCT with a signature that is not encoded as valid DER.
The Python ecdsa module silently accepts it but OpenSSL, for example,
does not.

Change the Python code to identify and reject such signatures.


pphaneuf:
LGTM, pending Travis.

PR #989, opened 2015-10-21
Rename LoggedCertificate to LoggedEntry

AlCutter:
Just a mechanical rename of `LoggedCertificate` to `LoggedEntry` to prepare the way for logging other things.

### DIFF
--- a/cpp/fetcher/remote_peer.cc
+++ b/cpp/fetcher/remote_peer.cc
@@ -8,7 +8,6 @@
 #include "monitoring/monitoring.h"
 #include "util/util.h"
 
-using cert_trans::LoggedCertificate;
 using ct::SignedTreeHead;
 using std::bind;
 using std::chrono::seconds;
--- a/cpp/fetcher/remote_peer.h
+++ b/cpp/fetcher/remote_peer.h
@@ -7,8 +7,6 @@
 
 namespace cert_trans {
 
-class LoggedCertificate;
-
 
 class RemotePeer : public Peer {
  public:
--- a/cpp/log/etcd_consistent_store-inl.h
+++ b/cpp/log/etcd_consistent_store-inl.h
@@ -9,7 +9,6 @@
 
 #include "base/notification.h"
 #include "log/etcd_consistent_store.h"
-#include "log/logged_certificate.h"
 #include "monitoring/event_metric.h"
 #include "monitoring/latency.h"
 #include "monitoring/monitoring.h"
--- a/cpp/log/logged_entry.h
+++ b/cpp/log/logged_entry.h
@@ -1,7 +1,7 @@
 /* -*- mode: c++; indent-tabs-mode: nil -*- */
 
-#ifndef LOGGED_CERTIFICATE_H
-#define LOGGED_CERTIFICATE_H
+#ifndef LOGGED_ENTRY_H
+#define LOGGED_ENTRY_H
 
 #include <glog/logging.h>
 
@@ -13,10 +13,10 @@
 
 namespace cert_trans {
 
-class LoggedCertificate : public ct::LoggedCertificatePB {
+class LoggedEntry : public ct::LoggedEntryPB {
  public:
   std::string Hash() const {
-    return Sha256Hasher::Sha256Digest(Serializer::LeafCertificate(entry()));
+    return Sha256Hasher::Sha256Digest(Serializer::LeafData(entry()));
   }
 
   uint64_t timestamp() const {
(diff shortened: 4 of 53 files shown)


cpp/log/logged_entry.h
```diff
@@ -1,7 +1,7 @@
 /* -*- mode: c++; indent-tabs-mode: nil -*- */
 
-#ifndef LOGGED_CERTIFICATE_H
-#define LOGGED_CERTIFICATE_H
+#ifndef LOGGED_ENTRY_H
+#define LOGGED_ENTRY_H
```
pphaneuf: Hmm, there's no `// LOGGED_ENTRY_H` at the `#endif`? I'm missing some tool that would have told us this, I think. :smiley:

cpp/log/logged_entry.h
```diff
@@ -53,12 +53,18 @@ class LoggedCertificate : public ct::LoggedCertificatePB {
...
+        return Serializer::SerializeX509Chain(entry().x509_entry(), dst) ==
+               Serializer::OK;
+        break;
+      case ct::PRECERT_ENTRY:
+        return Serializer::SerializePrecertChainEntry(entry().precert_entry(),
+                                                      dst) == Serializer::OK;
+        break;
+      default:
+        LOG(FATAL) << "Unknown entry type: " << entry().type();
+    }
```
pphaneuf: Could you use the pattern where each case calls `return`, and not use a `default` clause, so that compiler give us warnings on unhandled `enum` values?

cpp/log/logged_entry_test.cc
```diff
@@ -0,0 +1,7 @@
+#include "log/logged_entry.h"
+
+#include <gtest/gtest.h>
+
+typedef testing::Types<cert_trans::LoggedEntry> TestType;
+
+#include "log/logged_test-inl.h"
```
pphaneuf: That thing is included just *once*, why, oh, why isn't it just *here*?

pphaneuf: Don't do this in that PR, though, it'll be messy. I'll do a quick one once you merged this...

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -9,7 +9,7 @@
 
 #include "base/notification.h"
 #include "log/etcd_consistent_store.h"
-#include "log/logged_certificate.h"
+#include "log/logged_entry.h"
```
pphaneuf: Includes it, but doesn't use it?

cpp/fetcher/remote_peer.h
```diff
@@ -7,7 +7,7 @@
 
 namespace cert_trans {
 
-class LoggedCertificate;
+class LoggedEntry;
```
pphaneuf: And doesn't use it? Might have done so in the past, then refactored it away...

cpp/fetcher/remote_peer.cc
```diff
@@ -8,7 +8,7 @@
 #include "monitoring/monitoring.h"
 #include "util/util.h"
 
-using cert_trans::LoggedCertificate;
+using cert_trans::LoggedEntry;
```
pphaneuf: Same here? Although I'm guessing rather surprised!

pphaneuf:
Review done, just a few niggles, should be easy peasy!

cpp/log/logged_entry.h
AlCutter: added

cpp/log/logged_entry.h
AlCutter: Yeah, good call - this was only necessary as I'd added (and later removed for the PR) another entry type.

cpp/log/etcd_consistent_store-inl.h
AlCutter: Removed

cpp/fetcher/remote_peer.h
AlCutter: Removed

cpp/fetcher/remote_peer.cc
AlCutter: Removed

pphaneuf:
LGTM

cpp/log/logged_entry.h
Martin2112: Can you check this works in both clang and gcc? I think you might get a warning in gcc, which becomes an error with the default options on some platforms.

PR #991, opened 2015-10-22
Python: Use flags only after initialization.

eranmes:
Using FLAGS to get a default value for a method's parameter evlauates
it during import time, before FLAGS was actually initialized.


pphaneuf:
LGTM

PR #992, opened 2015-10-22
Factor out cert handler

AlCutter:
This one is mostly just moving stuff around; pulling out the CT/certificate specific stuff from `handler.cc` and putting it in a subclass, then there's also the changes to support registering these handler subclasses with the `Server` rather than the server creating the handler instance(s).

I'm not super happy with the current state of the Server interface, but I think that's a job for another day.

### DIFF
--- a/cpp/log/cert_submission_handler.cc
+++ b/cpp/log/cert_submission_handler.cc
@@ -23,7 +23,7 @@ using util::StatusOr;
 
 // TODO(ekasper): handle Cert errors consistently and log some errors here
 // if they fail.
-CertSubmissionHandler::CertSubmissionHandler(CertChecker* cert_checker)
+CertSubmissionHandler::CertSubmissionHandler(const CertChecker* cert_checker)
     : cert_checker_(CHECK_NOTNULL(cert_checker)) {
 }
 
@@ -76,7 +76,8 @@ bool CertSubmissionHandler::X509ChainToEntry(const CertChain& chain,
 }
 
 Status CertSubmissionHandler::ProcessX509Submission(CertChain* chain,
-                                                    LogEntry* entry) {
+                                                    LogEntry* entry) const {
+  entry->set_type(ct::X509_ENTRY);
   if (!chain->IsLoaded())
     return Status(util::error::INVALID_ARGUMENT, "empty submission");
 
@@ -99,12 +100,12 @@ Status CertSubmissionHandler::ProcessX509Submission(CertChain* chain,
     }
     x509_entry->add_certificate_chain(der_cert);
   }
-  entry->set_type(ct::X509_ENTRY);
   return Status::OK;
 }
 
 Status CertSubmissionHandler::ProcessPreCertSubmission(PreCertChain* chain,
-                                                       LogEntry* entry) {
+                                                       LogEntry* entry) const {
+  entry->set_type(ct::PRECERT_ENTRY);
   PrecertChainEntry* precert_entry = entry->mutable_precert_entry();
   const Status status(cert_checker_->CheckPreCertChain(
       chain, precert_entry->mutable_pre_cert()->mutable_issuer_key_hash(),
@@ -125,7 +126,6 @@ Status CertSubmissionHandler::ProcessPreCertSubmission(PreCertChain* chain,
       return Status(util::error::INTERNAL, "could not DER-encode the chain");
     precert_entry->add_precertificate_chain(der_cert);
   }
-  entry->set_type(ct::PRECERT_ENTRY);
   return Status::OK;
 }
 
--- a/cpp/log/frontend_test.cc
+++ b/cpp/log/frontend_test.cc
@@ -101,8 +101,8 @@ class FrontendTest : public ::testing::Test {
         pool_(2),
         store_(base_.get(), &pool_, &etcd_client_, &election_, "/root", "id"),
         log_signer_(TestSigner::DefaultLogSigner()),
-        frontend_(new CertSubmissionHandler(&checker_),
-                  new FrontendSigner(db(), &store_, log_signer_.get())),
+        submission_handler_(&checker_),
+        frontend_(new FrontendSigner(db(), &store_, log_signer_.get())),
         cert_dir_(FLAGS_test_srcdir + "/test/testdata") {
   }
 
@@ -141,6 +141,7 @@ class FrontendTest : public ::testing::Test {
   NiceMock<MockMasterElection> election_;
   EtcdConsistentStore<LoggedEntry> store_;
   unique_ptr<LogSigner> log_signer_;
+  CertSubmissionHandler submission_handler_;
   FE frontend_;
   const string cert_dir_;
   string leaf_pem_;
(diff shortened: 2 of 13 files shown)


Martin2112:
This looks related to my WIP in #878.

Martin

cpp/server/handler.cc
```diff
@@ -1,237 +1,78 @@
...
-using cert_trans::CertChain;
-using cert_trans::CertChecker;
-using cert_trans::Counter;
-using cert_trans::HttpHandler;
-using cert_trans::JsonOutput;
-using cert_trans::Latency;
-using cert_trans::LoggedEntry;
-using cert_trans::Proxy;
-using cert_trans::ScopedLatency;
+#include "server/handler.h"
```
pphaneuf: What the heck happened to this file? Its header should be first, and generally, this reshuffling is making it rather difficult to see what's going on with *this* PR...

pphaneuf:
Could you review #995, and rebase on top once I've merged it? :stuck_out_tongue:

cpp/server/handler.cc
AlCutter: Ah, this is because this file briefly turned into a template (remember?) and then back out to a regular class.
I've moved the class header back up top, though.

AlCutter:
Ping

cpp/server/handler.cc
```diff
@@ -3,160 +3,77 @@
...
-using cert_trans::Latency;
-using cert_trans::LoggedEntry;
-using cert_trans::Proxy;
-using cert_trans::ScopedLatency;
+#include "util/libevent_wrapper.h"
+#include "util/sync_task.h"
+#include "util/task.h"
+
+
+DEFINE_int32(max_leaf_entries_per_response, 1000,
```
pphaneuf: Did those change at all, or did they just migrate for the coming winter months? :wink: 

I was kind of hoping for "diff minimisation". :smiley:

cpp/server/handler.cc
```diff
@@ -3,160 +3,77 @@
...
+             "maximum number of entries to put in the response of a "
+             "get-entries request");
+DEFINE_int32(staleness_check_delay_secs, 5,
+             "number of seconds between node staleness checks");
+
+
+namespace cert_trans {
```
pphaneuf: I think if you moved lines 26-33 to line 47, it'd make more sense? Small diff *and* reasonable thing to do? (it might be reasonable now, but you're making my pretty little head hurt!)

AlCutter: moved

pphaneuf: All right, thanks! I can now see that many relevant things *did* happen within that noise! :smiley:

cpp/server/handler.cc
```diff
@@ -3,160 +3,77 @@
 #include <algorithm>
 #include <event2/buffer.h>
```
pphaneuf: I *think* this one might not be needed either, with `ExtractChain` having moved? Not 100% sure about `event2/http.h`, though.

AlCutter: Dinged those two (on a related note, we should run IWYU over the codebase rather than trying to do this manually.)

AlCutter:
Ping

cpp/log/cert_submission_handler.cc
```diff
@@ -76,7 +76,8 @@ bool CertSubmissionHandler::X509ChainToEntry(const CertChain& chain,
 }
 
 Status CertSubmissionHandler::ProcessX509Submission(CertChain* chain,
-                                                    LogEntry* entry) {
+                                                    LogEntry* entry) const {
+  entry->set_type(ct::X509_ENTRY);
```
pphaneuf: This here, was this a separate bug?!?

cpp/log/frontend_test.cc
```diff
@@ -162,8 +163,11 @@ TYPED_TEST(FrontendTest, TestSubmitValid) {
   CertChain chain(this->leaf_pem_);
   EXPECT_TRUE(chain.IsLoaded());
 
+  LogEntry entry;
   SignedCertificateTimestamp sct;
-  EXPECT_OK(this->frontend_.QueueX509Entry(&chain, &sct));
+  EXPECT_OK(this->frontend_.QueueProcessedEntry(
+      this->submission_handler_.ProcessX509Submission(&chain, &entry), entry,
```
pphaneuf: Passing the same `entry` twice, with one of them looking "by value" (yeah, I know, thank you C++!), that looked rather alarming.

To my eye, it really looks like `ProcessX509Submission` and `ProcessPreCertSubmission` want to return a `StatusOr<LogEntry>`...

AlCutter: Yep, that's a good idea and actually fits very well with what `QueueProcessedEntry` really does, I just kept things the same 'shape' as they were previously, but we can do this change in a separate PR.

cpp/server/handler.h
```diff
@@ -32,21 +33,27 @@ class ThreadPool;
...
 
   void Add(libevent::HttpServer* server);
 
- private:
+  void SetOutput(JsonOutput* json_output);
+  void SetProxy(Proxy* proxy);
+
+ protected:
+  // Implemented by subclasses which want to add their own extra http handlers.
+  virtual void AddHandlers(libevent::HttpServer* server) = 0;
```
pphaneuf: I'm still fuzzy on why it's not just `HttpHandler::Add` that shouldn't just become virtual?

cpp/server/certificate_handler.cc
```diff
@@ -0,0 +1,181 @@
...
+
+
+CertificateHttpHandler::CertificateHttpHandler(
+    LogLookup<LoggedEntry>* log_lookup,
+    const ReadOnlyDatabase<LoggedEntry>* db,
+    const ClusterStateController<LoggedEntry>* controller,
+    const CertChecker* cert_checker, Frontend* frontend, ThreadPool* pool,
+    libevent::Base* event_base)
+    : HttpHandler(log_lookup, db, controller, pool, event_base),
+      cert_checker_(cert_checker),
```
pphaneuf: Some `CHECK_NOTNULL`?

pphaneuf:
Review done! **I can see!**

cpp/server/handler.h
AlCutter: I could do, it's just that I figured that all subclasses will want to map the `get-entries`, `get-proof-by-hash`, `get-sth`, and `get-sth-consistency` handlers, so may as well just bake them in here.
If `Add` is virtual then the subclass has to either add them all itself, or include a call to the superclass `Add` implementation.  The former is a bunch of extra boilerplate for no real reason, and the latter would be ok but costs an extra line and could also lead to confusing "why are only *some* handlers missing" bugs.

cpp/server/certificate_handler.cc
AlCutter: `HttpHandler` checks all of its params for not-nullness, `frontend` is permitted to be null, but I've added a check for `cert_checker`

pphaneuf:
LGTM

PR #995, opened 2015-10-23
Split out the query parsing functions.

pphaneuf:
Putting them in `libevent_wrapper.{h,cc}`, because they're based on `evhttp_parse_query_str`.

Strongly inspired by @AlCutter himself, and done so I could manage to review #992. :wink:

### DIFF
--- a/cpp/util/libevent_wrapper.cc
+++ b/cpp/util/libevent_wrapper.cc
@@ -8,6 +8,7 @@
 #endif
 #include <climits>
 #include <evhtp.h>
+#include <event2/keyvalq_struct.h>
 #include <event2/thread.h>
 #include <glog/logging.h>
 #include <math.h>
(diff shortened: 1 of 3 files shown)


cpp/util/libevent_wrapper.cc
```diff
@@ -383,6 +386,79 @@ void HttpServer::HandleRequest(evhttp_request* req, void* userdata) {
...
+  }
+
+  return retval;
+}
+
+
+bool GetParam(const QueryParams& query, const string& param, string* value) {
+  CHECK_NOTNULL(value);
+
+  auto it = query.find(param);
```
pphaneuf: This here is the only actual change (it used to be `multimap<string, string>::const_iterator`), beside the `QueryParams` typedef. I lifted it straight out of #992, but it's now more equivalent to `QueryParams::iterator` than `QueryParams::const_iterator`. `const auto` wouldn't work, as it would prevent advancing the iterator.

I think it's all right like this, though, the rest of the function is short enough to be able to comprehensively review it.

PR #996, opened 2015-10-27
Updates to proto and serialization for CT V2

Martin2112:
I think this is mostly complete. A couple of things are still unclear e.g. whether to rename / update some protos for V2, which will cause churn elsewhere in the code but will match the draft RFC. Note at this level there are no checks other than that the data being read / written correspond to a supported version. This should be done at a higher level where necessary.

### DIFF
--- a/cpp/log/etcd_consistent_store-inl.h
+++ b/cpp/log/etcd_consistent_store-inl.h
@@ -355,15 +355,6 @@ util::Status EtcdConsistentStore<Logged>::GetPendingEntries(
 
 
 template <class Logged>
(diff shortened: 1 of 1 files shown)


Martin2112:
Sorry I missed out handling the precert chain serialization. No point reviewing it just yet.

Martin2112:
OK this is turning out to be more complex than hoped. I can't do the proto / serializer in isolation because adding constants to messages causes problems compiling the signer etc. I need to to more work on this.

Martin2112:
This was closed as a result of merging / fixing up conflicts with the xjson code. I'm going to send out another PR shortly.

PR #998, opened 2015-10-27
Certinfo refactor in advance of V2 serialization support

Martin2112:
Rename PreCert to CertInfo as in draft RFC as in v2 it will be used for both certs and precerts in signed entries. Unfortunately can't use CertInfo exclusively in SignedEntry proto as this would be incompatible with the format of existing saved proto instances.

### DIFF
--- a/build/Makefile.glog
+++ b/build/Makefile.glog
@@ -1,6 +1,6 @@
 UNAME := $(shell uname)
 ifeq ($(UNAME),FreeBSD)
-EXTRA_LDFLAGS="-lexecinfo"
+EXTRA_LDFLAGS="-lexecinfo -lunwind"
 endif
 
 all: Makefile
--- a/build/Makefile.glog
+++ b/build/Makefile.glog
@@ -1,6 +1,6 @@
 UNAME := $(shell uname)
 ifeq ($(UNAME),FreeBSD)
-EXTRA_LDFLAGS="-lexecinfo -lunwind"
+EXTRA_LDFLAGS="-lexecinfo"
 endif
 
 all: Makefile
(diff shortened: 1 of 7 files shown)


build/Makefile.glog
```diff
@@ -1,6 +1,6 @@
 UNAME := $(shell uname)
 ifeq ($(UNAME),FreeBSD)
-EXTRA_LDFLAGS="-lexecinfo"
+EXTRA_LDFLAGS="-lexecinfo -lunwind"
```
benlaurie: Not to do with this change?

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
```diff
@@ -176,19 +176,19 @@ public static ParsedLogEntry parseLogEntry(InputStream merkleTreeLeaf, InputStre
       signedEntryBuilder.setX509(x509);
 
     } else if (entryType == Ct.LogEntryType.PRECERT_ENTRY_VALUE) {
-      Ct.PreCert.Builder preCertBuilder = Ct.PreCert.newBuilder();
+      Ct.PreCert.Builder certInfoBuilder = Ct.CertInfo.newBuilder();
```
benlaurie: This is still building a precert, right?

build/Makefile.glog
Martin2112: Yes and no. This PR exposed an issue with the static build of glog. I tried this as a fix but as it didn't work I'm rolling it back.

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
Martin2112: Yeah it's still wrong. Sorry. I was trying to debug the Go test issue first.

Martin2112:
On reflection I think it's probably best to abandon this PR and make a fully backwards compatible proto change in the other PR. It's more work for me to do but it won't affect CT v1 users in Java and python etc. then.

PR #1002, opened 2015-10-29
Turn JsonOutput into a pair of free functions.

pphaneuf:


cpp/server/json_output.cc
```diff
@@ -68,12 +68,8 @@ string LogRequest(evhttp_request* req, int http_status, int resp_body_length) {
...
 
 
-JsonOutput::JsonOutput(libevent::Base* base) : base_(CHECK_NOTNULL(base)) {
-}
-
-
-void JsonOutput::SendJsonReply(evhttp_request* req, int http_status,
-                               const JsonObject& json) {
+void SendJsonReply(libevent::Base* base, evhttp_request* req, int http_status,
+                   const JsonObject& json) {
```
AlCutter: `CHECK_NOTNULL(base)` `CHECK_NOTNULL(req)`
?

cpp/server/json_output.cc
```diff
@@ -95,20 +91,20 @@ void JsonOutput::SendJsonReply(evhttp_request* req, int http_status,
...
   } else {
     send_reply();
   }
 }
 
 
-void JsonOutput::SendError(evhttp_request* req, int http_status,
-                           const string& error_msg) {
+void SendJsonError(libevent::Base* base, evhttp_request* req, int http_status,
+                   const string& error_msg) {
```
AlCutter: same here

cpp/server/json_output.cc
pphaneuf: Done.

cpp/server/json_output.cc
pphaneuf: They're passed straight through to the other function, so I'll let it do it...

AlCutter: ah yes, of course they are!

PR #1006, opened 2015-10-29
Fix build flags in go/x509 package for darwin

jacobhaven:
Currently, CGO build on darwin is failing with

```
$ go get github.com/google/certificate-transparency/go/x509
# github.com/google/certificate-transparency/go/x509
src/github.com/google/certificate-transparency/go/x509/root.go:15: undefined: initSystemRoots
src/github.com/google/certificate-transparency/go/x509/verify.go:213: c.systemVerify undefined (type *Certificate has no field or method systemVerify)
src/github.com/google/certificate-transparency/go/x509/verify.go:213: not enough arguments to return
```

The `x509/root_darwin.go` file is only being built on `!darwin` platforms... a bit backwards. :)

### DIFF
--- a/go/x509/root_darwin.go
+++ b/go/x509/root_darwin.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// +build !darwin,cgo
+// +build darwin,cgo
 
 package x509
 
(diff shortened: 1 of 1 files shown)


jacobhaven:
Actually, on link time this fails with
```
/usr/local/go/pkg/tool/darwin_amd64/link: running clang failed: exit status 1
duplicate symbol _FetchPEMRoots in:
    /var/folders/yy/pzrhg_qs08x3xyrncyx57gc80000gn/T/go-link-706434319/000001.o
    /var/folders/yy/pzrhg_qs08x3xyrncyx57gc80000gn/T/go-link-706434319/000004.o
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
FetchPEMRoots is being redefined from the one in the standard library https://github.com/golang/go/blob/master/src/crypto/x509/root_cgo_darwin.go#L23

go/x509/root_darwin.go
```diff
@@ -13,14 +13,14 @@ package x509
 #include <CoreFoundation/CoreFoundation.h>
 #include <Security/Security.h>
 
-// FetchPEMRoots fetches the system's list of trusted X.509 root certificates.
+// _FetchPEMRoots fetches the system's list of trusted X.509 root certificates.
```
jacobhaven: Name updated so as to not conflict with the `FetchPEMRoots` defined in the standard library.

pphaneuf: Might be best to avoid an underscore prefix (reserved names)? Some sufficiently obnoxious prefix should do the trick (`GoX509FetchPEMRoots`, say?)...

jacobhaven: @pphaneuf @AlCutter I've updated it to have the suffix `CTX509`, to distinguish it from other applications that might use this CGO to fetch Darwin system roots.

AlCutter:
Cool, thanks Jacob!

PR #1007, opened 2015-10-30
Fixes for FreeBSD.

trasz:
A bunch of fixes required to get this building on FreeBSD, with dependencies from ports/packages.

Sponsored by: The FreeBSD Foundation

### DIFF
--- a/cpp/net/connection_pool.cc
+++ b/cpp/net/connection_pool.cc
@@ -449,7 +449,7 @@ unique_ptr<ConnectionPool::Connection> ConnectionPool::Get(const URL& url) {
                                     kZeroMillis};
     evhtp_connection_set_timeouts(handle->connection(), &read_timeout,
                                   &write_timeout);
-    evhtp_set_hook(&handle->connection()->hooks, evhtp_hook_on_conn_error,
+    evhtp_set_hook(&handle->connection()->hooks, evhtp_hook_on_error,
                    reinterpret_cast<evhtp_hook>(
                        EvConnection::ConnectionErrorHook),
                    reinterpret_cast<void*>(conn.get()));
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -459,8 +459,8 @@ static void MakeCert() {
 // (This means the relevant section should be last in the configuration.)
 // 1.2.3.1=DER:[raw encoding of proof]
 static void WriteProofToConfig() {
-  CHECK(!FLAGS_sct_token.empty()) << google::ProgramUsage();
-  CHECK(!FLAGS_extensions_config_out.empty()) << google::ProgramUsage();
+  CHECK(!FLAGS_sct_token.empty()) << gflags::ProgramUsage();
+  CHECK(!FLAGS_extensions_config_out.empty()) << gflags::ProgramUsage();
 
   string sct;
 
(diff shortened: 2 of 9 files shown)


cpp/net/connection_pool.cc
```diff
@@ -449,7 +449,7 @@ unique_ptr<ConnectionPool::Connection> ConnectionPool::Get(const URL& url) {
                                     kZeroMillis};
     evhtp_connection_set_timeouts(handle->connection(), &read_timeout,
                                   &write_timeout);
-    evhtp_set_hook(&handle->connection()->hooks, evhtp_hook_on_conn_error,
+    evhtp_set_hook(&handle->connection()->hooks, evhtp_hook_on_error,
```
pphaneuf: @AlCutter, did evhtp rename their hooks?!?

pphaneuf:
We might have to have an autoconf test for the gflags namespace... The gflags installed on my Ubuntu still has the `google` namespace, and the newer versions have a flag to use that namespace (so `google` is the most compatible option short of having a test, if you're willing to build gflags).

AlCutter:
I think this whole PR is predicated on attempting to build CT the "old" way, using ports dependencies rather than the new 'gclient' way which I suspect sidesteps a lot of these issues. The current master branch builds just find on `FreeBSD 10.2-RELEASE` using the recommended gclient method.

cpp/net/connection_pool.cc
AlCutter: No; they added `evhtp_hook_on_conn_error` in `1.2.10`, the FreeBSD `ports/www/libevhtp` (at least on my `10.2-RELEASE` machine) is at `1.2.9`.

AlCutter:
I'm kinda surprised that this builds at all on travis, there must be some compatibility `using namespace` shenanigans going on somewhere for the `google::` vs `gflags::` issue.

The hook bit is probably either not covered by tests, or perhaps that `evhtp_hook_on_error` does get called before/after `evhtp_hook_on_conn_error`, but it'll probably *also* get called in other circumstances which the hook code will probably not be expecting and it might well blow up.

trasz:
I signed it!

trasz:
Erm, please ignore the evhtp_hook_on_conn_error thing; it was a leftover of my earlier attempts to make it build using earlier version of libevhtp.

As for the gclient -  I didn't check to see what's neccessary to make it work with the ports/packages infrastructure yet.

benlaurie:
Are you trying to make a port for CT?

If so, the gclient stuff should work OK within a port, but you'd have to get gclient itself from somewhere, of course. I don't think there's a port for depot_tools yet...

pphaneuf:
The gclient dependencies for gflags is probably pinned on a version that has the old name? Or we might be getting it from the system, accidentally?

trasz:
I've already made a port for CT; it needs some more testing (as in, runtime), but it builds properly, with dependencies and all.  This PR is an attempt to upstream local patches required for it to build.

pphaneuf:
Ok, so we've discussed this a bit here...

Fixing the hook name, that's out, since it's really adding a bug (you'll have to update your libevhtp). :smiley:

You can fix that quickly, with a `git reset --hard 5183713 && git push -f`, or something along those lines.

The other two, we're sort of interested, but not in their present form. For the gflags namespace, we'd like to remain compatible with different versions of gflags which are still out there, and they've been flipping back and forth on that, so we'd need to detect it in `autoconf`. I would definitely understand if you didn't want to do that. I just filed #1035 to track this separately.

The one about the regex conflict in gtest, that's probably the most interesting, but we'd like to expand on it a bit more. We've "fixed" this on our side by using a custom flag in libevhtp to disable their regex support, but it's not very good. The way you've done it removes future options for us, but at the moment we're not using the gtest regexes, and it would make building easier, so we could take it, *with* updates to the documentation and gclient scripts?

I might just do it, taking in your commit, and adding some commits on top?

I'll leave this pull request open, to have a place to grab it from, but once I do that, I will probably close this PR without merging the rest.

PR #1009, opened 2015-10-30
De-template SQLiteDB

pphaneuf:


benlaurie:
Don't we need templates to permit other types of data in the tree?

pphaneuf:
We're going with making the data structures more flexible (`LoggedCertificate`is now `LoggedEntry`, for example).

@AlCutter's `xjson-server` doesn't use different types, for example, it extends `LoggedEntry`.

benlaurie:
aha, fair enough.

PR #1013, opened 2015-11-02
Remove use of obsolete member __state_file in __repr__ and __str__

phad:
str(some monitor) and repr(some monitor) failed because they referenced a class member removed some time ago.  updated to remove those refs.


pphaneuf:
LGTM.

If only we had some kind of program that could process the code and make sure it's okay... :stuck_out_tongue:

PR #1014, opened 2015-11-02
Proto and serialization changes for V2 structures, should align with draft v10 rfc

Martin2112:
Updates serialization to provide V2 functionality. Main differences are that STHs can have extensions and that log entries for certs now have the same structure as precerts (see CertInfo draft doc). As an additional complication the older structures are retained in the proto and code to avoid breaking V1 APIs in other languages.

Mods outside serialization are just placeholders to avoid compiler issues. Adds tests for new stuff and strengthens some old ones. This is unfortunately fiddly code to work with and there is a reasonable chance that an error has crept in somewhere.

### DIFF
--- a/cpp/log/etcd_consistent_store-inl.h
+++ b/cpp/log/etcd_consistent_store-inl.h
@@ -275,6 +275,10 @@ bool LeafEntriesMatch(const Logged& a, const Logged& b) {
     case ct::PRECERT_ENTRY:
       return a.entry().precert_entry().pre_certificate() ==
              b.entry().precert_entry().pre_certificate();
+    case ct::PRECERT_ENTRY_V2:
+      // TODO(mhs): V2 implementation required here
+      LOG(FATAL) << "CT V2 not yet implemented";
+      break;
     case ct::X_JSON_ENTRY:
       return a.entry().x_json_entry().json() ==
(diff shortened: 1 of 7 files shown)


Martin2112:
This code is being reviewed on Rietveld. The review is at: 

https://codereview.appspot.com/270590043/

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -275,6 +275,9 @@ bool LeafEntriesMatch(const Logged& a, const Logged& b) {
     case ct::PRECERT_ENTRY:
       return a.entry().precert_entry().pre_certificate() ==
              b.entry().precert_entry().pre_certificate();
+    case ct::PRECERT_ENTRY_V2:
+      // TODO(mhs): V2 implementation required here
```
AlCutter:
```c++
LOG(FATAL) << "Not implemented";
```

(otherwise the FATAL error message below will send you on a wild goose chase :)

Martin2112:
Review is done at Reitveld. I'm planning to merge later today unless there are other comments.

Martin2112:
Another round of rework done. Extension handling has been rewritten and should now be correct.

pphaneuf:
LGTM

PR #1015, opened 2015-11-02
Break out a libserver.a ready to create a v2 server binary.

Martin2112:
Starting this work over because somebody got in first again. Split out a server library so I can do other exciting stuff in the near future.

### DIFF
--- a/cpp/server/handler.h
+++ b/cpp/server/handler.h
@@ -37,7 +37,7 @@ class HttpHandler {
               const ReadOnlyDatabase<LoggedEntry>* db,
               const ClusterStateController<LoggedEntry>* controller,
               ThreadPool* pool, libevent::Base* event_base);
-  ~HttpHandler();
+  virtual ~HttpHandler();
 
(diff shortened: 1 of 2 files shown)


cpp/server/handler.h
```diff
@@ -37,7 +37,7 @@ class HttpHandler {
               const ReadOnlyDatabase<LoggedEntry>* db,
               const ClusterStateController<LoggedEntry>* controller,
               ThreadPool* pool, libevent::Base* event_base);
-  ~HttpHandler();
+  virtual ~HttpHandler();
```
pphaneuf: Unrelated! But I'll let it slip... this time! :wink: 

Did you get a warning for it?

pphaneuf:
LGTM, pending Travis.

cpp/server/handler.h
Martin2112: Yes, it should be virtual. I can move this to the next PR though.

PR #1018, opened 2015-11-03
Improve async behavior in Prober so that it can be correctly stopped.

phad:
No tests exist for this code.  I've tried to write some but it seems very hard to test code involving the Twisted reactor.  I think some refactoring will help - injecting a reactor instance into the Prober instead of using the global 'reactor' for one.  I'll come back to that in future if that's OK.


pphaneuf:
LGTM

PR #1019, opened 2015-11-03
De-template log lookup.

pphaneuf:


pphaneuf:
I had forgotten the change needed to `x_json_handler.{h,cc}` after I had rebased it, fixed now.

PR #1022, opened 2015-11-03
De-template the Server class.

pphaneuf:
Unlike the others, the follow-up nice things are a bit more involved in this case, so I'll do them in separate PRs.


Martin2112:
Thanks for doing this. I hope it will make my changes easier.

pphaneuf:
There might still be some problems at linking, since this PR doesn't move the flags into a `.cc` file, but I've got another PR coming right after #1023 that does.

PR #1023, opened 2015-11-03
Combine the internal libraries.

pphaneuf:
There's some ordering issues between them, which could probably be fixed, but these are not installed anyway, so might as well put everything in the same library, and let the linker sort it out.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -51,7 +51,6 @@ endif
 
 noinst_LIBRARIES = \
 	cpp/libcore.a \
-	cpp/libserver.a \
 	cpp/libtest.a
 
 check_PROGRAMS = \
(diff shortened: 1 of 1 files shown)


pphaneuf:
Actually, this might be in the interest of @Martin2112, he might be more motivated to review? :wink:

Martin2112:
I broke out a server lib to try to get a handle on the dependencies. Assuming this doesn't create any linking issues then I'm OK with going back to one library.

pphaneuf:
Yeah, there was some dependencies between them, because the cut-line wasn't *exactly* right, and maybe it's fixable, maybe it isn't, but everything in one lib "fixes" it, so here we go. :smiley:

pphaneuf:
I think hit the same linking problem you had yesterday with my next PR, and this did fix it, by the way, so this should be all right.

Martin2112:
OK. Are there any more followups you have for server refactoring? Or can I go back to moving code around?

pphaneuf:
One more, just a sec, merging this one.

PR #1024, opened 2015-11-04
move the Server implementation out of the header file

pphaneuf:
Cut in two commit, the first one is a straight copy of `server.h` to `server.cc`. Turns out GitHub only detects moves, not copies, but still, it lets you see which parts I changed, and which parts are straight cut-and-paste, in the second commit.

I'll squash them before merging, this is only to help reviewing.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,6 +170,7 @@ cpp_libcore_a_SOURCES = \
 	cpp/proto/serializer.cc \
 	cpp/server/metrics.cc \
 	cpp/server/proxy.cc \
+	cpp/server/server.cc \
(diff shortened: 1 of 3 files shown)


Martin2112:
Implementations in header files eh?

pphaneuf:
Well, it *was* a template before, so you kind of have to. :smiley: 

But it isn't anymore, so let's not!

pphaneuf:
I'll do a little round of "include what you use", maybe add a few `using` statements, etc, but separately, to avoid noise in this PR (you get the nice big block cut-and-pasted without any modifications!).

Issue #880, opened 2015-08-26
Test Flake: EtcdConsistentStoreTest.TestRejectsAddsWhenOverCapacity


Martin2112:
Just saw this fail then pass when rerun. Log snippet follows.

```
[ RUN      ] EtcdConsistentStoreTest.TestRejectsAddsWhenOverCapacity cpp/log/etcd_consistent_store_test.cc:782: Failure
Value of: GetNumEtcdEntries()
  Actual: 201
Expected: 0
[  FAILED  ] EtcdConsistentStoreTest.TestRejectsAddsWhenOverCapacity (2002 ms)
[----------] 23 tests from EtcdConsistentStoreTest (6018 ms total)

[----------] Global test environment tear-down
[==========] 31 tests from 2 test cases ran. (6919 ms total)
[  PASSED  ] 30 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] EtcdConsistentStoreTest.TestRejectsAddsWhenOverCapacity
```

Martin2112:
Exhibit B:

F1104 11:33:25.355763 10203 etcd_consistent_store_test.cc:579] Check failed: false 
*** Check failure stack trace: ***
F1104 11:33:25.355797 10202 etcd_consistent_store_test.cc:579] Check failed: false 
*** Check failure stack trace: ***
    @     0x7ff27c6d2dbd  google::LogMessage::Fail()
    @     0x7ff27c6d2dbd  google::LogMessage::Fail()
    @     0x7ff27c6d4c5d  google::LogMessage::SendToLog()
    @     0x7ff27c6d4c5d  google::LogMessage::SendToLog()
    @     0x7ff27c6d29ac  google::LogMessage::Flush()
    @     0x7ff27c6d29ac  google::LogMessage::Flush()
    @     0x7ff27c6d557e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7ff27c6d557e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7ff27df3af38  _ZNSt17_Function_handlerIFvRKN10cert_trans6UpdateIN2ct14SignedTreeHeadEEEEZNS0_44EtcdConsistentStoreTest_WatchServingSTH_Test8TestBodyEvEUlS6_E_E9_M_invokeERKSt9_Any_dataS6_
[...]

PR #1027, opened 2015-11-04
Fix up a couple of test flakes

AlCutter:
There are a bunch of `sleep()` sync things which are brittle, but it's pretty complicated to hang off of any event which guarantees that the prior stuff has completed.


cpp/log/cluster_state_controller_test.cc
```diff
@@ -379,7 +379,7 @@ TEST_F(ClusterStateControllerTest, TestUsesLargestSTHWithIdenticalTimestamp) {
   cns3.mutable_newest_sth()->set_timestamp(1004);
   cns3.mutable_newest_sth()->set_tree_size(999);
   store3_->SetClusterNodeState(cns3);
-  sleep(1);
+  sleep(2);
```
Martin2112: Let's move quickly past that one.

Martin2112:
The etcd test fixes look pretty solid for both cases in #880.

PR #1030, opened 2015-11-05
Untangle the server / handler registration a bit.

Martin2112:
Remove RegisterHandler and have the servers do the same thing with one
less layer of indirection. Server no longer refers to HttpHandler purely
for this.

### DIFF
--- a/cpp/server/server.cc
+++ b/cpp/server/server.cc
@@ -14,7 +14,6 @@
 #include "log/cluster_state_controller.h"
 #include "log/etcd_consistent_store.h"
 #include "monitoring/monitoring.h"
-#include "server/handler.h"
 #include "util/thread_pool.h"
 #include "util/uuid.h"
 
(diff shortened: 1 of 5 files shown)


cpp/server/server.cc
```diff
@@ -191,6 +183,14 @@ ContinuousFetcher* Server::continuous_fetcher() {
   return fetcher_.get();
 }
 
+Proxy* Server::proxy() {
```
pphaneuf: Needs two lines between top level declarations/definitions.

cpp/server/server.cc
```diff
@@ -191,6 +183,14 @@ ContinuousFetcher* Server::continuous_fetcher() {
...
+  return proxy_.get();
+}
+
```
pphaneuf: Here too.

pphaneuf:
You might be able to get rid of the `#include "server/handler.h"` in `server.cc`, no? Otherwise done with the review, just those nits. :smiley:

Martin2112:
All done.

pphaneuf:
LGTM

PR #1031, opened 2015-11-05
Refactoring to pull out code + flags that creates a database

Martin2112:
I'm expecting to pull more of the flag interactions code into the helper class in future PRs. I know the name server_helper isn't great but naming things is hard!

### DIFF
--- a/cpp/server/ct-mirror.cc
+++ b/cpp/server/ct-mirror.cc
@@ -25,11 +25,7 @@
 #include "log/ct_extensions.h"
 #include "log/database.h"
 #include "log/etcd_consistent_store.h"
-#include "log/file_db.h"
-#include "log/file_storage.h"
-#include "log/leveldb_db.h"
 #include "log/log_lookup.h"
-#include "log/sqlite_db.h"
 #include "log/strict_consistent_store.h"
 #include "merkletree/compact_merkle_tree.h"
 #include "merkletree/merkle_verifier.h"
@@ -104,7 +100,6 @@ using cert_trans::RemotePeer;
 using cert_trans::SQLiteDB;
 using cert_trans::ScopedLatency;
 using cert_trans::Server;
-using cert_trans::ServerHelper;
 using cert_trans::SplitHosts;
 using cert_trans::StrictConsistentStore;
 using cert_trans::ThreadPool;
@@ -171,14 +166,6 @@ static bool ValidateRead(const char* flagname, const string& path) {
 
 static const bool pubkey_dummy =
     RegisterFlagValidator(&FLAGS_target_public_key, &ValidateRead);
-
-static bool ValidateIsNonNegative(const char* flagname, int value) {
-  if (value < 0) {
-    std::cout << flagname << " must not be negative" << std::endl;
-    return false;
-  }
-  return true;
-}
 
 static bool ValidateIsPositive(const char* flagname, int value) {
   if (value <= 0) {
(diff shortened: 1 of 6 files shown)


pphaneuf:
Reviewed the first commit, moving on.

cpp/server/ct-mirror.cc
```diff
@@ -331,40 +295,19 @@ int main(int argc, char* argv[]) {
...
 
-  const std::unique_ptr<EtcdClient> etcd_client(
-      stand_alone_mode
-          ? new FakeEtcdClient(event_base.get())
-          : new EtcdClient(&internal_pool, &url_fetcher,
-                           SplitHosts(FLAGS_etcd_servers)));
+  const unique_ptr<EtcdClient> etcd_client(
+      stand_alone_mode ? new FakeEtcdClient(event_base.get())
+                       : new EtcdClient(&internal_pool, &url_fetcher,
+                                        SplitHosts(FLAGS_etcd_servers)));
```
pphaneuf: Is this just whitespace reformatting? If so, grr...

cpp/server/ct-server.cc
```diff
@@ -386,11 +323,10 @@ int main(int argc, char* argv[]) {
...
 
-  std::unique_ptr<EtcdClient> etcd_client(
-      stand_alone_mode
-          ? new FakeEtcdClient(event_base.get())
-          : new EtcdClient(&internal_pool, &url_fetcher,
-                           SplitHosts(FLAGS_etcd_servers)));
+  unique_ptr<EtcdClient> etcd_client(
+      stand_alone_mode ? new FakeEtcdClient(event_base.get())
+                       : new EtcdClient(&internal_pool, &url_fetcher,
+                                        SplitHosts(FLAGS_etcd_servers)));
```
pphaneuf: All this noise!

cpp/server/ct-mirror.cc
Martin2112: No I removed a 'std::' and clang-format did the rest.

cpp/server/ct-server.cc
Martin2112: OK that one is clang-format.

pphaneuf:
I generally like the "unique_ptr for db reference" commit (other than the noise), but I can't help but think it'd be even better if `ProvideDatabase` returned a `unique_ptr<Database>`. In "modern C++", that's a common pattern to return a smart pointer, so that even if you screw up and ignore the return value, you don't leak.

pphaneuf:
General comment: I find it nicer to have two blank lines between most top level things, à la Python. Look at `database.h` for an example.

pphaneuf:
All right, went over all this, let me know when you've pushed some updates.

cpp/server/ct-mirror.cc
pphaneuf: I approve of removing that `std::`, but I disapprove of not doing it separately. :wink:

cpp/server/ct-server.cc
pphaneuf: There was an `std::` here too! Oh, how I'd love highlights *within* a diff line...

Martin2112:
I'm not as keen on this because formatters don't support it and as I use an
IDE I can navigate around code easily without worrying about its fiddly
layout.

Plus most of our code doesn't have this so any alleged benefits we might
get from it don't currently exist anyway. And it's something done by Python
programmers (guilt by association). I rest my case m'lud.

Martin2112:
ProvideDatabase() now returns a unique_ptr

Martin2112:
Think it's mostly done. PTAL.

pphaneuf:
LGTM, but I'm very curious to see what rebasing/squashing/merging strategy you'll take... :wink: 

image

Martin2112:
I think I shall be using the "Friday Afternoon" strategy.

PR #1033, opened 2015-11-06
Add a simple first cut of a SparseMerkleTree.

AlCutter:
Very early first cut of a C++ sparse merkle tree / map implementation.
Inspired by the [revocation transparency](http://www.links.org/files/RevocationTransparency.pdf) paper, but with some differences (as noted in the header) for security and to be useful as a general purpose verifiable map.

Lots left to do, just trying to break it up into digestible chunks.

### DIFF
--- a/cpp/merkletree/sparse_merkle_tree.cc
+++ b/cpp/merkletree/sparse_merkle_tree.cc
@@ -1,0 +1,217 @@
+#include "cpp/merkletree/sparse_merkle_tree.h"
+
+#include <stddef.h>
+#include <algorithm>
+#include <vector>
+
+#include "merkletree/merkle_tree_math.h"
+#include "util/util.h"
+
+using std::make_pair;
+using std::ostream;
+using std::ostringstream;
+using std::reverse;
+using std::string;
+using std::unique_ptr;
+using std::unordered_map;
+using std::vector;
+
+
+const vector<string>* GetNullHashes(const TreeHasher& hasher) {
+  static unique_ptr<const vector<string>> null_hashes;
+  if (!null_hashes) {
+    vector<string> r{hasher.HashLeaf("")};
+    for (int i(1); i < hasher.DigestSize() * 8; ++i) {
+      r.emplace_back(hasher.HashChildren(r.back(), r.back()));
+    }
+    reverse(r.begin(), r.end());
+    null_hashes.reset(new vector<string>(std::move(r)));
+  }
+  return null_hashes.get();
+}
+
+
+SparseMerkleTree::SparseMerkleTree(SerialHasher* hasher)
+    : serial_hasher_(CHECK_NOTNULL(hasher)->Create()),
+      treehasher_(hasher),
+      null_hashes_(GetNullHashes(treehasher_)) {
+}
+
+
+void SparseMerkleTree::EnsureHaveLevel(size_t level) {
+  if (tree_.size() < (level + 1)) {
+    tree_.resize(level + 1);
+  }
+}
+
+
+void SparseMerkleTree::SetLeafByHash(const string& key, const string& data) {
+  serial_hasher_->Reset();
+  serial_hasher_->Update(key);
+  return SetLeaf(PathFromBytes(serial_hasher_->Final()), data);
+}
+
+
+void SparseMerkleTree::SetLeaf(const Path& path, const string& data) {
+  CHECK_EQ(treehasher_.DigestSize(), path.size());
+  // Mark the tree dirty:
+  root_hash_.clear();
+
+  IndexType node_index(0);
+  for (int depth(0); depth <= kDigestSizeBits; ++depth) {
+    node_index += PathBit(path, depth);
+    EnsureHaveLevel(depth);
+    auto it(tree_[depth].find(node_index));
+    if (it == tree_[depth].end()) {
+      CHECK(tree_[depth]
+                .emplace(make_pair(node_index, TreeNode(path, data)))
+                .second);
+      return;
+    } else if (it->second.type_ == TreeNode::INTERNAL) {
+      // Mark the internal node hash dirty
+      it->second.hash_.reset();
+    } else if (it->second.leaf_->path_ == path) {
+      // replacement
+      CHECK_EQ(TreeNode::LEAF, it->second.type_);
+      it->second.leaf_->value_ = data;
+      return;
+    } else {
+      // restructure: push the existing node down a level and replace this one
+      // with an INTERNAL node
+      CHECK_LT(depth, kDigestSizeBits);
+      EnsureHaveLevel(depth + 1);
+      IndexType child_index((node_index << 1) +
+                            PathBit(it->second.leaf_->path_, depth + 1));
+      CHECK(tree_[depth + 1]
+                .emplace(make_pair(child_index, std::move(it->second)))
+                .second);
+      it->second.type_ = TreeNode::INTERNAL;
+      it->second.leaf_.reset();
+    }
+    node_index <<= 1;
+  }
+  LOG(FATAL) << "Failed to set " << path << " to " << data;
+}
+
+
+void SparseMerkleTree::DumpTree(ostream* os, size_t depth,
+                                IndexType index) const {
+  if (tree_.size() <= depth) {
+    return;
+  }
+  const string indent((depth + 1) * 2, '-');
+  for (int side(0); side < 2; ++side) {
+    auto child(tree_[depth].find(index + side));
+    if (child != tree_[depth].end()) {
+      *os << indent << side << ": " << child->second.DebugString() << "\n";
+      DumpTree(os, depth + 1, (index + side) << 1);
+    }
+  }
+}
+
+
+string SparseMerkleTree::Dump() const {
+  ostringstream ret;
+  ret << "\nTree [Root: " << util::ToBase64(root_hash_) << "]:\n";
+  if (!tree_.empty()) {
+    DumpTree(&ret, 0, 0);
+  }
+  return ret.str();
+}
+
+
+string SparseMerkleTree::CalculateSubtreeHash(size_t depth, IndexType index) {
+  if (tree_.size() <= depth) {
+    return null_hashes_->at(depth);
+  }
+
+  auto it(tree_[depth].find(index));
+  if (it != tree_[depth].end()) {
+    switch (it->second.type_) {
+      case TreeNode::INTERNAL: {
+        if (it->second.hash_) {
+          return *(it->second.hash_);
+        }
+        IndexType left_child_index(index << 1);
+        const string left(CalculateSubtreeHash(depth + 1, left_child_index));
+        const string right(
+            CalculateSubtreeHash(depth + 1, left_child_index + 1));
+        it->second.hash_.reset(
+            new string(treehasher_.HashChildren(left, right)));
+        return *(it->second.hash_);
+      }
+
+      case TreeNode::LEAF: {
+        string ret(treehasher_.HashLeaf(it->second.leaf_->value_));
+        for (int i(kDigestSizeBits - 1); i > depth; --i) {
+          if (PathBit(it->second.leaf_->path_, i) == 0) {
+            ret = treehasher_.HashChildren(ret, null_hashes_->at(i));
+          } else {
+            ret = treehasher_.HashChildren(null_hashes_->at(i), ret);
+          }
+        }
+        // TODO(alcutter): maybe cache this?
+        return ret;
+      }
+    }
+    LOG(FATAL) << "Unknown node type " << it->second.type_ << " !";
+  }
+
+  return null_hashes_->at(depth);
+}
+
+
+string SparseMerkleTree::CurrentRoot() {
+  if (root_hash_.empty()) {
+    root_hash_ = treehasher_.HashChildren(CalculateSubtreeHash(0, 0),
+                                          CalculateSubtreeHash(0, 1));
+  }
+  return root_hash_;
+}
+
+
+std::vector<string> SparseMerkleTree::InclusionProof(const Path& path) {
+  // TODO(alcutter): implement
+  LOG(FATAL) << "Not implemented.";
+}
+
+
+string SparseMerkleTree::TreeNode::DebugString() const {
+  ostringstream os;
+  os << "[TreeNode ";
+  switch (type_) {
+    case INTERNAL:
+      os << "INTL hash: ";
+      if (hash_) {
+        os << util::ToBase64(*hash_);
+      } else {
+        os << "(unset)";
+      }
+      break;
+    case LEAF:
+      os << "LEAF leaf: ";
+      os << leaf_->DebugString();
+      break;
+  }
+  os << "]";
+  return os.str();
+}
+
+
+string SparseMerkleTree::Leaf::DebugString() const {
+  ostringstream os;
+  os << "[Leaf path: " << path_ << " value: " << value_ << "]";
+  return os.str();
+}
+
+
+ostream& operator<<(ostream& out, const SparseMerkleTree::Path& path) {
+  for (size_t i(0); i < path.size(); ++i) {
+    uint8_t t(path[i]);
+    for (size_t b(8); b > 0; --b) {
+      out << (t & 0x80 ? '1' : '0');
+      t <<= 1;
+    }
+  }
+  return out;
+}
--- a/cpp/merkletree/sparse_merkle_tree.h
+++ b/cpp/merkletree/sparse_merkle_tree.h
@@ -1,0 +1,263 @@
+#ifndef CERT_TRANS_MERKLETREE_SPARSE_MERKLE_TREE_H
+#define CERT_TRANS_MERKLETREE_SPARSE_MERKLE_TREE_H
+
+#include <glog/logging.h>
+#include <stddef.h>
+#include <array>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "merkletree/merkle_tree_interface.h"
+#include "merkletree/tree_hasher.h"
+
+class SerialHasher;
+
+
+// Calculates the set of "null" hashes:
+// ...H(H(H("")||H(""))||H("")||(H(""))||...)...
+//
+// Visible out here because it's useful for testing too.
+const std::vector<std::string>* GetNullHashes(const TreeHasher& hasher);
+
+// Implementation of a Sparse Merkle Tree.
+//
+// The design is inspired by the tree described in
+// http://www.links.org/files/RevocationTransparency.pdf), but with some
+// tweaks, most notably:
+//   1) Leaf values are hashed before being incorporated into the tree.
+//   2) Similar to the way it works in the CT MerkleTree, hashes are domain
+//      separated by prefixing the preimage with \x00 for leaves, and \x01 for
+//      internal nodes.
+//
+//
+// These mean that level 2 nodes are of the form:
+//      H(\x01||H(\0x00||valueL)||H(\0x00||valueR))
+// and so on.
+//
+// Nodes are addressed by a Path, which is a bit-string of the same length as
+// the output of the hashing fuction used.  This string describes a path down
+// from the root to a leaf, with the 0-bits indicating the path takes the
+// left-hand child branch, and 1-bits the right. e.g:
+//        Root
+//        /  \
+//      0/    \1
+//      /      \
+//     i0      i3
+//   0/ \1   0/  \1
+//   /   \   /    \
+//  l0  l1  l2    l3
+//
+//  The paths to the 4 leaves would then be:
+//  l0: "00"
+//  l1: "01"
+//  l2: "10"
+//  l3: "11"
+//
+// To help with memory consumption, leaves inserted into the tree are stored
+// at the first unused node along their path.  An example is given below:
+//
+// * Empty tree:
+//      Root
+//
+// * Add "10" = "hi":
+// Since the tree is empty, the first bit of the added path is sufficient to
+// identify a unique prefix, so the leaf is stored as the "1" entry
+// immediately below the root.
+//              Root
+//                |
+//                |_______1
+//                       p:"10"
+//                       v:"hi"
+//
+// * Add "11" = "to":
+// The first bit of the added path is not enough to provide a unique
+// prefix so the leaf node currently occupying the "1" node immediately below
+// the root must be pushed down a level, resulting in:
+//              Root
+//                |
+//                |_______1
+//                        |
+//                        |
+//                  0_____|_____1
+//                  |           |
+//                p:"10"      p:"11"
+//                v:"hi"      v:"to"
+//
+// (In the case where paths are longer and multiple bits of the prefix collide,
+// the existing node is repeated pushed down a level until a unique prefix is
+// found.)
+//
+// * Add "00" = "aa":
+// The first bit of the added path is unique, and so the resulting tree is:
+//              Root
+//                |
+//        0_______|_______1
+//        |               |
+//      p:"00"            |
+//      v:"aa"      0_____|_____1
+//                  |           |
+//                p:"10"      p:"11"
+//                v:"hi"      v:"to"
+//
+// * Calculating the root hash
+// Calculating the root of the tree is similar to a regular MerkleTree, but is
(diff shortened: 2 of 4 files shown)


cpp/merkletree/sparse_merkle_tree_test.cc
```diff
@@ -0,0 +1,315 @@
...
+  const string smt_root(tree_.CurrentRoot());
+  LOG(INFO) << "Calculating Reference Root";
+  const string ref_root(ref.HStar2(256, values));
+  LOG(INFO) << "Comparing";
+  EXPECT_EQ(ToBase64(ref_root), ToBase64(smt_root));
+}
+
+
+/*
+These are mainly useful to measuring memory consumption, but they take a while
```
pphaneuf: Use `DISABLED_`, so they're still compiled. And if you're going to disable a big block of code like that, always use `#if 0` anyway, to protect against comments in the block!

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,243 @@
...
+// Implementation of a Sparse Merkle Tree.
+//
+// The design is inspired by the tree described in
+// http://www.links.org/files/RevocationTransparency.pdf), but with some
+// of tweaks, most notably:
+//   1) Leaf values are hashed before being incorporated into the tree.
+//   2) Similar to the way it works in the CT MerkleTree, hashes are domain
+//      separated by prefixing the preimage with \x00 for leaves, and \x01 for
+//      internal nodes.  (This prevents generation of maps with identical root
+//      hashes by manipulating leaf data between neighbouring nodes.)
```
aeijdenberg: I don't quite understand while this is necessary.  Unlike CT MerkleTrees, the tree here is always of a fixed size and thus I wouldn't expect it to be susceptible in the same manner.   Can you give a specific example of what you were worried about?

pphaneuf: @AlCutter explained it to me in person... He tells me that in the paper, it's just the hash of the two concatenated bits of data, so if you have two leafs under the same node that contain "hello" and "Al", you could change them to contain "hell" and "oAl", and all the hashes would remain the same.

Not sure if it really needs to be clarified *here*, since this is explaining how some *other* thing works?

pphaneuf:
That looks bad:

```
[==========] Running 5 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 5 tests from SparseMerkleTreeTest
[ RUN      ] SparseMerkleTreeTest.ReferenceEmptyTreeRootKAT
[       OK ] SparseMerkleTreeTest.ReferenceEmptyTreeRootKAT (4 ms)
[ RUN      ] SparseMerkleTreeTest.EmptyTreeRootKAT
[       OK ] SparseMerkleTreeTest.EmptyTreeRootKAT (2 ms)
[ RUN      ] SparseMerkleTreeTest.SimpleTest
[       OK ] SparseMerkleTreeTest.SimpleTest (16 ms)
[ RUN      ] SparseMerkleTreeTest.RandomReferenceTest
/bin/bash: line 5: 16079 Killed                  ${dir}$tst
FAIL: cpp/merkletree/sparse_merkle_tree_test
```

https://s3.amazonaws.com/archive.travis-ci.org/jobs/89673013/log.txt

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+}
+
+
+string SparseMerkleTree::Dump() const {
+  ostringstream ret;
+  ret << "\nTree [Root: " << util::ToBase64(root_hash_) << "]:\n";
+  if (!tree_.empty()) {
+    DumpTree(&ret, 0, 0);
```
pphaneuf: Is that really wise?

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,243 @@
...
+  explicit SparseMerkleTree(SerialHasher* hasher);
+
+  // Length of a node (i.e., a hash), in bytes.
+  virtual size_t NodeSize() const {
+    return treehasher_.DigestSize();
+  };
+
+  // Number of non-default leaves in the tree.
+  virtual size_t LeafCount() const {
+    LOG(FATAL) << "Not implemented.";
```
pphaneuf: This seems hopeless, I'd ding it entirely, apply YAGNI, and really hope you never do need it.

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,243 @@
...
+  std::vector<std::string> PathToCurrentRoot(size_t leaf);
+
+  std::string Dump() const;
+
+ private:
+  // WARNING WARNING WARNING
+  // 64 < 256 !
+  // WARNING WARNING WARNING
+  // TODO(alcutter): BIGNUM probably.
+  typedef uint64_t IndexType;
```
pphaneuf: We did mention an entirely different implementation... While I don't advise doing it, I think it would be handy to have a think about "what if this was storing its data in LevelDB?". And the other way sounds like it would fit that too, and still behave fairly reasonably in-memory?

As a reminder, the idea was to store things in an `std::map`, with the full path as the key. That could use more data, but since the entries would be ordered, I think it would be possible to lazily compute the internal nodes (possibly every time)?

pphaneuf:
I think I'll let you have a swipe at it before continuing.

cpp/merkletree/sparse_merkle_tree.h
aeijdenberg: @pphaneuf that explains (1) (and makes total sense) but I don't think it explains (2).

AlCutter: @pphaneuf, (1) [also] fixes that.
@aeijdenberg it's because they're doing different things - one is reducing the leaf values into leaf hashes, and the other is building the tree itself, so it's nice to keep the two domains separate.  You can think of it like 'strong typing' for hashes because in effect it's as if we had two different hashing functions.

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: Yeeeaaaahhhh, it's fine :)
(note that it doesn't print the *whole* tree - just the tree representation as it is in memory, so no "default" nodes are printed, and the leaves are printed in their "unique prefix" node homes, e.g.:

```
Tree [Root: UxrRt9C5yoOtWK2JWCMLVQD5Ks923n8AftFItEdlP6Q=]:
--0: [TreeNode LEAF leaf: [Leaf path: 0101111101010101110100110101010010111001000100111001001110101011101001000011001010010010100010110011000110111010001011100110110100001100101010010100101000110110101010100010100111001000010101100100001010110010001111011100000110110000100111010101101110001011 value: three]]
--1: [TreeNode INTL hash: (unset)]
----1: [TreeNode INTL hash: (unset)]
------1: [TreeNode INTL hash: (unset)]
--------0: [TreeNode LEAF leaf: [Leaf path: 1110110100110001111101000011111111010111101011011100000001000011011100010110111000001100111010100011010000110010000100101000011110001000110011010110011011101110001110100000101100000010001111001000000010111011010000000011010110101101110000111001001001110110 value: one]]
--------1: [TreeNode LEAF leaf: [Leaf path: 1111001111100010100001011110011011011011101100000111011011010011110100111101011111000001000111001101010011101101100011010110110001100101111100000000111111110011011100011001111111011001110000001110001001110000010110000111010011111110110011001100010000111111 value: two]]
```

cpp/merkletree/sparse_merkle_tree.h
aeijdenberg: @AlCutter - thanks.  Should we strike the text in parentheses then? (or maybe I'm still missing something?)

cpp/merkletree/sparse_merkle_tree.h
AlCutter: Yeah, so I actually have another branch with that implemented, currently it's slower but I suspect I know why that is so I'll send it out once I've fixed it.

That would then give a very tight API for cache/page/DB-ability for nodes.

cpp/merkletree/sparse_merkle_tree.h
AlCutter: It's probably in the wrong place, but either (1) or (2) protects against the shuffling data so I'll just move the test up to the preceding paragraph.

cpp/merkletree/sparse_merkle_tree.h
AlCutter: Seems like it'd be pretty easy to keep track of, but I'll ding it for now until I know for sure we want it.

AlCutter:
:point_right: prod prod

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,230 @@
...
+
+    std::string DebugString() const;
+
+    enum { INTERNAL, LEAF } type_;
+    // TODO(alcutter): sort this out
+    std::unique_ptr<std::string> hash_;
+    std::unique_ptr<Leaf> leaf_;
+  };
+
+
```
pphaneuf: One line too many.

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,214 @@
...
+  }
+  reverse(ret.begin(), ret.end());
+  return ret;
+}
+
+
+Path PathFromBytes(const std::string& bytes) {
+  Path path;
+  for (int i(0); i < path.size(); ++i) {
+    path[i] = (bytes[i / 8] >> (i % 8)) & 0x01;
```
aeijdenberg: How about:
>> (-(i + 1) % 8)
instead of:
>> (i % 8)

That'll give 7, 6, 5, 4, 3, 2, 1, 0, 7, ...

It also seems as though the result of this is being used as the path from the node itself back up through the tree, whereas I'd have expected it to be the other way.  Then again, maybe this makes the inclusion proof code more intuitive?

cpp/merkletree/sparse_merkle_tree_test.cc
AlCutter: Done.

cpp/merkletree/sparse_merkle_tree.h
AlCutter: Dinged.

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: Bug - well spotted

aeijdenberg: Thanks - the new code works as expected.  I'll note that it wasn't *just* the bits in the byte that were around backwards, but I think another part of your code traversed the array in the opposite order as well.  It looks like you changed that too and this now matches my previous Python implementation.

pphaneuf:
I think you might need to add some include? On Mac, Travis says this:

```
In file included from cpp/merkletree/sparse_merkle_tree.cc:1:
./cpp/merkletree/sparse_merkle_tree.h:198:16: error: implicit instantiation of undefined template 'std::__1::array<unsigned char, 32>'
    const Path path_;
               ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__tuple:69:65: note: template is declared here
template <class _Tp, size_t _Size> struct _LIBCPP_TYPE_VIS_ONLY array;
                                                                ^
```

https://api.travis-ci.org/jobs/90952101/log.txt

AlCutter:
prod prod prod :fast_forward:

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,265 @@
...
+// Creates a Path from the bits passed in.
+inline SparseMerkleTree::Path PathFromBytes(const std::string& bytes) {
+  SparseMerkleTree::Path path;
+  // Path size must be a multiple of 8 for now.
+  CHECK_EQ(bytes.size(), path.size());
+
+  for (int i(0); i < path.size(); ++i) {
+    path[i] = bytes[i];
+  }
```
pphaneuf: This is `std::copy(bytes.begin(), bytes.end(), path.begin());`, I believe.

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,265 @@
...
+  return path;
+}
+
+
+// Extracts the |n|th bit from |path|
```
pphaneuf: Maybe add the words "most significant" in there? :smiley:

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,265 @@
...
+inline int PathBit(const SparseMerkleTree::Path& path, size_t bit) {
```
pphaneuf: Not sure why, but I get this vibe that I'd like `bool` to be the closest thing to a "bit"? Would let you ding the ternary operator below.

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+using std::reverse;
+using std::string;
+using std::unique_ptr;
+using std::unordered_map;
+using std::vector;
+
+
+const vector<string>& GetNullHashes(const TreeHasher& hasher) {
+  static unique_ptr<const vector<string>> null_hashes;
+  if (!null_hashes.get()) {
```
pphaneuf: I think you can skip the `.get()` here.

pphaneuf: Or maybe not. Have a go!

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+    vector<string> r{hasher.HashLeaf("")};
+    for (int i(1); i < hasher.DigestSize() * 8; ++i) {
+      r.emplace_back(hasher.HashChildren(r.back(), r.back()));
+    }
+    reverse(r.begin(), r.end());
+    null_hashes.reset(new vector<string>(r.begin(), r.end()));
```
pphaneuf: A copy?!? We are not savages!

```C++
null_hashes.reset(new vector<string>(move(r)));
```

Also, you're avoiding building it in-place to try to dodge threading issues, aren't you? Hahaha! As if.

pphaneuf: Or you could skip the `reverse()` and initialize it with `r.rbegin(), r.rend()`, maybe?

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+  }
+  return *null_hashes.get();
+}
+
+
+SparseMerkleTree::SparseMerkleTree(SerialHasher* hasher)
+    : serial_hasher_(CHECK_NOTNULL(hasher)->Create()),
+      treehasher_(hasher),
```
pphaneuf: `CHECK_NOTNULL`.

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,265 @@
...
+  // Get the |index|-th node at level |level|. Indexing starts at 0;
+  // caller is responsible for ensuring tree is sufficiently up to date.
+  std::string Node(size_t level, size_t index) const;
+
+  // Maybe add a new tree level.
+  void EnsureHaveLevel(size_t n);
+
+  std::unique_ptr<SerialHasher> serial_hasher_;
+  TreeHasher treehasher_;
+  const std::vector<std::string>& null_hashes_;
```
pphaneuf: I find this line very disturbing to look at, because const references are allowed to be temporaries, but putting them in a `struct`/`class` doesn't extend the lifetime of the temporaries. I think.

http://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary

Might be best to be dealing in (const) pointers for this?

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+      null_hashes_(GetNullHashes(treehasher_)),
+      root_hash_(null_hashes_[0]) {
+}
+
+
+void SparseMerkleTree::EnsureHaveLevel(size_t level) {
+  if (tree_.size() < (level + 1)) {
+    tree_.emplace_back(unordered_map<IndexType, TreeNode>());
```
pphaneuf: This is only good if you're one off the desired size, which I think might always the case, looking at it below, but you could make it live up to its name by using `vector::resize`?

cpp/merkletree/sparse_merkle_tree.h
AlCutter: It is, you're right.

cpp/merkletree/sparse_merkle_tree.h
AlCutter: done.

cpp/merkletree/sparse_merkle_tree.h
AlCutter: But then I'd have to add the ternary operator in (currently) 3 extra places in `sparse_merkle_tree.cc` instead...

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: Dinged!

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: I am, occasionally.
std::moving, now, anyway.

cpp/merkletree/sparse_merkle_tree.h
AlCutter: Changed.

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: I think if that ever happened it'd be because of a bug elsewhere, but I'll change it for `resize()` anyway.

AlCutter:
PTAL

cpp/merkletree/sparse_merkle_tree.cc
pphaneuf: If `unique_ptr<>::reset` isn't atomic, this could be Bad (tm) if two threads were trying to race for the construction of the first `SparseMerkleTree`.

My recommendation is "don't do that". :wink:

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+    }
+    LOG(FATAL) << "Unknown node type " << it->second.type_ << " !";
+  }
+
+  return null_hashes_->at(depth);
+}
+
+
+string SparseMerkleTree::CurrentRoot() {
+  root_hash_ = treehasher_.HashChildren(CalculateSubtreeHash(0, 0),
```
pphaneuf: Only calculate `root_hash_` if it's `empty()`?

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+
+
+void SparseMerkleTree::SetLeaf(const Path& path, const string& data) {
+  CHECK_EQ(treehasher_.DigestSize(), path.size());
+  // Mark the tree dirty:
+  root_hash_.clear();
+
+  IndexType node_index(0);
+  for (int depth(0); depth <= kDigestSizeBits; ++depth) {
+    int side(PathBit(path, depth));
```
pphaneuf: Make `const`, or inline (only used on the next line)?

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -0,0 +1,263 @@
...
+  std::string Node(size_t level, size_t index) const;
+
+  // Maybe add a new tree level.
+  void EnsureHaveLevel(size_t n);
+
+  std::unique_ptr<SerialHasher> serial_hasher_;
+  TreeHasher treehasher_;
+  const std::vector<std::string>* const null_hashes_;
+  // TODO(alcutter): investigate other structures
+  std::vector<std::unordered_map<IndexType, TreeNode>> tree_;
```
pphaneuf: Would it be that much more expensive to just have an `std::array` of `kDigestSizeBits`, right off? Could have done without `EnsureHaveLevel`, etc...

Not saying to do that, just asking. :smiley:

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+      CHECK(tree_[depth]
+                .emplace(make_pair(node_index, TreeNode(path, data)))
+                .second);
+      return;
+    } else if (it->second.type_ == TreeNode::INTERNAL) {
+      // Mark the internal node hash dirty
+      it->second.hash_.reset();
+    } else if (it->second.leaf_->path_ == path) {
+      // replacement
+      CHECK_EQ(TreeNode::LEAF, it->second.type_);
```
pphaneuf: If you're here, that's implicitly true, because of line 72, right? And you'd also have already crashed at line 75 (because `it->second.leaf_` would likely have been null)?

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+      it->second.leaf_->value_ = data;
+      return;
+    } else {
+      // restructure: push the existing node down a level and replace this one
```
pphaneuf: Add a `CHECK_LT(depth, kDigestSizeBits)` here?

cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -0,0 +1,216 @@
...
+}
+
+
+void SparseMerkleTree::SetLeaf(const Path& path, const string& data) {
+  CHECK_EQ(treehasher_.DigestSize(), path.size());
+  // Mark the tree dirty:
+  root_hash_.clear();
+
+  IndexType node_index(0);
+  for (int depth(0); depth <= kDigestSizeBits; ++depth) {
```
pphaneuf: Maybe have a `CHECK_LT(depth, sizeof(node_index) * 8)` (or something like that)?

cpp/merkletree/sparse_merkle_tree.h
pphaneuf: I know it's going away and everything, but it'd be great if you explained that type a bit better, like how for example it's right-shifted to have the useless bits be the most significants (it's not a truncated path, which is sort of what I expected).

pphaneuf: And that it's important the the unique bits be truncated out, even if it was just a truncated path, since that's important.

pphaneuf:
Done with this round of reviewing.

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: Done

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: done.

cpp/merkletree/sparse_merkle_tree.h
AlCutter: It would be fine (as in I don't think it'd break the algorithm as it stands), but in most cases you'll never need a tree that deep since you just need to have a unique prefix.

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: Yep, this was just precaution against future me messing the conditionals up.

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: Done

cpp/merkletree/sparse_merkle_tree.cc
AlCutter: Don't want to do that - there's a big WARNING comment about how 64 is less that 256 in the header somewhere, I'll be replacing the `uin64_t` which `IndexType` currently is defined as with a `BigNum` wrapper RSN.

pphaneuf:
LGTM, pending Travis!

PR #1038, opened 2015-11-09
Rename LogVerifier::VerifyResult to LogVerifyResult.

pphaneuf:
There was no less than *three* things VerifyResult, one less is progress!

### DIFF
--- a/cpp/client/ct.cc
+++ b/cpp/client/ct.cc
@@ -259,7 +259,7 @@ static bool VerifySCTAndPopulateSSLClientCTData(
   sct_info->mutable_sct()->CopyFrom(sct);
   const unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
   string merkle_leaf;
-  LogVerifier::VerifyResult result =
+  LogVerifier::LogVerifyResult result =
       verifier->VerifySignedCertificateTimestamp(
           ct_data->reconstructed_entry(), sct, &merkle_leaf);
   if (result != LogVerifier::VERIFY_OK) {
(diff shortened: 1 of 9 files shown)


pphaneuf:
It'd be even better if it was replaced by a `util::Status`, but one thing at a time. :wink: 

(I already had this branch lying around)

cpp/client/ct.cc
```diff
@@ -259,7 +259,7 @@ static bool VerifySCTAndPopulateSSLClientCTData(
   sct_info->mutable_sct()->CopyFrom(sct);
   const unique_ptr<LogVerifier> verifier(GetLogVerifierFromFlags());
   string merkle_leaf;
-  LogVerifier::VerifyResult result =
+  LogVerifier::LogVerifyResult result =
```
Martin2112: Is there still a name collision on LogVerifyResult or can this be shortened?

cpp/log/log_verifier.h
```diff
@@ -33,7 +33,7 @@ class LogVerifier {
     INVALID_MERKLE_PATH,
```
Martin2112: When we unify with Status we should maybe have a detail message slot in Status so we can pass the failure reason back to users.

cpp/client/ct.cc
pphaneuf: It still needs to be qualified, annoyingly... But this shouldn't live too long.

cpp/log/log_verifier.h
pphaneuf: You mean, in addition to the normal `util::Status` code and message?

Martin2112: Yes I meant adding a detail message to Status where appropriate, for
logging and potentially error responses to clients.

PR #1039, opened 2015-11-10
Proposal to split the node staleness logic out of HttpHandler

Martin2112:
This is because HttpHandler is V1 specific and we'd otherwise be duplicating this code.


Martin2112:
This PR will be reviewed at:

http://codereview.appspot.com/275890043

PR #1040, opened 2015-11-10
use pkg-config to find evhtp

philandstuff:
ellzey/libevhtp#166 (released in 1.2.11) installs a pkgconfig file in the right place which can be used to pull in headers and libs appropriately.

I found this useful when trying to build on Nixos.  Haven't yet tested in other environments but happy to test ubuntu at least.


philandstuff:
ACK on the CLA. Will get back to you.

pphaneuf:
This breaks our continuous build, so it's unlikely that we'd accept it in this form...

Two options:

 - have the call to `configure` (in `build/configure-ct`) include the necessary `evhtp_CFLAGS` and `evhtp_LIBS`
 - update to the newer version of libevhtp

philandstuff:
I think I worked out why the build was failing, and have updated the branch which (hopefully) should resolve the build errors.

benlaurie:
Hmm. You seem to have broken the CLA checking.

philandstuff:
maybe it can't cope with force pushes? I can close this and open another PR if that would help?

pphaneuf:
I think you might have changed the email address you were using in the commits?

pphaneuf:
We have forced many pushes in the past, I'm sure it can deal with those. :smiley:

philandstuff:
@pphaneuf yeah I think the email address is probably it. I'll update my user.email setting for this repo for future.

pphaneuf:
You can fix this in a few ways... First, I think you can add additional email addresses to your CLA signature?

Alternatively, you could squash your commits into a single commit, make sure your `user.email` setting is correct, and do `git commit --amend -C HEAD --reset-author`. This will, well, reset the author in the commit. :smiley:

philandstuff:
I used git filter-branch to rewrite the commits in the end.

pphaneuf:
Would it be possible to have this without upgrading evhtp? As in, modify the `configure-ct` script to add the `evhtp_CFLAGS` and `evhtp_LIBS` flags?

benlaurie:
On Thu, 19 Nov 2015 at 19:48 Philip Potter <notifications@github.com> wrote:

> I used git filter-branch to rewrite the commits in the end.
>

A corner of git-land I have not explored!

philandstuff:
@pphaneuf I'm not sure I fully understand but I've opened #1062 to do what I think you're asking for? :)

philandstuff:
Now that #1062 has been merged, I've updated this PR.  It's *much* simpler now :)

pphaneuf:
I like simple *a lot*. :smiley: 

Not sure if we'll merge that in immediately, I'd like to look at the changes in evhtp a bit first, but at least, anyone can build with what they want, easily (so you shouldn't be blocked).

Thanks!

philandstuff:
agree, happy for this to sit for a while. :+1:

PR #1041, opened 2015-11-10
Borrow Chromium's ScopedOpenSSL types.

AlCutter:
With a few mods to make it work with OpenSSL and use `unique_ptr` directly.

### DIFF
--- a/cpp/util/openssl_scoped_types.h
+++ b/cpp/util/openssl_scoped_types.h
@@ -1,0 +1,148 @@
+/* Copyright (c) 2015, Google Inc.
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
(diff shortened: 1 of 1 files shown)


cpp/util/openssl_scoped_types.h
```diff
@@ -0,0 +1,148 @@
...
+
+template<typename StackType, typename T, void (*func)(T*)>
+using ScopedOpenSSLStack =
+    std::unique_ptr<StackType, OpenSSLStackDeleter<StackType, T, func>>;
+
+template<typename T, typename CleanupRet, void (*init_func)(T*),
+         CleanupRet (*cleanup_func)(T*)>
+class ScopedOpenSSLContext {
+ public:
+  ScopedOpenSSLContext() {
```
pphaneuf: You should delete the copy and move constructors?

AlCutter: Probably a good idea :)
Done.

pphaneuf:
LGTM

PR #1046, opened 2015-11-12
Move the other periodic server tasks out to log_processes.cc

Martin2112:
Dedup more server code and cleanup a lot of includes and using statements that are no longer used.


pphaneuf:
LGTM

I was going to say the flag was put in the wrong place, but the next commit fixes it, so all good. :smiley:

PR #1047, opened 2015-11-12
Add the infrastructure for v2 versions of ct-server and ct-mirror 

Martin2112:
This registers v2 request handlers but they reject all requests with HTTP_NOTIMPLEMENTED.

At the moment there seems to be no good alternative than duplicating the handlers as it's not yet clear how much of the code will remain in common with v1. Hopefully, there will be some common code that can be refactored out.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,7 +32,9 @@ cpp/version.o: \
 bin_PROGRAMS = \
 	cpp/client/ct \
 	cpp/server/ct-mirror \
+	cpp/server/ct-mirror-v2 \
 	cpp/server/ct-server \
+	cpp/server/ct-server-v2 \
 	cpp/server/xjson-server \
 	cpp/tools/ct-clustertool
 
@@ -213,10 +215,24 @@ cpp_server_ct_mirror_LDADD = \
 	$(leveldb_LIBS) \
 	-lprotobuf -lsqlite3
 cpp_server_ct_mirror_SOURCES = \
-  cpp/client/async_log_client.cc \
+	cpp/client/async_log_client.cc \
 	cpp/server/ct-mirror.cc \
 	cpp/server/certificate_handler.cc \
 	cpp/server/handler.cc \
+	cpp/server/json_output.cc \
+	cpp/server/server_helper.cc
+
+cpp_server_ct_mirror_v2_LDADD = \
+	cpp/libcore.a \
+	$(json_c_LIBS) \
+	$(libevent_LIBS) \
+	$(leveldb_LIBS) \
+	-lprotobuf -lsqlite3
+cpp_server_ct_mirror_v2_SOURCES = \
+	cpp/client/async_log_client.cc \
+	cpp/server/ct-mirror_v2.cc \
+	cpp/server/certificate_handler_v2.cc \
+	cpp/server/handler_v2.cc \
 	cpp/server/json_output.cc \
 	cpp/server/server_helper.cc
 
@@ -227,12 +243,27 @@ cpp_server_ct_server_LDADD = \
 	$(leveldb_LIBS) \
 	-lprotobuf -lsqlite3
 cpp_server_ct_server_SOURCES = \
-  cpp/client/async_log_client.cc \
+	cpp/client/async_log_client.cc \
 	cpp/server/ct-server.cc \
 	cpp/server/certificate_handler.cc \
 	cpp/server/handler.cc \
 	cpp/server/json_output.cc \
-        cpp/server/log_processes.cc \
+	cpp/server/log_processes.cc \
+	cpp/server/server_helper.cc
+
+cpp_server_ct_server_v2_LDADD = \
+	cpp/libcore.a \
+	$(json_c_LIBS) \
+	$(libevent_LIBS) \
+	$(leveldb_LIBS) \
+	-lprotobuf -lsqlite3
+cpp_server_ct_server_v2_SOURCES = \
+	cpp/client/async_log_client.cc \
+	cpp/server/ct-server_v2.cc \
+	cpp/server/certificate_handler_v2.cc \
+	cpp/server/handler_v2.cc \
+	cpp/server/json_output.cc \
+	cpp/server/log_processes.cc \
 	cpp/server/server_helper.cc
 
 cpp_server_xjson_server_LDADD = \
@@ -242,10 +273,10 @@ cpp_server_xjson_server_LDADD = \
 	$(leveldb_LIBS) \
 	-lprotobuf -lsqlite3
 cpp_server_xjson_server_SOURCES = \
-  cpp/client/async_log_client.cc \
+	cpp/client/async_log_client.cc \
 	cpp/server/handler.cc \
 	cpp/server/json_output.cc \
-        cpp/server/log_processes.cc \
+	cpp/server/log_processes.cc \
 	cpp/server/server_helper.cc \
 	cpp/server/xjson-server.cc \
 	cpp/server/x_json_handler.cc \
--- a/cpp/server/certificate_handler_v2.h
+++ b/cpp/server/certificate_handler_v2.h
@@ -32,7 +32,6 @@ class CertificateHttpHandlerV2 : public HttpHandlerV2 {
   const CertSubmissionHandler submission_handler_;
   Frontend* const frontend_;
 
-  void GetEntries(evhttp_request* req) const;
   void GetRoots(evhttp_request* req) const;
   void AddChain(evhttp_request* req);
   void AddPreChain(evhttp_request* req);
--- a/cpp/server/certificate_handler_v2.h
+++ b/cpp/server/certificate_handler_v2.h
@@ -1,0 +1,52 @@
+#ifndef CERT_TRANS_SERVER_CERTIFICATE_HANDLER_V2_H_
+#define CERT_TRANS_SERVER_CERTIFICATE_HANDLER_V2_H_
+
+#include "log/cert_submission_handler.h"
+#include "log/database.h"
+#include "log/logged_entry.h"
+#include "server/handler_v2.h"
+#include "server/staleness_tracker.h"
+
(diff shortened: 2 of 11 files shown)


Makefile.am
```diff
@@ -232,7 +248,22 @@ cpp_server_ct_server_SOURCES = \
 	cpp/server/certificate_handler.cc \
 	cpp/server/handler.cc \
 	cpp/server/json_output.cc \
-        cpp/server/log_processes.cc \
+  cpp/server/log_processes.cc \
```
pphaneuf: The indentation seems to be a bit strange, looking at it on GitHub... I'm suspecting spaces-vs-tabs? Remember that leading tabs are significant for `make`!

Makefile.am
```diff
@@ -220,6 +222,20 @@ cpp_server_ct_mirror_SOURCES = \
...
 	cpp/server/server_helper.cc
 
+cpp_server_ct_mirror_v2_LDADD = \
+	cpp/libcore.a \
+	$(json_c_LIBS) \
+	$(libevent_LIBS) \
+	$(leveldb_LIBS) \
+	-lprotobuf -lsqlite3
+cpp_server_ct_mirror_v2_SOURCES = \
+  cpp/client/async_log_client.cc \
```
pphaneuf: Same here.

Makefile.am
Martin2112: Wasn't tabs, for some reason it's not displaying correctly in my editor. It worked in vi so I'm about to push a fix.

cpp/server/handler_v2.h
```diff
@@ -0,0 +1,85 @@
...
+  bool IsNodeStale() const;
+  void UpdateNodeStaleness();
+
+  LogLookup* const log_lookup_;
+  const ReadOnlyDatabase* const db_;
+  const ClusterStateController<LoggedEntry>* const controller_;
+  Proxy* proxy_;
+  ThreadPool* const pool_;
+  libevent::Base* const event_base_;
+  StalenessTracker* staleness_tracker_;
```
pphaneuf: This pointer should be `const`, like (most of) the others.

cpp/server/handler_v2.cc
```diff
@@ -0,0 +1,185 @@
...
+
+void HttpHandlerV2::BlockingGetEntries(evhttp_request* req, int64_t start,
+                                     int64_t end, bool include_scts) const {
+  return SendJsonError(event_base_, req, HTTP_NOTIMPLEMENTED,
+                       "Not yet implemented.");
+}
+
+
+bool HttpHandlerV2::IsNodeStale() const {
+  return staleness_tracker_->IsNodeStale();
```
pphaneuf: I think we could do without this method. :smiley:

cpp/server/handler_v2.cc
```diff
@@ -0,0 +1,185 @@
...
+}
+
+
+void HttpHandlerV2::UpdateNodeStaleness() {
+  staleness_tracker_->UpdateNodeStaleness();
```
pphaneuf: And this one too (which isn't even called, anyway?).

cpp/server/handler_v2.h
Martin2112: Who wrote the v1 version then?

cpp/server/handler_v2.cc
Martin2112: It was in HttpHandler before I refactored it to avoid duplicating the code. Maybe it was never being used.

cpp/server/handler_v2.h
pphaneuf: You did? :smiley: 

https://github.com/google/certificate-transparency/commit/3a33dd564560c317dc39b16ea99c392154b98fd8#diff-c7c0fdfd5ebf9c348a1a70cffcb69bc8R77

Martin2112: OK I now see it only applies to the one I added. I'll go back to my corner.

Makefile.am
```diff
@@ -213,13 +215,27 @@ cpp_server_ct_mirror_LDADD = \
...
 	cpp/server/server_helper.cc
 
+cpp_server_ct_mirror_v2_LDADD = \
+	cpp/libcore.a \
+	$(json_c_LIBS) \
+	$(libevent_LIBS) \
+	$(leveldb_LIBS) \
+	-lprotobuf -lsqlite3
+cpp_server_ct_mirror_v2_SOURCES = \
+        cpp/client/async_log_client.cc \
```
pphaneuf: This one is still odd?

Makefile.am
pphaneuf: Still this one (but it looks like it's improving!).

Makefile.am
```diff
@@ -232,7 +248,22 @@ cpp_server_ct_server_SOURCES = \
...
+	cpp/server/server_helper.cc
+
+cpp_server_ct_server_v2_LDADD = \
+	cpp/libcore.a \
+	$(json_c_LIBS) \
+	$(libevent_LIBS) \
+	$(leveldb_LIBS) \
+	-lprotobuf -lsqlite3
+cpp_server_ct_server_v2_SOURCES = \
+  cpp/client/async_log_client.cc \
```
pphaneuf: And this.

Makefile.am
```diff
@@ -232,7 +248,22 @@ cpp_server_ct_server_SOURCES = \
...
+	cpp/server/ct-server_v2.cc \
+	cpp/server/certificate_handler_v2.cc \
+	cpp/server/handler_v2.cc \
+	cpp/server/json_output.cc \
+  cpp/server/log_processes.cc \
```
pphaneuf: This.

Makefile.am
```diff
@@ -245,7 +276,7 @@ cpp_server_xjson_server_SOURCES = \
   cpp/client/async_log_client.cc \
 	cpp/server/handler.cc \
 	cpp/server/json_output.cc \
-        cpp/server/log_processes.cc \
+  cpp/server/log_processes.cc \
```
pphaneuf: This (finally!).

cpp/server/handler_v2.cc
pphaneuf: When I embark on a refactoring, I often do a "pre-refactoring" of trimming off any dead code. Non-existent code being the easiest to refactor... :wink:

Martin2112: Where's the fun in that?

Makefile.am
pphaneuf: I give up, just look at it on GitHub, it's all fairly clear there, somehow? `cpp_server_xjson_server_SOURCES` below looks totally wacky before this PR, even (your one change to it seems to be for the better, but the other lines are all randomly indented?).

Good luck. :smiley:

Martin2112: OK now I fixed it.

cpp/server/certificate_handler_v2.h
```diff
@@ -0,0 +1,52 @@
...
+
+ protected:
+  void AddHandlers(libevent::HttpServer* server) override;
+
+ private:
+  const CertChecker* const cert_checker_;
+  const CertSubmissionHandler submission_handler_;
+  Frontend* const frontend_;
+
+  void GetEntries(evhttp_request* req) const;
```
pphaneuf: This method doesn't exist (and shouldn't exist, it's spurious in `CertificateHttpHandler` as well).

Martin2112: GetEntries? It's bound in HttpHandler.

pphaneuf: The non-virtual `HttpHandlerV2::GetEntries` is bound there. Since it isn't virtual, `CertificateHttpHandlerV2::GetEntries` never comes into play, which is a good thing, since it isn't defined anywhere (so if you managed to call it, the binary wouldn't link). :wink:

cpp/server/handler_v2.cc
```diff
@@ -0,0 +1,175 @@
...
+  AddProxyWrappedHandler(server, "/ct/v2/get-sth-consistency",
+                         bind(&HttpHandlerV2::GetConsistency, this, _1));
+
+  // Now add any sub-class handlers.
+  AddHandlers(server);
+}
+
+
+void HttpHandlerV2::SetProxy(Proxy* proxy) {
+  LOG_IF(FATAL, proxy_) << "Attempting to re-add a HttpHandlerV2.";
```
pphaneuf: Re-add a `HttpHandlerV2`? Re-add a proxy, right?

pphaneuf:
Done with the review, and other than that log message and spurious `GetEntries` declaration, it's nice, but...

... it might be make your upcoming PRs easier if you collapsed the two handlers in one (maybe keep the `CertificateHttpHandlerV2` name, to remind ourselves that it's specific to certificates)? Would avoid some going back and forth? Once we get to do non-certificate variants of the V2 protocol, we can split a working handler back out easily...

But that's your call, just an idea (and it depends on how you plan to go at this, anyway), I'm fine with this too.

cpp/server/certificate_handler_v2.h
Martin2112: Removed.

cpp/server/handler_v2.cc
Martin2112: I reckon so, was copied from the old class. Fixed.

Martin2112:
I don't know whether to reorganize this yet as I don't know what the result is going to look like. Some of the handlers will probably be identical but will serve V2 objects whereas others like precert submission will be very different.

pphaneuf:
Ah, potentially sharing a base class between V1 and V2 for most of the methods, once you see how things turn out? Sounds good, I like the approach of doing it with some duplicate and refactoring the commonalities out once it's done (since we're not 100% sure), rather than refactoring everything, only to find out it possibly doesn't share much.

Taking another look.

pphaneuf:
LGTM

PR #1048, opened 2015-11-13
Add BIGNUM wrapper.

AlCutter:


cpp/util/bignum.cc
```diff
@@ -0,0 +1,35 @@
...
+
+BigNum::~BigNum() {
+  BN_free(&bn_);
+}
+
+
+inline const BigNum& AsBigNum(const BigNum& n) {
+  return n;
+}
+
```
phad: No double newline here

cpp/util/bignum.h
```diff
@@ -0,0 +1,162 @@
+#ifndef CERT_TRANS_UTIL_BIGNUM_H_
+#define CERT_TRANS_UTIL_BIGNUM_H_
+
+#include <glog/logging.h>
```
phad: Don't think you need this in your .h file!

cpp/util/bignum.h
```diff
@@ -0,0 +1,162 @@
...
+
+
+class BigNum {
+ public:
+  BigNum();
+  explicit BigNum(int64_t w);
+  BigNum(const BigNum& other);
+  ~BigNum();
+
+  BIGNUM* bn();
```
phad: Can you do without this exposing of privates?

cpp/util/bignum.h
phad: This is for CHECK I guess

cpp/util/bignum.h
```diff
@@ -0,0 +1,162 @@
...
+
+ private:
+  BIGNUM bn_;
+
+  friend bool operator==(const BigNum& a, const BigNum& b);
+  friend bool operator==(const BigNum& a, int64_t b);
+  friend bool operator<(const BigNum& a, const BigNum& b);
+  friend bool operator<(const BigNum& a, int64_t b);
+  friend bool operator>(const BigNum& a, const BigNum& b);
+  friend bool operator>(const BigNum& a, int64_t b);
```
pphaneuf: Am I correct in thinking that those operators are *exactly* the ones you need?

cpp/util/bignum.h
```diff
@@ -0,0 +1,162 @@
...
+  return r;
+}
+
+
+inline bool operator==(const BigNum& a, const BigNum& b) {
+  return BN_cmp(&a.bn_, &b.bn_) == 0;
+}
+
+
+namespace {
```
phad: Not sure of the utility of doing this in a .h file

cpp/util/bignum.h
```diff
@@ -0,0 +1,162 @@
...
+}
+
+
+template <typename T, typename U>
+inline bool operator<(const T& a, const U& b) {
+  return AsBigNum(a) < AsBigNum(b);
+}
+
+
+inline bool operator>(const BigNum& a, const BigNum& b) {
```
phad: Any need for operator<= or operator >= ?

cpp/util/bignum.h
pphaneuf: Or maybe make it `const` (return value and method), at least? But I have a feeling this is to pass into some OpenSSL API that should have a `const BIGNUM*`, but forgot the `const`, so we're all going to cry...

cpp/util/bignum_test.cc
```diff
@@ -0,0 +1,194 @@
...
+
+#include <gtest/gtest.h>
+
+#include "util/testing.h"
+
+
+namespace cert_trans {
+
+
+class BigNumTest : public ::testing::Test {};
```
phad: This seems redundant.  All the TEST_F macros might as well be TEST, no?

cpp/util/bignum_test.cc
```diff
@@ -0,0 +1,194 @@
...
+
+
+TEST_F(BigNumTest, TestDefaultCtor) {
+  BigNum n;
+  EXPECT_EQ(n, 0);
+}
+
+
+TEST_F(BigNumTest, TestWordCtor) {
+  const int kValue(0x1234);
```
phad: Why int, and not int64_t here and below?

cpp/util/bignum.cc
AlCutter: Done

cpp/util/bignum.h
AlCutter: Yep

cpp/util/bignum.h
AlCutter: This was an escape hatch to still be able to use any `libcrypto` routines (`BN_` or otherwise) which takes a `BIGNUM*`). 

To make your day(s), I've added MOAR bn()!11oneeleventy

cpp/util/bignum.h
AlCutter: More or less, yes

cpp/util/bignum.h
AlCutter: Good point, renamed the namespace to `internal`

cpp/util/bignum.h
AlCutter: Not at the moment...

cpp/util/bignum_test.cc
AlCutter: Good call, done.

cpp/util/bignum_test.cc
AlCutter: Just 'cos I didn't really care about the size and I knew I was going to be boring and pick small test values (resting my weary fingers etc. etc.), can change them if you like?

cpp/util/bignum.h
phad: Can you do a lighter-weight include of something to provide just asserts?

AlCutter: I could just use `assert()` instead I guess

PR #1049, opened 2015-11-16
clang-tidy fix

AlCutter:
Just the output of clang-tidy over all the `.cc` and `.h` files.


Martin2112:
This week we have a special on merge conflicts ... I suppose it's got to be done!

cpp/fetcher/fetcher.cc
```diff
@@ -146,7 +146,7 @@ void FetchState::WalkEntries() {
 
   int64_t index(start_);
   int num_fetch(0);
-  for (Range* current = entries_.get(); current;
+  for (Range *current = entries_.get(); current;
```
Martin2112: I thought our style was Range* ?

cpp/log/cert.cc
```diff
@@ -583,8 +583,8 @@ util::Status Cert::OctetStringExtensionData(int extension_nid,
   // Callers don't care whether extension is missing or invalid as they
   // usually call this method after confirming it to be present.
   const StatusOr<void*> ext_struct = ExtensionStructure(extension_nid);
-  if (!ext_struct.ok()
-      && ext_struct.status().CanonicalCode() == Code::NOT_FOUND) {
+  if (!ext_struct.ok() &&
```
Martin2112: Is that right? I thought we broke on operators, not after.

cpp/fetcher/fetcher.cc
AlCutter: It is :)
Weirdly, if we move `Range* current` out of the  `for` loop then it works as expected.
Bug in clang-format maybe?

cpp/log/cert.cc
AlCutter: Don't think we had an opinion on this...
It's always been:
1) Google C++ Style Guide (which, AFAIR doesn't talk about exactly where to break)
2) then tie-break by falling back to what clang-format does.

cpp/client/ssl_client.cc
```diff
@@ -189,7 +189,8 @@ int SSLClient::VerifyCallback(X509_STORE_CTX* ctx, void* arg) {
                 cert_trans::NID_ctSignedCertificateTimestampList)
           : util::Status::UNKNOWN;
 
-  // First check for embedded proof, if not present then look for the proof in a
+  // First check for embedded proof, if not present then look for the proof in
+  // a
```
pphaneuf: Some awful here.

cpp/fetcher/remote_peer.cc
```diff
@@ -83,7 +83,8 @@ void RemotePeer::Impl::DoneGetSTH(
     const LogVerifier::LogVerifyResult result(verifier_->VerifySignedTreeHead(
         *new_sth, 0, (time(NULL) + 10) * kNumMillisPerSecond));
 
-    // TODO(alcutter): We should probably log these invalid STHs somewhere a bit
+    // TODO(alcutter): We should probably log these invalid STHs somewhere a
+    // bit
```
pphaneuf: More awful.

cpp/log/cert.h
```diff
@@ -164,9 +164,11 @@ class Cert {
   // you're doing.
   //
   // Returns OK if the extension data could be fetched and decoded.
-  // Returns NOT_FOUND if the extension is not present, or is present but is not
+  // Returns NOT_FOUND if the extension is not present, or is present but is
+  // not
```
pphaneuf: Fix all the awful in this file.

cpp/log/ct_extensions.cc
```diff
@@ -129,10 +129,10 @@ static X509V3_EXT_METHOD ct_poison_method = {
...
 // DNS id in the cert. See RFC section 3.2.2.
 static X509V3_EXT_METHOD ct_redaction_count_method = {
-    0,                         // ext_nid, NID, will be created by OBJ_create()
-    0,                         // flags
-    ASN1_ITEM_ref(REDACTED_LABEL_COUNT),
-    0, 0, 0, 0,                // ignored since the above is set
+    0,  // ext_nid, NID, will be created by OBJ_create()
+    0,  // flags
+    ASN1_ITEM_ref(REDACTED_LABEL_COUNT), 0, 0, 0,
+    0,  // ignored since the above is set
```
pphaneuf: This might be slightly wrong now?

cpp/log/ct_extensions.h
```diff
@@ -10,12 +10,12 @@ void LoadCtExtensions();
 
 // Defines structure for redacted label count to hold a stack of integers
 typedef struct RedactedLabelCount_st {
-  STACK_OF(ASN1_INTEGER)* redacted_labels;
+  STACK_OF(ASN1_INTEGER) * redacted_labels;
```
pphaneuf: Eeek! :smiley:

cpp/log/cert.cc
Martin2112: OK. AFAIK the Google Java style guide does specify break before operator so it's what I'm used to seeing.

cpp/log/ct_extensions.cc
Martin2112: Yes, those 4 values related to that comment.

cpp/log/ct_extensions.h
AlCutter: I guess we should look at replacing a lot of this with the new scoped ssl pointers and, if we're lucky, avoid this mess that way :)

I'll leave this as-is for now though...

cpp/log/ct_extensions.cc
AlCutter: I've reverted the formatting on those blocks and slapped clang-format's wrist.

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -813,8 +812,8 @@ util::StatusOr<int64_t> EtcdConsistentStore<Logged>::CleanupOldEntries() {
     LOG(INFO) << "No current serving_sth, nothing to do.";
     return 0;
   }
-  const int64_t clean_up_to_sequence_number(
-      serving_sth_->Entry().tree_size() - 1);
+  const int64_t clean_up_to_sequence_number(serving_sth_->Entry().tree_size() -
```
Martin2112: That looks weird?

cpp/log/cert.h
AlCutter: less awful'd

cpp/log/tree_signer-inl.h
```diff
@@ -115,10 +115,10 @@ util::Status TreeSigner<Logged>::SequenceNewEntries() {
   for (const auto& m : mapping.Entry().mapping()) {
     // Go home clang-format, you're drunk.
     CHECK(
-        sequenced_hashes.insert(
-                             std::make_pair(m.entry_hash(),
-                                            std::make_pair(m.sequence_number(),
-                                                           false))).second);
+        sequenced_hashes
```
Martin2112: I think that's been made less readable by the wonders of automation.

cpp/util/etcd.cc
```diff
@@ -997,9 +1011,9 @@ void EtcdClient::MaybeLogEtcdVersion() {
   }
   logged_version_ = true;
 
-  const UrlFetcher::Request req(
-      URL("http://" + etcds_.front().first + ":" +
-              to_string(etcds_.front().second) + "/version"));
+  const UrlFetcher::Request req(URL("http://" + etcds_.front().first + ":" +
```
Martin2112: It seems to have done the opposite with this than with some of the other changes.

AlCutter:
Some things it's going to get "wrong" because it's not human, and while there is a vast array of knobs and levers you can twiddle with to try to get it less wrong, in practice all that happens is that you get the bit you're currently looking at "right" and something else goes wonky instead (but don't let that discourage you from having a pop at it if you're so minded ;)

Ultimately, though, it's better to have a tool which is the final arbiter on format, even if it's not 100% perfect.

cpp/util/etcd.cc
```diff
@@ -41,9 +41,11 @@ using util::Task;
 
 DEFINE_int32(etcd_watch_error_retry_delay_seconds, 5,
              "delay between retrying etcd watch requests");
-DEFINE_bool(etcd_consistent, true, "Add consistent=true param to all requests. "
+DEFINE_bool(etcd_consistent, true,
+            "Add consistent=true param to all requests. "
```
pphaneuf: Should put back the doc string in a single string, and let it re-split here?

cpp/util/etcd.cc
```diff
@@ -41,9 +41,11 @@ using util::Task;
...
             "Do not turn this off unless you *know* what you're doing.");
-DEFINE_bool(etcd_quorum, true, "Add quorum=true param to all requests. "
+DEFINE_bool(etcd_quorum, true,
+            "Add quorum=true param to all requests. "
```
pphaneuf: Same.

cpp/log/ct_extensions.h
pphaneuf: SGTM

pphaneuf:
Done with the review.

pphaneuf:
LGTM

Martin2112:
LGTM, within reasonable doubt.

cpp/log/cert.cc
benlaurie: I prefer it the old way. Apparently we can ask clang-format to do:

BreakBeforeBinaryOperators: BOS_NonAssignment

pphaneuf: I prefer the old way, but only by a small margin (with old one, operator is more obvious, with new one, nicer alignment), so I didn't might it too much. If there's a `clang-format` option to do it, sure, why not.

PR #1051, opened 2015-11-17
go/x509: Add DisableTimeChecks option to verifier

katjoyce:
DisableTimeChecks option to be used in fix-chain code.

### DIFF
--- a/go/x509/verify.go
+++ b/go/x509/verify.go
@@ -126,10 +126,11 @@ func (e SystemRootsError) Error() string {
 // VerifyOptions contains parameters for Certificate.Verify. It's a structure
 // because other PKIX verification APIs have ended up needing many options.
 type VerifyOptions struct {
-	DNSName       string
-	Intermediates *CertPool
-	Roots         *CertPool // if nil, the system roots are used
-	CurrentTime   time.Time // if zero, the current time is used
+	DNSName           string
+	Intermediates     *CertPool
+	Roots             *CertPool // if nil, the system roots are used
+	CurrentTime       time.Time // if zero, the current time is used
+	DisableTimeChecks bool
 	// KeyUsage specifies which Extended Key Usage values are acceptable.
 	// An empty list means ExtKeyUsageServerAuth. Key usage is considered a
 	// constraint down the chain which mirrors Windows CryptoAPI behaviour,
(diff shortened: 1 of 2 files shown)


AlCutter:
Could you add a simple test or two for this in `go/x509/verify_test.go` too?

go/x509/verify.go
```diff
@@ -145,12 +146,14 @@ const (
...
-	now := opts.CurrentTime
-	if now.IsZero() {
-		now = time.Now()
-	}
-	if now.Before(c.NotBefore) || now.After(c.NotAfter) {
-		return CertificateInvalidError{c, Expired}
+	if !opts.DisableTimeChecks {
+		now := opts.CurrentTime
+		if now.IsZero() {
+			now = time.Now()
```
eranmes: Nit: Seems that time.Now() returns the current local, are the NotBefore/NotAfter values in the certificate converted to local time? (otherwise they'd be UTCTime, https://tools.ietf.org/html/rfc5280#section-4.1.2.5)

benlaurie: Hmm. That would be a bug in the upstream x509 code!

AlCutter: Might not be a terrible idea to update our fork inline with the new upstream version...

eranmes: I've confirmed with agl this is an unfixed bug upstream. I intend to send a patch upstream.

go/x509/verify_test.go
```diff
@@ -83,6 +84,30 @@ var verifyTests = []verifyTest{
...
+	{
+		leaf:              googleLeaf,
+		intermediates:     []string{thawteIntermediate},
+		roots:             []string{verisignRoot},
+		currentTime:       10000000000,
+		dnsName:           "www.google.com",
+		disableTimeChecks: true,
+
+		expectedChains: [][]string{
+			{"Google", "Thawte", "VeriSign"},
```
benlaurie: Should also test that disableTimeChecks: false leads to errors?

katjoyce: disableTimeChecks defaults to false, so every test that does not specify it to be true is testing it set to false.  I believe there is one test that checks for an expiration error if currentTime = 1, but you're right that it could also do with a similar test for currentTime set to way in the future.

benlaurie: What I mean is a check along the lines of "expired/not-yet-valid cert with disableTimeChecks: false, fails because of validity period" + "same cert with disableTimeChecks: true" passes.

katjoyce: As I understand it, the expired/not-yet-valid cert is simulated by changing the value of currentTime so that the cert appears to be expired/not-yet-valid.  With this in mind I have just added a test and restructured them so that it is clear that a test that failed because of its validity period then passes if disableTimeChecks is true, for a cert that is expired, and a cert that is not-yet-valid against the currentTime.

PR #1052, opened 2015-11-17
Separate out CMS scoped types as they may be something other than OpenSSL CMS in the future

AlCutter:
/cc @katjoyce


pphaneuf:
LGTM

Issue #1053, opened 2015-11-18
SEGV when --etcd_servers receives bad input


philandstuff:
If I pass an `--etcd_servers` value of `"localhost:4001"` (ie literal double quotes, after handling by the shell), I get a segfault:

```
$ ct-server --trusted_cert_file trusted.pem --key private_key.pem --leveldb_db=log.ldb --port 8080 --server localhost --i_know_stand_alone_mode_can_lose_data --etcd_servers \"localhost:4001\"
*** Aborted at 1447840403 (unix time) try "date -d @1447840403" if you are using GNU date ***
PC: @     0x7fdd705bce90 (unknown)
*** SIGSEGV (@0x20) received by PID 20528 (TID 0x7fdd70ffc780) from PID 32; stack trace: ***
    @     0x7fdd6ec253e0 (unknown)
    @     0x7fdd705bce90 (unknown)
    @     0x55ca074b2e37 cert_trans::URL::URL()
    @     0x55ca074c7ee1 cert_trans::EtcdClient::MaybeLogEtcdVersion()
    @     0x55ca074c8447 cert_trans::EtcdClient::Generic()
    @     0x55ca074c83b3 cert_trans::EtcdClient::Get()
    @     0x55ca074c65f3 cert_trans::EtcdClient::WatchRequestDone()
    @     0x55ca074ca0da cert_trans::EtcdClient::Watch()
    @     0x55ca07545d4b cert_trans::EtcdConsistentStore<>::WatchServingSTH()
    @     0x55ca07541408 cert_trans::EtcdConsistentStore<>::EtcdConsistentStore()
    @     0x55ca07541510 cert_trans::EtcdConsistentStore<>::EtcdConsistentStore()
    @     0x55ca074bf203 cert_trans::Server::Server()
    @     0x55ca0743f59c main
    @     0x7fdd6de5a995 __libc_start_main
    @     0x55ca0742cf09 _start
Segmentation fault
```
It would be nicer to have a friendly validation error message.

(How did I discover this?  I was writing a systemd service for ct-server; I misunderstood how systemd handled quotes and so hadn't realized they were being passed straight through.  I found it thoroughly confusing that the command line worked fine from bash (as it was interpreting the quotes) but the "same" command line from systemd caused a segfault.)

Martin2112:
Thanks for reporting this. There should be a proper error message for this now.

pphaneuf:
I'll close it, it should be fixed now, re-open if that's not the case.

[...]

PR #1054, opened 2015-11-18
Additional validation on etcd_servers flag.

Martin2112:
Ensure each element can be parsed as a URL when it is split up from the flag value. Add tests. Addresses #1053.


pphaneuf:
There was a segfault in that bug report, where is that coming from? I'm guessing from here?

https://github.com/google/certificate-transparency/blob/master/cpp/net/url.cc#L29

Maybe add a `CHECK_NOTNULL` over there, to have a slightly less crude crash (we use that `URL` class in a few different places).

Martin2112:
Yes, the value was passed directly to URL, which expects the result to be not null. I'll add a check.

Martin2112:
Done.

pphaneuf:
LGTM

PR #1057, opened 2015-11-18
Put scoped types into the cert_trans namespace.

katjoyce:


cpp/client/ct.cc
```diff
@@ -169,7 +169,7 @@ static LogVerifier* GetLogVerifierFromFlags() {
 static void AddOctetExtension(X509* cert, int nid, const unsigned char* data,
                               int data_len, int critical) {
   // The extension as a single octet string.
-  ScopedASN1_OCTET_STRING inner(ASN1_OCTET_STRING_new());
+  util::ScopedASN1_OCTET_STRING inner(ASN1_OCTET_STRING_new());
```
pphaneuf: You could put some `using` statements at the top (starting at line 127, keep them sorted), instead of changing it all over the file.

cpp/util/openssl_scoped_types.h
```diff
@@ -44,6 +44,7 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 
+namespace util {
```
pphaneuf: I reckon that if you put this in the `cert_trans` namespace, it would both make sense, and not need qualifying in most places. :wink:

cpp/client/ct.cc
katjoyce: Done.

cpp/util/openssl_scoped_types.h
katjoyce: cert_trans namespace instead of util namespace?

pphaneuf: Exactly. Then, all the things that are inside a `namespace cert_trans` won't need the qualification (for one example, the `Signer` class, but I'm sure most others as well). This might possibly reduce the number of affected files almost all the way to 2-3!

pphaneuf: Would still leave `ssl_client.h`, `test_signer.cc`, `sparse_merkle_tree_test.cc` needing some `using` or qualifiers, in addition to the three header files, so 6 files affected.

katjoyce: Done

pphaneuf:
LGTM

PR #1060, opened 2015-11-19
Fix a few govet warnings.

pphaneuf:


pphaneuf:
Emacs also wants me to remove `"github.com/google/certificate-transparency/go"` from `logclient_test.go`? I'm not sure why it wants to do that, though, I ran `gofmt -s` and `govet`, didn't complain about that...

AlCutter:
It's because (I suspect) that Emacs is incorrectly assuming that the package you get from importing `github.com/google/certificate-transparency/go` will be called `go` rather than `ct` which it actually is.

pphaneuf:
Makes sense, although it *is* slightly odd that our name doesn't match... :smiley:

PR #1062, opened 2015-11-23
be more specific when searching for evhtp

philandstuff:
Rather than setting blanket CFLAGS and LIBS that add evhtp to every
single compile or link command, add specific evhtp_CFLAGS and evhtp_LIBS
variables and include them only where needed.

This prepares the way for a smooth migration to use pkg-config to find
libevhtp (available in 1.2.11 but gclient currently pulls in 1.2.10).


build/configure-ct
```diff
@@ -39,5 +39,5 @@ then
       ;;
   esac
 
-  ./configure GMOCK_DIR=${GMOCK} ${CONFIGURE_FLAGS} PROTOC=${INSTALL_DIR}/bin/protoc PKG_CONFIG_PATH="${INSTALL_DIR}/lib/pkgconfig" CPPFLAGS="-I${INSTALL_DIR}/include" CXXFLAGS="${CXXFLAGS} ${EXTRA_CXXFLAGS}" LDFLAGS="-L${INSTALL_DIR}/lib $EXTRA_LDFLAGS" LIBS="${EXTRA_LIBS}" INSTALL_DIR=${INSTALL_DIR} || tail -1000 config.log
+  ./configure GMOCK_DIR=${GMOCK} ${CONFIGURE_FLAGS} PROTOC=${INSTALL_DIR}/bin/protoc PKG_CONFIG_PATH="${INSTALL_DIR}/lib/pkgconfig" CPPFLAGS="-I${INSTALL_DIR}/include" CXXFLAGS="${CXXFLAGS} ${EXTRA_CXXFLAGS}" LDFLAGS="-L${INSTALL_DIR}/lib $EXTRA_LDFLAGS" LIBS="${EXTRA_LIBS}" evhtp_CFLAGS="-I${INSTALL_DIR}/include/evhtp" evhtp_LIBS="-levhtp" INSTALL_DIR=${INSTALL_DIR} || tail -1000 config.log
```
pphaneuf: One thing here: I believe that the `evhtp_CFLAGS` is wrong. The `CPPFLAGS` used to have only `-I${INSTALL_DIR}/include` (and still do). Is there even anything in `${INSTALL_DIR}/include/evhtp`?

pphaneuf: You might need to still provide *something* for `evhtp_CFLAGS`, even if only a redundant `-I${INSTALL_DIR}/include`, I think `PKG_CHECK_MODULES` gives an error if both are not set?

philandstuff: This is the value that pkg-config would have provided.  You can see here that pkg-config's cflags for evhtp will *always* end in the string `/evhtp`: https://github.com/ellzey/libevhtp/blob/develop/evhtp.pc.in#L3

Now, I don't know whether that directory is actually needed or not; you might well be right that this isn't the right value to pull the headers in.  But it is the value that pkg-config will provide in evhtp 1.2.11 and I don't see any reason to deviate from that value.

If evhtp's pkg-config is returning wrong values, maybe we need to submit something upstream to them?

pphaneuf: Heh, we might need to submit something upstream...

In the meantime, could you add a commit to this PR (separate, so we can remove it easily if it's wrong) that removes the `/evhtp` there, and see how that goes on Travis?

pphaneuf: Ok, I looked at this closer...

https://github.com/ellzey/libevhtp/commit/7065156a31a37e3b69f1d04d873ac34af87acc3c

In 1.2.10 (which we're still using, as of this PR), the headers are in `${INSTALL_DIR}/include`, they've changed the location in 1.2.11.

What I would say is this: right now, don't preemptively put the 1.2.11 location, but make it what needs to compile (which might be even `evhtp_CFLAGS=""`!), and when we do upgrade to 1.2.11, we'll switch away from specifying `evhtp_CFLAGS` and `evhtp_LIBS` and let the `configure` script get them from `pkg-config`, which will then "just work".

philandstuff: Top sleuthing! I'll add this right away.

philandstuff:
@pphaneuf updated, just waiting for travis now

philandstuff:
updated again. The empty string didn't work so I went with the generic include dir as an argument

benlaurie:
Bit confused by this discussion - shouldn't packages that use pkg-config have their flags provided by pkg-config, not by us?

philandstuff:
@benlaurie the issue is that libevhtp 1.2.10 does not install pkg-config information. 1.2.11 does (which is what #1040 uses) but @pphaneuf has [asked to separate bumping the libevhtp dependency version from updating the Makefile.am to use evhtp_LIBS flag](https://github.com/google/certificate-transparency/pull/1040#issuecomment-158365829).  In order to do that, configure-ct needs to pass in the evhtp_CFLAGS and evhtp_LIBS values.

pphaneuf:
Doing it this way, we'll *know* that it works with either. `pkg-config` is supposed to help, but can be told what to do, and I wanted to make sure that the "tell it what to do" option worked fine.

pphaneuf:
LGTM, merging.

philandstuff:
:fireworks:

PR #1063, opened 2015-11-23
Use BigNum in SparseMerkleTree

AlCutter:


cpp/merkletree/sparse_merkle_tree.cc
```diff
@@ -188,21 +193,20 @@ string SparseMerkleTree::TreeNode::DebugString() const {
...
-  for (size_t i(0); i < path.size(); ++i) {
-    uint8_t t(path[i]);
+  const string bytes(BytesFromPath(path));
+  for (auto it(bytes.rbegin()); it != bytes.rend(); ++it) {
+    uint8_t t(*it);
     for (size_t b(8); b > 0; --b) {
-      out << (t & 0x80 ? '1' : '0');
-      t <<= 1;
+      out << (t & 0x01 ? '1' : '0');
+      t >>= 1;
```
benlaurie: Seems odd that bit ordering is reversed?

cpp/merkletree/sparse_merkle_tree.h
```diff
@@ -171,29 +172,25 @@ class SparseMerkleTree {
...
 
   std::string Dump() const;
 
  private:
-  // WARNING WARNING WARNING
-  // 64 < 256 !
-  // WARNING WARNING WARNING
-  // TODO(alcutter): BIGNUM probably.
-  typedef uint64_t IndexType;
+  typedef cert_trans::BigNum IndexType;
```
benlaurie: bignums are an obvious way to do this, but quite a lot of baggage to store what is really an array of bits that you never do any real arithmetic on?

AlCutter: Part of the reason was to take advantage of the expand-to-fit nature of bignums, and I must admit it didn't occur to me that a freshly init'd BIGNUM structure could possibly be 24 bytes in size.  Back to the drawing board.

benlaurie: Like I said: baggage.

Why is expend-to-fit useful? Paths are fixed size...

AlCutter: Yeah, but e.g. while a whole int for the sign?
Anyway, it's useful because in this implementation, the leaf nodes are rarely actually at the end of their paths, rather they live at the first unique prefix of their path, so the leaf nodes themselves need to store the remainder.  Currently they just store the whole thing though, leading to duplication.

benlaurie: Can't get excited about the duplication. Still not seeing an argument against a byte array.

AlCutter: That's because nobody's arguing against a byte array; the BIGNUM was meant to be an easy wrapper for effectively that with the dynamic size thing thrown in for "free", and as is often the case "free" turns out to be startlingly expensive.

PR #1065, opened 2015-11-25
Add Python demo of general purpose verifiable map.

aeijdenberg:
Includes demo of general purpose database.


aeijdenberg:
@AlCutter - here it is

PR #1067, opened 2015-11-27
Fix libevent build.

AlCutter:
A clean gclient sync was failing locally during a clean build because `libevent` wasn't building `libevent_openssl.a`.  This was due to it missing a bunch of `libdl` symbols (presumably for engine support?)

Anyway, either Travis wasn't affected by this for some reason, or we didn't notice because of the caching.

### DIFF
--- a/build/Makefile.libevent
+++ b/build/Makefile.libevent
@@ -12,7 +12,7 @@ all: Makefile
 	$(MAKE) install
 
 Makefile: configure
-	./configure --prefix=$(INSTALL_DIR) --disable-shared --enable-static --enable-openssl CPPFLAGS="-I$(INSTALL_DIR)/include" CFLAGS="-fPIC" LDFLAGS="-fPIC -L$(INSTALL_DIR)/lib $(EXTRA_LDFLAGS)" || tail -1000 config.out
+	./configure --prefix=$(INSTALL_DIR) --disable-shared --enable-static --enable-openssl CPPFLAGS="-I$(INSTALL_DIR)/include" CFLAGS="-fPIC" LDFLAGS="-fPIC -L$(INSTALL_DIR)/lib" OPENSSL_LIBADD="$(EXTRA_LDFLAGS)" || tail -1000 config.out
 	$(MAKE) clean
 
 configure: configure.ac


pphaneuf:
I think the `libevent_openssl` might be there in the basic image? So we might have been getting that one from the system (yikes!)...

build/Makefile.libevent
```diff
@@ -12,7 +12,7 @@ all: Makefile
 	$(MAKE) install
 
 Makefile: configure
-	./configure --prefix=$(INSTALL_DIR) --disable-shared --enable-static --enable-openssl CPPFLAGS="-I$(INSTALL_DIR)/include" CFLAGS="-fPIC" LDFLAGS="-fPIC -L$(INSTALL_DIR)/lib $(EXTRA_LDFLAGS)" || tail -1000 config.out
+	./configure --prefix=$(INSTALL_DIR) --disable-shared --enable-static --enable-openssl CPPFLAGS="-I$(INSTALL_DIR)/include" CFLAGS="-fPIC" LDFLAGS="-fPIC -L$(INSTALL_DIR)/lib" OPENSSL_LIBADD="$(EXTRA_LDFLAGS)" || tail -1000 config.out
```
pphaneuf: I... don't know what this even does?

```bash
$ grep -rlI OPENSSL_LIBADD .
$
```

AlCutter: Did you run `autogen.sh` before doing the grep?

I get:
```bash
$ grep -rlI OPENSSL_LIBADD .
./libevent/sample/Makefile
./libevent/sample/Makefile.in
./libevent/sample/Makefile.am
./libevent/configure.ac
./libevent/include/Makefile
./libevent/include/Makefile.in
./libevent/config.log
./libevent/Makefile
./libevent/test/Makefile
./libevent/test/Makefile.in
./libevent/test/Makefile.am
./libevent/autom4te.cache/output.2
./libevent/autom4te.cache/output.1
./libevent/autom4te.cache/traces.2
./libevent/autom4te.cache/traces.1
./libevent/autom4te.cache/output.0
./libevent/autom4te.cache/traces.0
./libevent/Makefile.in
./libevent/config.status
./libevent/configure
./libevent/ChangeLog
```

pphaneuf: No, I did one better: I ran it on `certificate-transparency`, not `openssl`. :smiley: 

I thought I was reading `configure-ct`, for some reason, let me get back to you.

pphaneuf:
I don't really understand how that has an effect, since every command that involves `OPENSSL_LIBADD` should also have `LDFLAGS` anyway, but if it fixes it, LGTM then?

AlCutter:
I believe it's because of the ordering of `-l` libs that configure tacks on to `ld` while searching for `SSL_new`; the passed in `LDFLAGS` come before configure's `-l` probes, but `OPENSSL_LIBADD` gets put on the end of the command line.

pphaneuf:
Which is normally the distinction between an `LDFLAGS` (with `-L` flags) and a `LIBADD` or `LDADD` flag (with `-l` flags), but yeah, maybe it's ordering...

PR #1069, opened 2015-11-30
Add guard flags to v2 servers so they can't be run accidentally

Martin2112:
Just for safety while they're in development and incomplete.


cpp/server/ct-mirror_v2.cc
```diff
@@ -248,6 +252,11 @@ int main(int argc, char* argv[]) {
 
   util::InitCT(&argc, &argv);
 
+  if (!FLAGS_i_know_v2_is_not_finished_yet) {
```
pphaneuf: Could have just used a straight `CHECK`, but sure.

pphaneuf:
LGTM

pphaneuf:
I think the build failure isn't related to your change, but it frightens me a little. Could you copy out the relevant log output to a new issue, and retry it?

Martin2112:
Test failure was because of a flake. Going to merge this.

PR #1071, opened 2015-12-01
Add versioning support to cluster protos. Try to prevent mismatched nodes joining.

Martin2112:
Nodes will exit if their version differs from cluster config. As an additional safety net exclude mismatched nodes from quorum calculations and don't start a fetcher for them.


pphaneuf:
I found that rather confusing, because of using `enum Version` but then stating this is the "binary version". This is more to separate the *type* of server, isn't it? Rather than have some sort of protocol versioning (which is what I understood by "binary version")?

I think protocol versioning is probably not that useful (too hard to get right for what it'll give users), and while I think protecting against mixing up server types is a good idea, `enum Version` seems very poorly suited to this, as it only covers "CT v1" and "CT v2", leaving out X-JSON and whatever else (which might not make sense to add to that particular enum).

And I didn't look at the diff in detail, but it looks *huge* for what it does?!? 10 commits, 20 files? I'd have expected a startup check, bail out, job done? Sounds like it's possibly adding a lot of coupling? But I haven't properly looked, so this is just a gut feeling...

cpp/log/cluster_state_controller-inl.h
```diff
@@ -105,6 +105,9 @@ ClusterStateController<Logged>::ClusterStateController(
           std::bind(&ClusterStateController<Logged>::ClusterServingSTHUpdater,
                     this)) {
   CHECK_NOTNULL(base_.get());
+  // Set this early to make sure we catch version mismatches as the node starts
+  local_node_state_.set_node_ct_version(supported_ct_version);
```
AlCutter: Why not at L105 - i.e. in the c'tor rather than in the callback?

cpp/log/cluster_state_controller-inl.h
```diff
@@ -244,11 +247,18 @@ ClusterStateController<Logged>::GetFreshNodes() const {
     const bool is_self(
         node.second->state().hostname() == local_node_state_.hostname() &&
         node.second->state().log_port() == local_node_state_.log_port());
-    if (!is_self && node.second->state().has_newest_sth() &&
-        node.second->state().newest_sth().tree_size() >=
-            actual_serving_sth_->tree_size()) {
-      VLOG(1) << "Node is fresh: " << node.second->state().node_id();
-      fresh.push_back(node.second->state());
+    // A node that is not serving the same version as the controller cannot
```
AlCutter: /same version/same CT version/

cpp/log/cluster_state_controller-inl.h
```diff
@@ -244,11 +247,18 @@ ClusterStateController<Logged>::GetFreshNodes() const {
...
+    if (IsSameVersion(node.second.get()->state().node_ct_version())) {
+      if (!is_self && node.second->state().has_newest_sth() &&
+          node.second->state().newest_sth().tree_size() >=
+              actual_serving_sth_->tree_size()) {
+        VLOG(1) << "Node is fresh: " << node.second->state().node_id();
+        fresh.push_back(node.second->state());
+      }
+    } else {
+      VLOG(1) << "Node is inconsistent version: "
+              << node.second->state().node_id();
```
AlCutter: Maybe log the node ID & hostname:port too to help locate it?

cpp/log/cluster_state_controller-inl.h
```diff
@@ -283,6 +294,13 @@ void ClusterStateController<Logged>::OnClusterStateUpdated(
                              update.handle_.Entry().log_port())) {
         all_peers_.erase(it);
         it = all_peers_.end();
+      } else if (it != all_peers_.end() &&
+                 !IsSameVersion(it->second.get()->state().node_ct_version())) {
```
AlCutter: Don't need the `get()` here

cpp/log/cluster_state_controller-inl.h
```diff
@@ -296,7 +314,15 @@ void ClusterStateController<Logged>::OnClusterStateUpdated(
...
         // parties.
         all_peers_.emplace(node_id, peer);
-        fetcher_->AddPeer(node_id, peer);
+
+        // Only matching versions are added to fetcher so we don't talk
+        // to nodes that have a different version
+        if (!node_version_mismatch) {
+          fetcher_->AddPeer(node_id, peer);
+        } else {
+          VLOG(1) << "Node has mismatched version, dropping: " << node_id;
```
AlCutter: Should probably promote that to a `WARNING`?

Martin2112: I wish I'd started this on Rietveld now. Probably too late.

Martin

cpp/log/cluster_state_controller.h
```diff
@@ -101,6 +102,12 @@ class ClusterStateController {
   // Thread entry point for ServingSTH updater thread.
   void ClusterServingSTHUpdater();
 
+  // Check if a peer matches our version
+  inline bool IsSameVersion(ct::Version peer_version) const {
```
AlCutter: `IsSameCTVersion` ?

cpp/log/etcd_consistent_store-inl.h
```diff
@@ -410,13 +412,21 @@ util::Status EtcdConsistentStore<Logged>::SetClusterNodeState(
...
+    // nobody else is updating our cluster state.
+    ct::ClusterNodeState local_state(state);
+    local_state.set_node_id(node_id_);
+    EntryHandle<ct::ClusterNodeState> entry(GetNodePath(node_id_), local_state);
+    const std::chrono::seconds ttl(FLAGS_node_state_ttl_seconds);
+    return ForceSetEntryWithTTL(ttl, &entry);
+  } else {
+    return util::Status(
+        util::error::FAILED_PRECONDITION,
+        "SetClusterNodeState does not accept a different version");
```
AlCutter: `a state with a different CT version"`?

cpp/log/etcd_consistent_store.h
```diff
@@ -146,6 +147,7 @@ class EtcdConsistentStore : public ConsistentStore<Logged> {
 
   util::Status MaybeReject(const std::string& type) const;
 
+  ct::Version supported_ct_version_;
```
AlCutter: const

cpp/server/ct-server_v2.cc
```diff
@@ -89,6 +89,9 @@ static const bool cert_dummy =
 
 }  // namespace
 
+// This is the supported CT version of this binary. It can be used to
+// apply sanity checks such as when creating or joining clusters.
+const ct::Version kSUPPORTED_CT_VERSION = ct::V1;
```
AlCutter: `V2` :)

PR #1075, opened 2015-12-04
Improve performance of serializeV1SCT

mehmooda:
BenchmarkSCTOld  1000000 2135 ns/op 448 B/op 13 allocs/op
BenchmarkSCTNew 10000000  151 ns/op  64 B/op  1 allocs/op

### DIFF
--- a/go/serialization.go
+++ b/go/serialization.go
@@ -227,19 +227,28 @@ func UnmarshalDigitallySigned(r io.Reader) (*DigitallySigned, error) {
 	}, nil
 }
 
+func marshalDigitallySignedHere(ds DigitallySigned, here []byte) ([]byte, error) {
+	sigLen := len(ds.Signature)
+	dsOutLen := 2 + SignatureLengthBytes + sigLen
+	if here == nil {
+		here = make([]byte, dsOutLen)
+	}
+	if len(here) < dsOutLen {
+		return nil, ErrNotEnoughBuffer
+	}
+	here = here[0:dsOutLen]
+
+	here[0] = byte(ds.HashAlgorithm)
+	here[1] = byte(ds.SignatureAlgorithm)
+	binary.BigEndian.PutUint16(here[2:4], uint16(sigLen))
+	copy(here[4:], ds.Signature)
+
+	return here, nil
+}
+
 // MarshalDigitallySigned marshalls a DigitallySigned structure into a byte array
 func MarshalDigitallySigned(ds DigitallySigned) ([]byte, error) {
-	var b bytes.Buffer
-	if err := b.WriteByte(byte(ds.HashAlgorithm)); err != nil {
-		return nil, fmt.Errorf("failed to write HashAlgorithm: %v", err)
-	}
-	if err := b.WriteByte(byte(ds.SignatureAlgorithm)); err != nil {
-		return nil, fmt.Errorf("failed to write SignatureAlgorithm: %v", err)
-	}
-	if err := writeVarBytes(&b, ds.Signature, SignatureLengthBytes); err != nil {
-		return nil, fmt.Errorf("failed to write HashAlgorithm: %v", err)
-	}
-	return b.Bytes(), nil
+	return marshalDigitallySignedHere(ds, nil)
 }
 
 func checkCertificateFormat(cert ASN1Cert) error {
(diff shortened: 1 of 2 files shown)


AlCutter:
Hi, thanks for the PR, apologies it's taken so long to get to it.
Few comments below.

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
-	if err := checkExtensionsFormat(sct.Extensions); err != nil {
-		return nil, err
-	}
-	var buf bytes.Buffer
-	if err := binary.Write(&buf, binary.BigEndian, V1); err != nil {
-		return nil, err
-	}
-	if err := binary.Write(&buf, binary.BigEndian, sct.LogID); err != nil {
-		return nil, err
+func serializeV1SCTHere(sct SignedCertificateTimestamp, here []byte) (int, error) {
```
AlCutter: How about returning `([]byte, error)`?
That way you can slice the buffer to the actual used length for the caller (they can always ignore that if it's not helpful for their use case), and if they pass in `nil` then you can allocate the right size, fill it and return - avoiding the slightly strange double call to this method below.

TBH though, I'm wondering whether all of this serialization section mightn't be made better/more idiomatic by being redone around `io.Writer`.

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+	if sct.SCTVersion != 0 {
```
AlCutter: Use the `V1` enum

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+		return 0, INVALID_VERSION
 	}
-	if err := binary.Write(&buf, binary.BigEndian, sct.Timestamp); err != nil {
-		return nil, err
+
+	//Determine Lenth of Serialized Output
```
AlCutter: Could you pop a space after the `//` in the comments?
Also, don't Capitalise Each Word (could you do all the other comments which currently have that form too?), and `Lenth` is missing a `g` :)

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+	ctxlen := len(sct.Extensions)
```
AlCutter: How about `extLen` ?

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+	siglen := len(sct.Signature.Signature)
```
AlCutter: `sigLen`

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+	sctoutput := 1 + 32 + 8 + 2 + ctxlen + 2 + 2 + siglen
```
AlCutter: `outSize` or `outLen` maybe?

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+
+	if sctoutput > 0xFFFF {
```
AlCutter: Could you add a comment with a reference to the 6962 section which defines this size limit?  Better still: add the `0xFFFF` as an exported constant with that comment at the top of this file.

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+	here[0] = byte(sct.SCTVersion)
+
+	//Write LogID
+	copy(here[1:33], sct.LogID[:])
+
+	//Write Timestamp
+	binary.BigEndian.PutUint64(here[33:41], sct.Timestamp)
+
+	//Write Extensions
+	binary.BigEndian.PutUint16(here[41:43], uint16(ctxlen))
```
AlCutter: You've removed the call to `checkExtensionsFormat()`, could you add it back in?
(I get that in effect it's covered by the check for `sctoutput > 0xFFFF` above, but if those format checks become more involved at some point in the future we'll have to remember to search around for 'inlined' serialisations like this one, increasing the chances of introducing bugs/inconsistencies.

go/serialization.go
```diff
@@ -350,31 +350,59 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+	n := 43 + ctxlen
+	copy(here[43:n], sct.Extensions)
+
+	//Write Signature Algorithm
```
AlCutter: How about adding a `MarshalDigitallySignedHere()` and calling that from here?
(Similar rationale to above on code duplication.)

go/types.go
```diff
@@ -342,3 +342,24 @@ type Precertificate struct {
 func (m *MerkleTreeLeaf) X509Certificate() (*x509.Certificate, error) {
 	return x509.ParseCertificate(m.TimestampedEntry.X509Entry)
 }
+
+type scterror int
```
AlCutter: sctError

go/types.go
```diff
@@ -342,3 +342,24 @@ type Precertificate struct {
...
+
+//Preallocate errors for performance
+var (
+	INVALID_VERSION   error = scterror(1)
```
AlCutter: MixedCaps for constant names too, and golang's pkgs seem to tend to prefix with Err.
e.g. `ErrInvalidVersion`

go/types.go
```diff
@@ -342,3 +342,24 @@ type Precertificate struct {
...
+	SCT_TOO_LARGE     error = scterror(2)
+	NOT_ENOUGH_BUFFER error = scterror(3)
+)
+
+func (e scterror) Error() string {
+	switch e {
+	case INVALID_VERSION:
+		return "Unable to serialize SignedCertificateTimestamp: Invalid version"
```
AlCutter: Don't capitalise the initial letter of the error strings (so that they read easier if they're later concatenated with other strings and printed out somewhere up the stack.)

go/types.go
```diff
@@ -342,3 +342,24 @@ type Precertificate struct {
...
+	case SCT_TOO_LARGE:
+		return "Unable to serialize SignedCertificateTimestamp: SCT too large"
+	case NOT_ENOUGH_BUFFER:
+		return "Unable to serialize SignedCertificateTimestamp: Not enough buffer"
+	}
+	return "Unknown error occured in sct"
```
AlCutter: /sct/SCT/ for consistency, also maybe add "serialisation" to the end, otherwise it sounds like the SCT is broken somehow.

mehmooda:
In regards to the 0xFFFF SCT size limit. After looking through the RFC it seems that limit only applies when the SCT is placed in the SCTList structure for OCSP and TLS responses (although that is the only place I can see it used). I've removed it from the serialisation code.

New Patch follows.

Benchmark results:
```
BenchmarkSCTOld  1000000 2135 ns/op 448 B/op 13 allocs/op
BenchmarkSCTNew 10000000  151 ns/op  64 B/op  1 allocs/op
```

Benchmark Code:
```
func BenchmarkSCTOld(b *testing.B) {
	sct := defaultSCT()
	for i := 0; i < b.N; i++ {
		SerializeSCT(sct)
	}
}
```

AlCutter:
Thanks!

PR #1077, opened 2015-12-07
Shuffle serialization

AlCutter:
Few more bits pulled out by @pphaneuf

### DIFF
--- a/cpp/proto/serializer.h
+++ b/cpp/proto/serializer.h
@@ -56,6 +56,112 @@ enum class DeserializeResult {
 
 std::ostream& operator<<(std::ostream& stream, const DeserializeResult& r);
 
+typedef google::protobuf::RepeatedPtrField<std::string> repeated_string;
+typedef google::protobuf::RepeatedPtrField<ct::SctExtension>
+    repeated_sct_extension;
+
+SerializeResult CheckExtensionsFormat(const std::string& extensions);
+SerializeResult CheckSctExtensionsFormat(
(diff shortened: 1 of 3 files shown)


cpp/proto/serializer.h
```diff
@@ -368,28 +400,6 @@ class Deserializer {
     return DeserializeResult::OK;
   }
 
- private:
```
pphaneuf: I think I dinged this `private:` by mistake?!? Oh, yeah, I meant for this to be in this commit, put it in the wrong one:

https://github.com/pphaneuf/certificate-transparency/commit/136e12edebd44c06d3c25dac046210e114faf543#diff-4ae97ce672fdbd80f64b37442df606d4R345

There's a similar thing for `Serializer` above, too:

https://github.com/pphaneuf/certificate-transparency/commit/136e12edebd44c06d3c25dac046210e114faf543#diff-4ae97ce672fdbd80f64b37442df606d4R258

cpp/proto/serializer.h
```diff
@@ -369,27 +401,6 @@ class Deserializer {
   }
 
  private:
```
pphaneuf: Could you add this?

```C++
// This class is mostly a namespace for static methods.
// TODO(pphaneuf): Make this into normal functions in a namespace.
Deserializer() = delete;
```

pphaneuf:
LGTM, once my minor mistake is corrected. :smiley:

Issue #1078, opened 2015-12-10
Lots of ASN1Errors when monitoring WoSign's CT Log using logobserver.py


linshunghuang:
Running python/utilities/logobserver/logobserver.py against https://ct.wosign.com raises a lot of ASN1Errors (and therefore excluded from scan results), for instance:
```
ASN1Error: Corrupt time: 151207195316
```
These errors started to appear since the certificate entry 11309. Looking at the certificate entry 11309 (https://crt.sh/?id=11142769) of https://ct.wosign.com, the `not_before` timestamp looks reasonable, so I'm wondering if there's a bug in the python ASN1 parser, or are these certificates indeed malformed? Thanks!

Repro steps:
1. Add WoSign CT Log to python/ct/config/logs.config
```
ctlog:{
  log_server:'https://ct.wosign.com'
  log_id:'nk/3PcPOIgtpIXyJnkaAdqv414Y21cz8haMadWKLqIs='
  public_key_info:{
    type:ECDSA
    pem_key:'-----BEGIN ECDSA PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0D'
    'AQcDQgAE1+wvK3VPN7yjQ7qLZWY8fWrlDCqmwuUm/gx9TnzwOrzi0yLcAdAfbkOcXG6DrZ'
    'wV9sSNYLUdu6NiaX7rp6oBmw==\n-----END ECDSA PUBLIC KEY-----'
  }
}
```
2.
```
certificate-transparency$ cd python/
python$ export PYTHONPATH=${PYTHONPATH}:$(pwd)
python$ python utilities/logobserver/logobserver.py
```
Wait a minute or so and observe:
```
WARNING:root:Monitor state file not found, assuming first run.
ERROR:root:Error scanning certificate 11309 in batch <11000, 11430> - it will be excluded from the scan results
Traceback (most recent call last):
  File "/MY_PATH/certificate-transparency/python/ct/client/reporter.py", line 52, in _scan_der_cert
    partial_result += check.check(certificate) or []
  File "/MY_PATH/certificate-transparency/python/ct/cert_analysis/validity.py", line 47, in check
    not_before = certificate.not_before()
  File "/MY_PATH/certificate-transparency/python/ct/crypto/cert.py", line 543, in not_before
    return (self._asn1_cert["tbsCertificate"]["validity"]["notBefore"].
[...]

PR #1080, opened 2015-12-17
parse of asn1 utc time which don't ends with "Z" in cert

liangdong2718:
according to https://github.com/google/certificate-transparency/issues/1078, there are some certficates will cause exception raise in python code, this kind of certifcates "nobefore" and "noafter" part don't ends with "Z", so add some code to parse utc time which don't ends with "Z" in cert

### DIFF
--- a/python/ct/config/logs.config
+++ b/python/ct/config/logs.config
@@ -1,13 +1,13 @@
-ctlog:{
-  log_server:'https://ct.googleapis.com/pilot'
-  log_id:'pLkJkLQYWBSHuxOizGdwCjw1mAT5G9+443fNDsgN3BA='
-  public_key_info:{
-    type:ECDSA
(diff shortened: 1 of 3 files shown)


eranmes:
Apologies for the delay reviewing this PR. I thought about it a bit, and it seems to me the right way about it is only to apply this heuristic fix in non-strict DER parsing mode.
Note that the ASN.1 parser in our project has strict mode and non-strict mode where it can accept some violations.
This fix, as you propose it, applies in the strict mode - it should not be the case. Such malformed certificates should fail strict parsing.

eranmes:
I've taken this pull request in a modified form - only applying the fix in non-strict mode.
Out for review in https://github.com/google/certificate-transparency/pull/1110.

PR #1083, opened 2015-12-19
Add script to update mirror instance VM images.

AlCutter:


RJPercival:
LGTM, bar that one comment.

PR #1085, opened 2016-01-13
Reinstate the pre/certificate …WithType serializers.

AlCutter:
Turns out these are used externally.

### DIFF
--- a/cpp/proto/cert_serializer.h
+++ b/cpp/proto/cert_serializer.h
@@ -30,6 +30,14 @@ SerializeResult SerializePrecertChainEntry(
     const std::string& pre_certificate,
     const repeated_string& precertificate_chain, std::string* result);
 
+// These two functions are depended on externally.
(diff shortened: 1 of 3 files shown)


cpp/proto/cert_serializer.h
```diff
@@ -30,6 +30,14 @@ SerializeResult SerializePrecertChainEntry(
     const std::string& pre_certificate,
     const repeated_string& precertificate_chain, std::string* result);
 
+// These two methods are depended on externally.
```
pphaneuf: These are functions, not methods. :wink:

pphaneuf:
LGTM, assuming Travis is happy, and with that one comment tweak.

cpp/proto/cert_serializer.h
AlCutter: D'oh, fixed.

PR #1087, opened 2016-01-14
Don't force callers of util::InitCT() to be CT V1.

AlCutter:
They should be able to decide for themselves (and in fact, already do causing this to explode.)


pphaneuf:
LGTM

Martin2112:
Odd. I thought this was fixed up in the review. Oh well.

PR #1094, opened 2016-01-25
Strengthen detection of invalid UTF-8 content when parsing certs

phad:
CT Logs Aviator and Pilot ingested a cert whose subject contains a field with ISO-8859-1 data being presented as UTF-8.  This results in the cert_desc.py class parsing the field into a string containing the invalid Unicode code point U+DBE0.

[Pilot entry 976685, Aviator entry 390950, cert: https://crt.sh/?id=977942]

This commit checks each character of a Unicode string being decoded from UTF-8, and uses the standard library function unicodedata.name() to check that valid codepoints are obtained.  If it finds an unknown codepoint then it has another go assuming the string was encoded with ISO-8859-1 instead.


python/ct/client/db/cert_desc.py
```diff
@@ -94,7 +96,15 @@ def from_cert(certificate, observations=[]):
 
 
 def to_unicode(str_):
-    return unicode(str_, 'utf-8', 'replace')
+    encoded = unicode(str_, 'utf-8', 'replace')
+    for ch in encoded:
+        try:
+            ignored_name = unicodedata.name(ch)
```
eranmes: _ = unicodedata.name(ch) 
or even:
unicodedata.name(ch)

python/ct/client/db/cert_desc.py
```diff
@@ -94,7 +96,15 @@ def from_cert(certificate, observations=[]):
 
 
 def to_unicode(str_):
```
eranmes: nit: Odd name for a parameter (suffixed with '_'). Consider changing the parameter name as you've changed the entire function.

eranmes:
Two minor comments, otherwise LGTM.

python/ct/client/db/cert_desc.py
phad: Done

python/ct/client/db/cert_desc.py
phad: Yes, it is.  Done.

PR #1096, opened 2016-01-27
Fixer portion of the Go fixchain & preload tool

taknira:
The Fixer element takes chains and tries to create a valid chain for each
certificate in the given chain.  Errors are reported to one channel, and
successfully fixed chains are pushed to another.

Simplified intended usage of this element:

fixer := NewFixer(...)
fixer.QueueChain(...)   // Repeat this line for as many chains as you have.
fixer.Wait()


go/fixchain/fix_test.go
```diff
@@ -0,0 +1,244 @@
...
+		TryNextExpected:
+			for j, expChain := range ft.expectedChains {
+				if seen[j] {
+					continue
+				}
+				if len(chain) != len(expChain) {
+					continue
+				}
+				for k, cert := range chain {
+					if strings.Index(nameToKey(&cert.Subject), expChain[k]) == -1 {
```
benlaurie: ```if string.Contains(nameToKey(&cert.Subject), expChain[k]) {```

go/fixchain/fix_test.go
```diff
@@ -0,0 +1,244 @@
...
+		chains, ferr = fix.handleChain()
+	}
+
+	if !ft.expectErr && ferr != nil {
+		t.Errorf("#%d: Failed to get valid chain: %s", i, ferr.TypeString())
+	}
+
+	// Check for 1:1 correspondance between expectedChains and the chains that
+	// were produced by function call
+	if len(ft.expectedChains) != len (chains) {
```
benlaurie: no space after ```len```.

go/fixchain/fix_test.go
```diff
@@ -0,0 +1,244 @@
...
+		roots: []string{verisignRoot},
+
+		function: "constructChain",
+		expectErr: true,
+	},
+
+	// fixChain()
+	{	// Correct chain results in duplicates of chain due to behaviour of
+		// augmentIntermediates() and allowance of duplicate certs in bag.
+		// TODO(katjoyce): Would it be better to prevent duplication in bag?
```
benlaurie: Yes :-)

go/fixchain/fix.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+		}
+	}
+	fix.fixer.notFixed++
+	return nil, &FixError{Type: FixFailed, Cert: fix.cert,
+		Chain: fix.chain.certs}
+}
+
+func (fix *toFix) augmentIntermediates(url string) {
+	// PKCS#7 additions
```
benlaurie: Perhaps note that these exist because I couldn't find a go PKCS#7 library I wanted to use.

go/fixchain/fix_error.go
```diff
@@ -0,0 +1,49 @@
...
+const (
+	None errorType = iota
+	ParseFailure
+	CannotFetchURL
+	FixFailed
+	LogPostFailed
+	VerifyFailed
+)
+
+// FixError is the structure to which errors in the fixing process are reported
```
benlaurie: surely "with which" not "to which"?

go/fixchain/fix_test.go
benlaurie: Err ! that, obvs.

go/fixchain/fix_test.go
katjoyce: Having looked at the x509.CertPool.AddCert() code, it already takes care of duplicate certs within a CertPool.  Something else is going on here.

katjoyce: Solved: The two Thawte certs are actually different certs.  The one I provide in test_base.go and the one that is downloaded by getURL() are different certs, containing the same key.

go/fixchain/fix_test.go
katjoyce: Done.

go/fixchain/fix_test.go
katjoyce: Done.

go/fixchain/fix.go
katjoyce: Done.

go/fixchain/fix_error.go
katjoyce: Done.

go/fixchain/url_cache.go
```diff
@@ -0,0 +1,64 @@
...
+		u.hit++
+		return r, nil
+	}
+	c, err := u.client.Get(url)
+	if err != nil {
+		u.errors++
+		return nil, err
+	}
+	defer c.Body.Close()
+	if c.StatusCode != 200 {
```
benlaurie: Also cache errors?

go/fixchain/sync_fixer.go
```diff
@@ -0,0 +1,74 @@
...
+	Roots *x509.CertPool
+}
+
+// Fix synchronously attempts to fix all of the certificate chains passed to it.
+// Fix returns a list of successfully constructed or fixed chains, and a list of
+// errors that is encountered along the way, each in the form of a FixError.
+func (f *Fixer) Fix(chainsToFix []*ChainToFix) ([][]*x509.Certificate, []*FixError) {
+	chains := make(chan []*x509.Certificate)
+	errors := make(chan *FixError)
+	af := NewAsyncFixer(100, chains, errors, f.client, false)
```
benlaurie: Make the 100 a constant somewhere.

go/fixchain/url_cache.go
taknira: In what way should errors be cached?  I'm hesitant to cache the error as the response to the URL in case it is just a one off error, and trying again the next time could succeed.  Also the caller of this function (augmentIntermediates in fix.go) reports a FixError each time getURL returns an error, so the error will be recorded for inspection later.

go/fixchain/sync_fixer.go
taknira: Done

go/fixchain/url_cache.go
benlaurie: Errors are classified as permanent or temporary (these are HTTP errors I am talking about). Cache permanent errors...

OK to leave as a TODO.

taknira: Ah, I was unaware of this.  Every day's a school day!

ThomasHabets:
Sorry for the delay.

A sync wrapper around an async API is needlessly complex. Synchronous logic is much easier to write and debug. A sync wrapper is not complicated, but then neither is an async wrapper.

The problem with sync wrapping async is that to understand the actual logic you have to at the same time understand the asynchronousness.

With async wrapping sync the logic is decoupled from concurrency. And the concurrency wrapping becomes (in Go) so simple that often it's not even worth adding this second API.

"Expose synchronous APIs, calling them concurrently is easy."
-- http://talks.golang.org/2013/bestpractices.slide#26

taknira:
Sync wrapper is no more, and exposed now is a synchronous Fix() function that works with the internal fixing functions (which have all been made fully synchronous too) to fix a chain.  This application logic is actually quite simple, and this division will make it really easy to improve on the core fixing algorithm in the future :D 

The async wrapper bit - the Fixer - is code we need for the overall fixchain tool.  So we need that code to still be around too.

ThomasHabets:
Sorry for the delay in reviewing. Looks good now.

PR #1097, opened 2016-01-28
Revert back to using requests in the Python log client.

eranmes:
This reverts commit 3499f4ff6bd20627e6a11ee13aa79d5f4d96da43.

### DIFF
--- a/python/README
+++ b/python/README
@@ -3,7 +3,7 @@ To run the python scripts
 1. Install the following additional packages
 gflags (python_gflags if using easy_install)
 gviz-api-py (easy_install 'https://google-visualization-python.googlecode.com/files/gviz_api_py-1.8.2.tar.gz')
-httplib2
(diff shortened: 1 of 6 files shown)


python/ct/client/scanner.py
```diff
@@ -105,11 +105,11 @@ def _scan(entry_queue, log_url, range_description):
         entries = client.get_entries(range_start, range_end)
         scanned = range_start
         for entry in entries:
+            scanned += 1
```
phad: I can't remember why the original commit moved this.  It looks like it's intended to fix an off-by-one issue.

phad: So I think you should not modify scanner.py in this commit.

eranmes: Done.

eranmes:
PTAL.

PR #1102, opened 2016-02-01
Golang: added Ticker interface to Scanner

kyprizel:
If we need to store the latest STH we fetched and then restore fetching process later it'll be a big problem in current Scanner implementation. Introducing Tickers we can handle it via interfaces (example of logging included).


kyprizel:
I signed it!

PR #1103, opened 2016-02-02
correct GetEntries slice size

### LINKED ISSUE #1100: Golang: GetEntries returns uninitialized entries
The slice that is returned is initialized using the requested number of entries (essentially `end-start`). Since this number may not be the number of entries that a server actually returns uninitialized entries can be left at the end of the slice.

Instead the slice should be initialized using `len(resp.Entries)`, the number of entries the server actually returned.

jmhodges:
Fixes #1100, and probably a few other problems folks were having.

Before this was corrected, we saw many "sequence truncated" from the asn1 library when running large GetEntries on the certly CT log.

### DIFF
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -299,7 +299,7 @@ func (c *LogClient) GetEntries(start, end int64) ([]ct.LogEntry, error) {
 	if err != nil {
 		return nil, err
 	}
-	entries := make([]ct.LogEntry, end-start+1, end-start+1)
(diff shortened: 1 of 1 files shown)


jmhodges:
By "large GetEntries", I mean `Scanner.Scan`s with `BatchSize`s of 2000.

AlCutter:
Thanks!

benlaurie:
It'd be nice if there were a test showing the bug...

PR #1104, opened 2016-02-03
gclient fixes: Build and use lzma

eranmes:
lzma is needed by libunwind, which would fail the ct configure script
if not present.
Also, explicitly depend on libunwind for tcmalloc.


eranmes:
Abandoning, there may be a simpler solution.

PR #1106, opened 2016-02-03
Making unittest the first import in tests.

eranmes:
There's no need for a magic marker in that case.

### DIFF
--- a/python/ct/cert_analysis/algorithm_test.py
+++ b/python/ct/cert_analysis/algorithm_test.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import unittest
+
 import mock
 import time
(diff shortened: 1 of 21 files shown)


phad:
I think there are about 30 Python unit tests, you've only covered half here!
 - most of the tests under ./ct/crypto
./ct/cert_analysis/base_check_test.py
./ct/cert_analysis/extensions_test.py
./ct/client/db/sqlite*_test.py
./ct/client/monitor_test.py
./ct/client/prober_test.py
./ct/client/reporter_test.py
./ct/client/state_test.py

eranmes:
PTAL - Done for all tests where applicable (some tests like monitor_test use twisted's unittest module so it's not relevant)

PR #1107, opened 2016-02-04
gclient woes: Fix Linux build.

eranmes:
Done by:
- Configuring libunwind such that it does not require liblzma
- Explicitly stating that tcmalloc should link libunwind

### DIFF
--- a/build/Makefile.unwind
+++ b/build/Makefile.unwind
@@ -2,7 +2,7 @@ all: Makefile
 	$(MAKE) install
 
 Makefile: configure
-	./configure --prefix=$(INSTALL_DIR) --disable-shared --enable-static CXXFLAGS="-fPIC" CFLAGS="-fPIC"
+	./configure --prefix=$(INSTALL_DIR) --disable-minidebuginfo --disable-shared --enable-static CXXFLAGS="-fPIC" CFLAGS="-fPIC"
 	# Don't really want to install latex:
 	sed -e 's/latex2man/echo/' -e 's/pdflatex/echo/' -ibak doc/Makefile
 


AlCutter:
The title's a _bit_ misleading - it works just fine on Linux in Travis for example, this is presumably a fix for a specific flavour & version of a Linux distro... :)

build/Makefile.unwind
```diff
@@ -2,7 +2,7 @@ all: Makefile
 	$(MAKE) install
 
 Makefile: configure
-	./configure --prefix=$(INSTALL_DIR) --disable-shared --enable-static CXXFLAGS="-fPIC" CFLAGS="-fPIC"
+	./configure --prefix=$(INSTALL_DIR) --disable-minidebuginfo --disable-shared --enable-static CXXFLAGS="-fPIC" CFLAGS="-fPIC" LDFLAGS="-L$(INSTALL_DIR)/lib"
```
AlCutter: Do you still need the `LDFLAGS=…` given you're not trying to link lzma anymore?

eranmes:
Yeah, sorry, the title should be 'Break Travis Mac build in favour of fixing the gclient build on some flavours of Linux'.

eranmes:
@AlCutter PTAL.

build/Makefile.tcmalloc
```diff
@@ -3,6 +3,10 @@ ifeq ($(UNAME),FreeBSD)
 EXTRA_LDFLAGS="-lexecinfo"
 endif
 
+ifeq ($(UNAME),Linux)
+EXTRA_LDFLAGS="-lunwind"
```
AlCutter: You _may_ end up want this on FreeBSD too, certainly the configure-ct script adds `-lunwind` to the CT `configure` run for FreeBSD.

eranmes: Maybe, but I do not have access to FreeBSD to confirm. Suggestions?

AlCutter: FreeBSD VM?

Actually, are you sure you really need this now?
I'm thinking that this previously worked (and continues to work on Travis etc.), and so I'm a bit surprised that you need to do anything more than add the `--disable-minidebuginfo` flag for the unwind build.

eranmes: Removed, still seems to work on my particular flavour of Linux.

PR #1111, opened 2016-02-10
Python: Fix submit_chain to work with all logs

eranmes:
This is done by making it use verify.create_key_info_from_raw_key
which can cope with RSA and ECDSA key.


pphaneuf:
LGTM, if Travis agrees.

PR #1112, opened 2016-02-12
Python: Add include to the geneted log list header

eranmes:
The stddef header was added in https://codereview.chromium.org/1535363003,
this commit reflects this change in the generated header.


pphaneuf:
You've got a typo in that commit message, you can fix it with `git commit --amend` (make sure `git status` is clean first), since it's likely nobody looked at this yet? You'll have to use `-f` when pushing, that counts as a sort of rebase.

PR #1117, opened 2016-02-18
Dnslookup

benlaurie:
Illustrative DNS client.

### DIFF
--- a/python/utilities/dnslookup/dnslookup.py
+++ b/python/utilities/dnslookup/dnslookup.py
@@ -24,21 +24,18 @@ ALL_LOGS_URL = 'https://www.certificate-transparency.org/known-logs/all_logs_lis
 basepath = os.path.dirname(sys.argv[0])
 sys.path.append(os.path.join(basepath, '../../../python'))
 from ct.crypto import merkle, verify
-from ct.proto import ct_pb2, client_pb2
-
-ctpath = os.path.join(basepath, '../../../cpp/client/ct')
-libpath = os.path.join(basepath, '../../../../install/lib')
+from ct.proto import client_pb2
 
 class CTDNSLookup:
-    def __init__(self, domain, verifier, nameservers, port = 53):
+    def __init__(self, domain, verifier, resolver=None):
         self.verifier = verifier
         self.domain = domain
-        self.resolver = dns.resolver.get_default_resolver()
-        self.resolver.nameservers = nameservers
-        self.resolver.port = port
+        self.resolver = resolver
+        if not self.resolver:
+            self.resolver = dns.resolver.get_default_resolver()
 
     def Get(self, name):
-        logging.info('get %s' % name)
+        logging.info('get %s', name)
         answers = self.resolver.query(name, 'TXT')
         assert answers.rdtype == dns.rdatatype.TXT
         return answers
@@ -68,18 +65,18 @@ class CTDNSLookup:
         return self.GetOne('%d.%d.%d.tree' % (level, index, size))
 
     def GetIndexFromHash(self, hash):
-        return self.GetOne('%s.hash' % base64.b32encode(hash).replace('=', ''))
+        return self.GetOne('%s.hash' % base64.b32encode(hash).rstrip('='))
 
 if __name__ == '__main__':
-    logging.basicConfig(level="INFO")
+    logging.basicConfig(level='INFO')
 
     index = sys.argv[1]
 
-    keypem = ("-----BEGIN PUBLIC KEY-----\n"
-              "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEfahLEimAoz2t01p\n"
-              "3uMziiLOl/fHTDM0YDOhBRuiBARsV4UvxG2LdNgoIGLrtCzWE0J\n"
-              "5APC2em4JlvR8EEEFMoA==\n"
-              "-----END PUBLIC KEY-----\n")
+    keypem = ('-----BEGIN PUBLIC KEY-----\n'
+              'MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEfahLEimAoz2t01p\n'
+              '3uMziiLOl/fHTDM0YDOhBRuiBARsV4UvxG2LdNgoIGLrtCzWE0J\n'
+              '5APC2em4JlvR8EEEFMoA==\n'
+              '-----END PUBLIC KEY-----\n')
     logurl = 'http://ct.googleapis.com/pilot';
     logdns = 'pilot.ct.googleapis.com'
 
@@ -88,7 +85,7 @@ if __name__ == '__main__':
     j = response.read()
     j = json.loads(j)
     leaf_input = j['entries'][0]['leaf_input']
-    logging.info('leaf = %s' % leaf_input)
+    logging.info('leaf = %s', leaf_input)
     leaf = base64.b64decode(leaf_input)
     leaf_hash = hashlib.sha256(chr(0) + leaf).digest()
 
@@ -97,11 +94,11 @@ if __name__ == '__main__':
     keyinfo.pem_key =  keypem
     log_verifier = verify.LogVerifier(keyinfo)
 
-    lookup = CTDNSLookup(logdns, log_verifier, ['8.8.8.8'])
+    lookup = CTDNSLookup(logdns, log_verifier)
     sth = lookup.GetSTH()
-    logging.info('sth = %s' % sth)
+    logging.info('sth = %s', sth)
 
-    logging.info('hash = %s' % base64.b64encode(leaf_hash))
+    logging.info('hash = %s', base64.b64encode(leaf_hash))
     verifier = merkle.MerkleVerifier()
     index = int(index)
     audit_path = []
@@ -109,7 +106,7 @@ if __name__ == '__main__':
     apl = verifier.audit_path_length(index, sth.tree_size)
     for level in range(0, apl):
         h = lookup.GetEntry(level, index, sth.tree_size)
-        logging.info('hash = %s' % base64.b64encode(h))
+        logging.info('hash = %s', base64.b64encode(h))
         audit_path.append(h[:32])
 
         if prev:
@@ -122,7 +119,7 @@ if __name__ == '__main__':
 
         prev = h
 
-    logging.info('path = %s' % map(base64.b64encode, audit_path))
+    logging.info('path = %s', map(base64.b64encode, audit_path))
 
     assert verifier.verify_leaf_hash_inclusion(leaf_hash, index, audit_path,
                                                sth)
--- a/python/utilities/dnslookup/dnslookup.py
+++ b/python/utilities/dnslookup/dnslookup.py
@@ -14,12 +14,8 @@ import hashlib
 import json
 import logging
 import os
-import subprocess
 import sys
-import tempfile
 import urllib2
-
-ALL_LOGS_URL = 'https://www.certificate-transparency.org/known-logs/all_logs_list.json'
 
 basepath = os.path.dirname(sys.argv[0])
 sys.path.append(os.path.join(basepath, '../../../python'))
--- a/python/utilities/dnslookup/dnslookup.py
+++ b/python/utilities/dnslookup/dnslookup.py
@@ -122,7 +122,7 @@ if __name__ == '__main__':
 
         prev = h
 
-    logging.info("path = %s" % map(base64.b64encode, audit_path))
+    logging.info('path = %s' % map(base64.b64encode, audit_path))
 
     assert verifier.verify_leaf_hash_inclusion(leaf_hash, index, audit_path,
                                                sth)
--- a/python/utilities/dnslookup/dnslookup.py
+++ b/python/utilities/dnslookup/dnslookup.py
@@ -6,6 +6,8 @@
 # does not start with an SSL connection, but instead fetches a log entry by
 # index and then verifies the proof over DNS.
 
+# You will need to install DNSPython (http://www.dnspython.org/)
+
 import base64
 import dns.resolver
 import hashlib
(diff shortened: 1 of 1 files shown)


python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
+
+# Implement DNS inclusion proof checking, see [TBD].
+#
+# Unfortunately, getting at the SCTs in general is hard in Python, so this
+# does not start with an SSL connection, but instead fetches a log entry by
+# index and then verifies the proof over DNS.
+
+import base64
+import dns.resolver
```
pphaneuf: Where does this module come from? It's not mentioned [here](https://docs.python.org/2.7/py-modindex.html)...

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,130 @@
...
+from ct.proto import ct_pb2, client_pb2
+
+ctpath = os.path.join(basepath, '../../../cpp/client/ct')
+libpath = os.path.join(basepath, '../../../../install/lib')
+
+class CTDNSLookup:
+    def __init__(self, domain, verifier, nameservers, port = 53):
+        self.verifier = verifier
+        self.domain = domain
+        self.resolver = dns.resolver.Resolver(configure=False)
```
pphaneuf: I'd personally favour passing in a `Resolver` instance. Especially as the way it is used below, it is probably closer to `configure=True`, effectively? I'm guessing that means "set yourself up from `/etc/resolv.conf`(or whatever is appropriate on your platform)"?

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,130 @@
...
+              "-----END PUBLIC KEY-----\n")
+    logurl = 'http://ct.googleapis.com/pilot';
+    logdns = 'pilot.ct.googleapis.com'
+
+    response = urllib2.urlopen(logurl + '/ct/v1/get-entries?start=' + index
+                               + '&end=' + index)
+    j = response.read()
+    j = json.loads(j)
+    leaf_input = j['entries'][0]['leaf_input']
+    logging.info('leaf = ' + leaf_input)
```
pphaneuf: Use `'leaf = %s', leaf_input`?

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,130 @@
...
+class CTDNSLookup:
+    def __init__(self, domain, verifier, nameservers, port = 53):
+        self.verifier = verifier
+        self.domain = domain
+        self.resolver = dns.resolver.Resolver(configure=False)
+        self.resolver.nameservers = nameservers
+        self.resolver.port = port
+
+    def Get(self, name):
+        logging.info("get " + name)
```
pphaneuf: Here too, `'get %s', name`.

pphaneuf: Also, use single quotes consistently (there's a few other instances of double-quotes.

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,130 @@
...
+        self.verifier.verify_sth(sth)
+
+        return sth
+
+    def GetEntry(self, level, index, size):
+        return self.GetOne(str(level) + '.' + str(index) + '.' + str(size)
+                           + '.tree')
+
+    def GetIndexFromHash(self, hash):
+        return self.GetOne(base64.b32encode(hash)[:-4] + '.hash')
```
pphaneuf: This `-4` here depends on the the hash algorithm, maybe use a constant for that? Not going to have you make it compute the padding from the length of the hash output... :wink:

pphaneuf:
Should be about it, I think.

python/utilities/dnslookup/dnslookup.py
benlaurie: Adding a note.

python/utilities/dnslookup/dnslookup.py
benlaurie: Good catch. This code was originally testing a local DNS server, so I'm guessing that's why I had it set up this way. I changed this to get_default_resolver() instead, is that OK?

python/utilities/dnslookup/dnslookup.py
benlaurie: Yeah, done, and all the other places I should've done it.

benlaurie:
PTAL

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+        return sth
+
+    def GetEntry(self, level, index, size):
+        return self.GetOne('%d.%d.%d.tree' % (level, index, size))
+
+    def GetIndexFromHash(self, hash):
+        return self.GetOne('%s.hash' % base64.b32encode(hash).replace('=', ''))
+
+if __name__ == '__main__':
+    logging.basicConfig(level="INFO")
```
pphaneuf: Double quotes here, and a few lines below.

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+ctpath = os.path.join(basepath, '../../../cpp/client/ct')
+libpath = os.path.join(basepath, '../../../../install/lib')
+
+class CTDNSLookup:
+    def __init__(self, domain, verifier, nameservers, port = 53):
+        self.verifier = verifier
+        self.domain = domain
+        self.resolver = dns.resolver.get_default_resolver()
+        self.resolver.nameservers = nameservers
+        self.resolver.port = port
```
pphaneuf: Getting the default resolver, then overriding its nameservers and port, does that make any sense?

I'd have expected something like `resolver=None` in the parameters, and then this:

```python
self.resolver = resolver
if not self.resolver:
  self.resolver = dns.resolver.get_default_resolver()
```

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+basepath = os.path.dirname(sys.argv[0])
+sys.path.append(os.path.join(basepath, '../../../python'))
+from ct.crypto import merkle, verify
+from ct.proto import ct_pb2, client_pb2
+
+ctpath = os.path.join(basepath, '../../../cpp/client/ct')
+libpath = os.path.join(basepath, '../../../../install/lib')
+
+class CTDNSLookup:
+    def __init__(self, domain, verifier, nameservers, port = 53):
```
pphaneuf: Missed that on my first pass, but there shouldn't be spaces around the "=".

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+
+ALL_LOGS_URL = 'https://www.certificate-transparency.org/known-logs/all_logs_list.json'
+
+basepath = os.path.dirname(sys.argv[0])
+sys.path.append(os.path.join(basepath, '../../../python'))
+from ct.crypto import merkle, verify
+from ct.proto import ct_pb2, client_pb2
+
+ctpath = os.path.join(basepath, '../../../cpp/client/ct')
+libpath = os.path.join(basepath, '../../../../install/lib')
```
pphaneuf: Missed that too, somehow: these two things aren't used anywhere, ding them?

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+        self.resolver.port = port
+
+    def Get(self, name):
+        logging.info('get %s' % name)
+        answers = self.resolver.query(name, 'TXT')
+        assert answers.rdtype == dns.rdatatype.TXT
+        return answers
+
+    def GetOne(self, name):
+        name += '.%s' % self.domain
```
pphaneuf: This was fine before, but sure. :smiley:

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+        sth.timestamp = int(parts[1])
+        sth.sha256_root_hash = base64.b64decode(parts[2])
+        sth.tree_head_signature = base64.b64decode(parts[3])
+
+        self.verifier.verify_sth(sth)
+
+        return sth
+
+    def GetEntry(self, level, index, size):
+        return self.GetOne('%d.%d.%d.tree' % (level, index, size))
```
pphaneuf: This, on the other hand, is a good call, I should have suggested it. :wink:

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def GetIndexFromHash(self, hash):
+        return self.GetOne('%s.hash' % base64.b32encode(hash).replace('=', ''))
```
pphaneuf: `.rstrip('=')` feels a bit safer to me, but there really shouldn't be any "=" in that, obviously...

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+class CTDNSLookup:
+    def __init__(self, domain, verifier, nameservers, port = 53):
+        self.verifier = verifier
+        self.domain = domain
+        self.resolver = dns.resolver.get_default_resolver()
+        self.resolver.nameservers = nameservers
+        self.resolver.port = port
+
+    def Get(self, name):
+        logging.info('get %s' % name)
```
pphaneuf: Oh, oops: in `logging.info` and friends, you don't have to use `%`, you just go `logging.info('get %s', name)`.

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+              "-----END PUBLIC KEY-----\n")
+    logurl = 'http://ct.googleapis.com/pilot';
+    logdns = 'pilot.ct.googleapis.com'
+
+    response = urllib2.urlopen('%s/ct/v1/get-entries?start=%s&end=%s'
+                               %  (logurl, index, index))
+    j = response.read()
+    j = json.loads(j)
+    leaf_input = j['entries'][0]['leaf_input']
+    logging.info('leaf = %s' % leaf_input)
```
pphaneuf: Same here.

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+    leaf_hash = hashlib.sha256(chr(0) + leaf).digest()
+
+    keyinfo = client_pb2.KeyInfo()
+    keyinfo.type = keyinfo.ECDSA
+    keyinfo.pem_key =  keypem
+    log_verifier = verify.LogVerifier(keyinfo)
+
+    lookup = CTDNSLookup(logdns, log_verifier, ['8.8.8.8'])
+    sth = lookup.GetSTH()
+    logging.info('sth = %s' % sth)
```
pphaneuf: And here.

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+
+    logging.info('hash = %s' % base64.b64encode(leaf_hash))
```
pphaneuf: Here too.

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+    verifier = merkle.MerkleVerifier()
+    index = int(index)
+    audit_path = []
+    prev = None
+    apl = verifier.audit_path_length(index, sth.tree_size)
+    for level in range(0, apl):
+        h = lookup.GetEntry(level, index, sth.tree_size)
+        logging.info('hash = %s' % base64.b64encode(h))
```
pphaneuf: Some more.

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,131 @@
...
+            if level < apl - 6:
+                assert prev[32:] == h[:-32]
+            else:
+                assert prev[32:] == h
+        else:
+            assert len(h) == 32 * min(7, apl)
+
+        prev = h
+
+    logging.info('path = %s' % map(base64.b64encode, audit_path))
```
pphaneuf: And finally. :smiley:

python/utilities/dnslookup/dnslookup.py
benlaurie: Done.

python/utilities/dnslookup/dnslookup.py
benlaurie: Ah, I see what you mean. Done.

python/utilities/dnslookup/dnslookup.py
benlaurie: No longer relevant.

python/utilities/dnslookup/dnslookup.py
benlaurie: Left over from when I was trying to use the ct utility. Dinged.

benlaurie: Also dinged the protobuf I don't use anymore. :-)

python/utilities/dnslookup/dnslookup.py
benlaurie: rstrip is clearer.

python/utilities/dnslookup/dnslookup.py
benlaurie: Done everywhere.

benlaurie:
PTAL

pphaneuf:
LGTM

python/utilities/dnslookup/dnslookup.py
```diff
@@ -0,0 +1,124 @@
...
+                               %  (logurl, index, index))
+    j = response.read()
+    j = json.loads(j)
+    leaf_input = j['entries'][0]['leaf_input']
+    logging.info('leaf = %s', leaf_input)
+    leaf = base64.b64decode(leaf_input)
+    leaf_hash = hashlib.sha256(chr(0) + leaf).digest()
+
+    keyinfo = client_pb2.KeyInfo()
+    keyinfo.type = keyinfo.ECDSA
```
eranmes: Please use create_key_info_from_raw_key from https://github.com/google/certificate-transparency/blob/master/python/ct/crypto/verify.py#L399  rather than hard-coding the key type.

pphaneuf:
Rebase, squash, merge --ff-only?

pphaneuf:
Or, at the very least, rebase and fix the comment from @eranmes?

PR #1118, opened 2016-02-18
Fix service script for changes to gcloud command syntax.

Martin2112:


Martin2112:
I added the note to this PR about the health check workaround.

PR #1119, opened 2016-02-22
Documentation: Update Java instructions.

eranmes:


README.md
```diff
@@ -53,6 +53,15 @@ replacing `branch` as appropriate
 gclient config --name="certificate-transparency" https://github.com/google/certificate-transparency.git@branch
 ```
 
+### Language Specific Notes ###
+
+#### Java ####
```
benlaurie: Isn't protoc also required for other languages?

pphaneuf: Yeah, and in the "old" Ubuntu quickstart, it had you install `libprotobuf-java`, which depends on `protobuf-compiler` (which was also listed explicitly anyway).

This is the "old" stuff, because the "new" stuff is gclient, but Java isn't included in gclient (and if you have me build a JVM and Java libraries as part of building a CT server, I might have to lodge a formal complaint of some sort). :wink:

eranmes:
I can modify the documentation to say that if gclient was used to build the project then the proto compiler would be under `protobuf/src/protoc`, will that do?

README.md
benlaurie: On 22 February 2016 at 12:23, Pierre Phaneuf <notifications@github.com>
wrote:

> In README.md
> <https://github.com/google/certificate-transparency/pull/1119#discussion_r53617603>
> :
>
> > @@ -53,6 +53,15 @@ replacing `branch` as appropriate
> >  gclient config --name="certificate-transparency" https://github.com/google/certificate-transparency.git@branch
> >  ```
> >
> > +### Language Specific Notes ###
> > +
> > +#### Java ####
>
> Yeah, and in the "old" Ubuntu quickstart, it had you install
> libprotobuf-java, which depends on protobuf-compiler (which was also
> listed explicitly anyway).
>
> This is the "old" stuff, because the "new" stuff is gclient, but Java
> isn't included in gclient (and if you have me build a JVM and Java
> libraries as part of building a CT server, I might have to lodge a formal
> complaint of some sort). [image: :wink:]
>

Is it possible to have optional parts in a gclient build?

benlaurie:
On 22 February 2016 at 16:01, Eran Messeri <notifications@github.com> wrote:

> I can modify the documentation to say that if gclient was used to build
> the project then the proto compiler would be under protobuf/src/protoc,
> will that do?
>

Probably not because we still have instructions for non-gclient builds. We
could consider dropping those, tho?

eranmes:
@benlaurie Do I understand correctly that what you're suggesting is:
(1) Remove the instructions for non-gclient builds (which I also find misleading)
(2) Indicate that after a gclient build the protobuf compiler would be in a particular location and inform how the Java build can be made to use it.
(3) Point out the protobuf.jar file still has to be obtained independently?

pphaneuf:
You'd still have to get the *correct* protobuf.jar, right? Since there's absolutely zero effort on the part of the protobuf team to support mismatched `protoc` and libraries, I'd say that the breaking point we know of (at 2.5.0) is only *one* that we happen to know about, rather than an exception to an otherwise promised backward-/forward-compatibility...

I'm not entirely trusting of gclient, myself, and I've kept on building without, to make sure it's not broken (and helped Al find and fix weird issue because I could still *understand* what was going on in my build!)... :wink:

If my first paragraph is correct, either gclient should also compile the Java bits of protobuf, or the Java bits aren't hermetic (and thus don't fit at all with gclient), I'd say?

Martin2112:
Couldn't we set up a separate gclient build for the java code? Then people
who just want the log servers wouldn't need to install the Java
dependencies.

Martin

pphaneuf:
I've been kind of thinking that maybe the Java stuff should go to a separate google/certificate-transparency-java repo... It only shares a tiny sliver that's more accidental than anything else (it's a protobuf, but I don't think it gets serialised anywhere, it's just used as an in-memory data structure?)...

eranmes:
That escalated  quickly. 
Is the suggestion to move the code to a different repo (making it less discoverable, harder to work with) just because gclient can't be made to provide the Java dependencies?
If so, it seems to me like a rather extreme measure to take - fixing the documentation would be much less effort than moving the Java code to its own repo and making a build system for it.

Martin2112:
I'm not convinced about moving it elsewhere. I just thought it wasn't ideal
to add it to the main gclient build because of Java dependencies, and they
might not be available on all platforms where people build log servers.

Martin

benlaurie:
I think making a separate gclient file for the Java is a good idea.

PR #1120, opened 2016-02-22
Python: Parse flags in tests.

eranmes:
Since these tests import modules that use gflags, flags must be parsed
at some point.


pphaneuf:
LGTM

PR #1122, opened 2016-02-23
Python: Make more tests use test_config

eranmes:
Tests that need a test TLDList have been converted, as well as a few tests that
were left behind on the previous round.

### DIFF
--- a/python/ct/cert_analysis/common_name_test.py
+++ b/python/ct/cert_analysis/common_name_test.py
@@ -8,6 +8,7 @@ from ct.cert_analysis import common_name
 from ct.cert_analysis import tld_check
 from ct.cert_analysis import tld_list
 from ct.crypto import cert
+from ct.test import test_config
 
 def gen_common_name(name):
     common_name = mock.Mock()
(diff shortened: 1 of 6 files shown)


python/ct/cert_analysis/common_name_test.py
```diff
@@ -19,9 +20,6 @@ def cert_with_urls(*args):
     certificate.subject_common_names = mock.Mock(return_value=list(args))
     return certificate
 
-tlds = tld_list.TLDList(tld_dir="ct/cert_analysis/test_data/",
```
phad: This was unused I guess?

python/ct/test/test_config.py
```diff
@@ -6,3 +6,6 @@
 
 def get_test_file_path(filename):
     return os.path.join(os.curdir, CRYPTO_TEST_DATA_DIR, filename)
+
+def get_tld_directory():
+    return os.path.join(os.curdir, "ct/cert_analysis/test_data")
```
phad: Make the string a named constant like the other one above.

python/ct/cert_analysis/common_name_test.py
eranmes: Correct.

python/ct/test/test_config.py
eranmes: Done.

eranmes:
PTAL

PR #1124, opened 2016-02-29
Add support for serializing SCT Lists

mehmooda:
depends on #1075

### DIFF
--- a/go/serialization.go
+++ b/go/serialization.go
@@ -515,6 +515,10 @@ func SerializeSTHSignatureInput(sth SignedTreeHead) ([]byte, error) {
 
 // SCTListSerializedLength determines the length of the required buffer should a SCT List need to be serialized
 func SCTListSerializedLength(scts []SignedCertificateTimestamp) (int, error) {
+	if len(scts) == 0 {
+		return 0, fmt.Errorf("SCT List empty")
+	}
+
 	sctListLen := 2
 	for i, sct := range scts {
 		n, err := sct.SerializedLength()
@@ -525,10 +529,6 @@ func SCTListSerializedLength(scts []SignedCertificateTimestamp) (int, error) {
 			return 0, fmt.Errorf("SCT in position %d too large: %d", i, n)
 		}
 		sctListLen += 2 + n
-	}
-
-	if sctListLen == 2 {
-		return 0, fmt.Errorf("SCT List empty")
 	}
 
 	if sctListLen > MaxSCTListLength+2 {
--- a/go/serialization.go
+++ b/go/serialization.go
@@ -510,3 +510,66 @@ func SerializeSTHSignatureInput(sth SignedTreeHead) ([]byte, error) {
 		return nil, fmt.Errorf("unsupported STH version %d", sth.Version)
 	}
 }
+
+// SCTListSerializedLength determines the length of the required buffer should a SCT List need to be serialized
+func SCTListSerializedLength(scts []SignedCertificateTimestamp) (int, error) {
+	sctListLen := 2
+	for i, sct := range scts {
+		n, err := sct.SerializedLength()
+		if err != nil {
+			return 0, fmt.Errorf("unable to determine length of SCT in position %d: %v", i, err)
+		}
+		if n > 0xFFFF {
+			return 0, fmt.Errorf("SCT in position %d too large: %d", i, n)
+		}
+		sctListLen += n + 2
+	}
+
+	if sctListLen == 2 {
+		return 0, fmt.Errorf("SCT List empty")
+	}
+
+	if sctListLen > 0xFFFF+2 {
+		return 0, fmt.Errorf("SCT List too large to serialize: %d", sctListLen)
+	}
+	return sctListLen, nil
+}
+
+// SerializeSCTListHere serializes the passed in slice of SignedCertificateTimestamp in the
+// SCT List structure. If the passed in byte slice is not nil it will be used, otherwise
+// a buffer will be allocated and returned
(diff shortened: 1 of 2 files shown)


AlCutter:
I've just merged #1075, could you rebase this on top of `master`?

Also, if you could put any changes to address comments in a separate commit for now that'd be super helpful to me when reviewing - we can rebase and squash as a last step before merging, thanks!

mehmooda:
In RFC6962 3.3.  Including the Signed Certificate Timestamp in the TLS Handshake
...
SignedCertificateTimestampList ::= OCTET STRING
...
I think it would be more useful to return the SCT List already encoded in asn1. 

What do you think? I've got an "implementation" at dc6cec588a37723eedf02a1a83236c39a28badee

AlCutter:
Hmm, that hand-rolled ASN.1 bit makes me feel a bit uncomfortable...
What's the performance win from that over using the `encoding/asn1` package?

mehmooda:
Reworked patch,

regarding exporting SCTListSerializedLength its currently the only way for callers to determine the length of buffer needed for the Here method unless they loop around themselves.

Also hand serializing ASN1 vs asn1.Marshal
```
ASN1byHand  300000  5085 ns/op  128 B/op  1 allocs/op
ASN1Marshal  30000 45368 ns/op 1072 B/op 11 allocs/op
```

benchmark run on a raspberry pi because I like to get a drink between builds

go/serialization.go
```diff
@@ -350,44 +361,84 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
-func serializeV1SCT(sct SignedCertificateTimestamp) ([]byte, error) {
-	if err := checkExtensionsFormat(sct.Extensions); err != nil {
-		return nil, err
+// SerializedLength will return the space (in bytes)
+func (sct SignedCertificateTimestamp) SerializedLength() (int, error) {
+	switch sct.SCTVersion {
+	case V1:
+		extLen := len(sct.Extensions)
+		sigLen := len(sct.Signature.Signature)
+		return 1 + 32 + 8 + 2 + extLen + 2 + 2 + sigLen, nil
```
rolandshoemaker: It would be nice to stick a comment here explaining the breakdown of these lengths, otherwise it's a little confusing.

go/serialization.go
```diff
@@ -350,44 +361,84 @@ func SerializeSCTSignatureInput(sct SignedCertificateTimestamp, entry LogEntry)
...
+		return nil, ErrNotEnoughBuffer
 	}
-	sig, err := MarshalDigitallySigned(sct.Signature)
-	if err != nil {
+	if err := checkExtensionsFormat(sct.Extensions); err != nil {
 		return nil, err
 	}
-	if err := binary.Write(&buf, binary.BigEndian, sig); err != nil {
+
+	here = here[0:sctLen]
```
rolandshoemaker: Laying out the byte structure of this more explicitly in a comment would make the next section a bit more readable.

go/serialization.go
```diff
@@ -461,3 +512,77 @@ func SerializeSTHSignatureInput(sth SignedTreeHead) ([]byte, error) {
...
+		if err != nil {
+			return 0, fmt.Errorf("unable to determine length of SCT in position %d: %v", i, err)
+		}
+		if n > MaxSCTInListLength {
+			return 0, fmt.Errorf("SCT in position %d too large: %d", i, n)
+		}
+		sctListLen += 2 + n
+	}
+
+	if sctListLen == 2 {
```
rolandshoemaker: Since `SerializedLength` can never return both a `0` length and a `nil` error why not move this check to above the `for` loop and check `len(scts) == 0`.

rolandshoemaker:
This looks great although I'm somewhat confused about the benefit of being able to pass in a `here` byte slice to fill. You allow slices that are larger than `here` but always begin writing to the start of the slice so you can't serialize multiple lists into a single slice.

Is this just to enable slice re-use between calls without having to initialize a new one? If so I think you should be checking that the provided slice is exactly the correct size or if not doesn't contain trailing garbage as you are essentially providing two ways to get the slice back, either the return arg, or from the slice you are passing in which allows people to very easily shoot themselves in the foot and re-use the `here` slice in a way that will create invalid serialization.

If it's the case that the `...Here` methods are being used for performance when a reusing input slice I think the gains are outweighed by the complex API, especially when this code will probably live in a path that isn't super performance sensitive.

jsha:
What's the latest on this? I agree with @rolandshoemaker's comment that we probably don't need a "Here" variant.

rolandshoemaker:
@AlCutter do you have any objections to the current revision? I think providing the `SerializeSCTList` method and the changes in `SerializeSCTListHere` are enough to keep people from shooting themselves in the feet (...probably).

We'd love to get this landed in the nearish future so we can start working on Boulder support.

jsha:
@mehmooda: Regarding the above comments on benchmarking, it looks like you've put a bunch of effort into making this very efficient. But I don't anticipate this code being used in a highly performance-sensitive context. In Boulder, where we intend to use it for Let's Encrypt, the difference between the 45 microseconds / op version based on `asn1.Marshal` and the 5 microseconds / op version based on hand-rolled ASN.1 is negligible, and we'd prefer the one based on `asn1.Marshal` on the grounds that it's easier to reason about and more likely to be correct. Is there a reason you prefer the heavily-optimized version?

AlCutter:
@mehmooda, I really appreciate the effort you've put in to optimising this, but @jsha sums it up well; the hand-rolled ASN.1 code comes with a cost in complexity that doesn't seem to be a good trade off currently, could you slim it down to call to the asn1 library?

PR #1125, opened 2016-03-01
Log SQLite error message on CHECK failure

RJPercival:
Makes CHECK failures involving SQLite more intelligible (see https://groups.google.com/forum/#!topic/certificate-transparency/GhrqLF7Xs18).

### DIFF
--- a/cpp/log/sqlite_db.cc
+++ b/cpp/log/sqlite_db.cc
@@ -61,33 +61,35 @@ sqlite3* SQLiteOpen(const string& dbfile) {
   if (ret == SQLITE_OK) {
     return retval;
   }
-  CHECK_EQ(SQLITE_CANTOPEN, ret);
+  CHECK_EQ(SQLITE_CANTOPEN, ret) << sqlite3_errmsg(retval);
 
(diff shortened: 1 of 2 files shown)


cpp/log/sqlite_db.cc
```diff
@@ -61,33 +61,35 @@ sqlite3* SQLiteOpen(const string& dbfile) {
...
            sqlite3_exec(
                retval,
                "CREATE TABLE trees(sth BLOB UNIQUE, timestamp INTEGER UNIQUE)",
-               nullptr, nullptr, nullptr));
+               nullptr, nullptr, nullptr)) << sqlite3_errmsg(retval);
 
   CHECK_EQ(SQLITE_OK,
            sqlite3_exec(retval, "CREATE TABLE node(node_id BLOB UNIQUE)",
-                        nullptr, nullptr, nullptr));
+                        nullptr, nullptr, nullptr)) < sqlite3_errmsg(retval);
```
AlCutter: `<<`? :)

RJPercival: Oops, good catch! Fixed.

PR #1127, opened 2016-03-07
Fixchain

taknira:
The rest of the open source fixchain code.

Tests for fix_and_log.go are not yet present, but will be added very soon.  Submitting pull request now so that code review can get under way while last tests are being completed.

Structure of new parts:

-  \*\_test.go and test\_\* files are all to do with tests.
-  post.go contains the synchronous function that posts a chain to a certificate transparency log.
-  logger.go contains the asynchronous mechanism that posts queued chains to a certificate transparency log.
-  fix_and_log.go contains an asynchronous mechanism that fires up a Fixer and a Logger and passes chains to the Fixer, and then takes the fixed chains output by the Fixer and passes them to the Logger to be logged.
-  main/fixchain.go is an open source example of how to use the FixAndLog structure to fix and log certficate chains, and store any errors that result.


go/fixchain/fix.go
```diff
@@ -72,44 +74,94 @@ func (fix *toFix) constructChain() ([][]*x509.Certificate, []*FixError) {
...
+	for i, cert := range dchain.certs {
+		lookup[hash(cert)] = i
+	}
+
+	for i, cert := range dchain.certs {
+		if explored[i] {
+			continue
+		}
+
+		seen := make(map[[hashSize]byte]bool)
```
benlaurie: Why create this here? Why not return it from buildChain?

benlaurie: Actually, I see why. :-)

go/fixchain/fix.go
```diff
@@ -72,44 +74,94 @@ func (fix *toFix) constructChain() ([][]*x509.Certificate, []*FixError) {
 }
 
 func (fix *toFix) fixChain() ([][]*x509.Certificate, []*FixError) {
```
benlaurie: Document what this function does (i.e. inputs and outputs).

go/fixchain/fix.go
```diff
@@ -72,44 +74,94 @@ func (fix *toFix) constructChain() ([][]*x509.Certificate, []*FixError) {
...
+
+	return nil, append(retferrs, &FixError{
 		Type:  FixFailed,
 		Cert:  fix.cert,
 		Chain: fix.chain.certs,
 	})
 }
 
-func (fix *toFix) augmentIntermediates(url string) *FixError {
+func (fix *toFix) buildChain(cert *x509.Certificate, length int, seen map[[hashSize]byte]bool) ([][]*x509.Certificate, []*FixError) {
```
benlaurie: Document this function.

go/fixchain/fix.go
```diff
@@ -72,44 +74,94 @@ func (fix *toFix) constructChain() ([][]*x509.Certificate, []*FixError) {
...
+	return nil, retferrs
+}
+
+func (fix *toFix) getIntermediates(url string) ([]*x509.Certificate, *FixError) {
+	var icerts []*x509.Certificate
 	// PKCS#7 additions as (at time of writing) there is no standard Go PKCS#7
 	// implementation
 	r := urlReplacement(url)
 	if r != nil {
 		log.Printf("Replaced %s: %+v", url, r)
```
benlaurie: Probably don't need this logged (except perhaps at some very verbose level).

go/fixchain/fix.go
taknira: Done.

go/fixchain/fix.go
taknira: Done.

go/fixchain/fix.go
taknira: Removed.

go/fixchain/fix.go
```diff
@@ -71,45 +72,141 @@ func (fix *toFix) constructChain() ([][]*x509.Certificate, []*FixError) {
...
 }
 
-func (fix *toFix) augmentIntermediates(url string) *FixError {
+// toFix.augmentIntermediates() builds all possible chains that stem from the
+// given cert, and adds every certificate it finds in these chains to the pool
+// of intermediate certs in toFix.opts.  Every time a new certificate is added
+// to this pool, it tries to re-verify toFix.cert wrt toFix.roots.
+// If this verification is ever successful, toFix.augmentIntermediates() returns
+// the verified chains for toFix.cert wrt toFix.roots.  Also returned are any
+// errors that were encountered along the way.
```
benlaurie: Because of this, we may actually not build all possible chains for toFix.cert, which may mean we skip some intermediates.

Not in this version, but let's add a TODO to make it try to construct all possible chains and attempt to log the resulting intermediates.

taknira: I had the exact same thought :)

go/fixchain/fix_and_log_test.go
```diff
@@ -0,0 +1,311 @@
...
+		chain: []string{testLeaf, megaLeaf, googleLeaf, thawteIntermediate, comodoRoot},
+
+		function: "QueueChain",
+		expLoggedChains: [][]string{
+			{"Leaf", "Intermediate2", "Intermediate1", "CA"},
+		},
+		expectedErrs: []errorType{VerifyFailed},
+	},
+	// Tests that add chains to the FixAndLog using QueueAllCertsInChain()
+	{ // Full chain successfully logged.
```
benlaurie: Does go fmt really indent this comment differently from the following ones?

go/fixchain/fix_error.go
```diff
@@ -14,18 +17,20 @@ const (
...
 	FixFailed
+	PostFailed
 	LogPostFailed
 	VerifyFailed
 )
 
 // FixError is the struct with which errors in the fixing process are reported
 type FixError struct {
-	Type  errorType
+	Type  errorType           // The type of error
```
benlaurie: The comment adds no information.

go/fixchain/fix_error.go
```diff
@@ -14,18 +17,20 @@ const (
...
 	Cert  *x509.Certificate   // The supplied leaf certificate
 	Chain []*x509.Certificate // The supplied chain
 	URL   string              // URL, if a URL is involved
 	Bad   []byte              // The offending bytes, if applicable
-	Error error               // And the error
+	Error error               // The error
+	Code  int                 // The response code (only for use with LogPostFailed)
```
benlaurie: The HTTP response code from the log?

go/fixchain/fix_error.go
```diff
@@ -39,6 +44,8 @@ func (e FixError) TypeString() string {
 		return "CannotFetchURL"
 	case FixFailed:
 		return "FixFailed"
+	case PostFailed:
```
benlaurie: What post failed?

go/fixchain/fix_error.go
```diff
@@ -14,18 +17,20 @@ const (
...
 )
 
 // FixError is the struct with which errors in the fixing process are reported
 type FixError struct {
-	Type  errorType
+	Type  errorType           // The type of error
 	Cert  *x509.Certificate   // The supplied leaf certificate
 	Chain []*x509.Certificate // The supplied chain
 	URL   string              // URL, if a URL is involved
 	Bad   []byte              // The offending bytes, if applicable
```
benlaurie: Are these offending bytes always a cert? If so, say so...

go/fixchain/fix_error_test.go
```diff
@@ -37,9 +48,120 @@ func TestTypeString(t *testing.T) {
...
+				Chain: []*x509.Certificate{
+					GetTestCertificateFromPEM(t, googleLeaf),
+					GetTestCertificateFromPEM(t, thawteIntermediate),
+					GetTestCertificateFromPEM(t, verisignRoot),
+				},
+				URL:   "https://www.test.com",
+				Error: errors.New("Log Post Failed"),
+				Code:  404,
+			},
+			"{\"Type\":\"LogPostFailed\"," +
```
benlaurie: Using golden JSON makes us vulnerable to legitimate changes in the JSON marshaller.

go/fixchain/fix_error_test.go
```diff
@@ -37,9 +48,120 @@ func TestTypeString(t *testing.T) {
...
+		}
+	}
+}
+
+func TestDumpPEM(t *testing.T) {
+	dumpPEMTests := []string{googleLeaf}
+
+	for i, test := range dumpPEMTests {
+		cert := GetTestCertificateFromPEM(t, googleLeaf)
+		if got := dumpPEM(cert.Raw); got != test {
```
benlaurie: Same potential issue here.

go/fixchain/fixer.go
```diff
@@ -24,20 +24,19 @@ type Fixer struct {
...
 // QueueChain adds the given cert and chain to the queue to be fixed by the
-// fixer, with respect to the given roots.
+// fixer, with respect to the given roots.  Note: chain is expected to be in the
+// order of cert --> root.
 func (f *Fixer) QueueChain(cert *x509.Certificate, chain []*x509.Certificate, roots *x509.CertPool) {
+	dchain := newDedupedChain(chain)
 	f.toFix <- &toFix{
 		cert:  cert,
-		chain: newDedupedChain(chain),
+		chain: dchain,
```
benlaurie: Why did you refactor this?

go/fixchain/hash.go
```diff
@@ -0,0 +1,60 @@
...
+func hashChain(ch []*x509.Certificate) (hash [hashSize]byte) {
+	h := newHash()
+	for _, c := range ch {
+		h.Write(newHash().Sum(c.Raw))
+	}
+	copy(hash[:], h.Sum(nil))
+	return
+}
+
+func hashBag(chain []*x509.Certificate) [hashSize]byte {
```
benlaurie: Its a precondition that the chain is de-duped, I assume? I forget why we need a bag hash...

Anyway, assuming that it is a precondition, should we verify that? If it isn't, I am confused. :-)

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+// Limiter is an interface to allow different rate limiters to be used with the
+// Logger.
+type Limiter interface {
+	Wait()
+}
+
+// Logger contains methods to asynchronously log certificate chains to a
+// Certificate Transparency log and properties to store information about each
+// attempt that is made to post a certificate chain to said log.
+type Logger struct {
```
benlaurie: This struct looks like go fmt has not been applied.

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func (l *Logger) getRoots() *x509.CertPool {
+	rootsJSON, err := l.client.Get(l.url + "/ct/v1/get-roots")
+	if err != nil {
+		log.Fatalf("can't get roots from %s: %s", l.url, err)
+	}
+	defer rootsJSON.Body.Close()
+	if rootsJSON.StatusCode != 200 {
+		log.Fatalf("can't deal with status other than 200: %d", rootsJSON.StatusCode)
```
benlaurie: Might be helpful to also log the URL.

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	j, err := ioutil.ReadAll(rootsJSON.Body)
+	if err != nil {
+		log.Fatalf("can't read roots: %s", err)
```
benlaurie: Again, log URL.

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	type Certificates struct {
+		Certificates [][]byte
+	}
+	var certs Certificates
+	err = json.Unmarshal(j, &certs)
+	if err != nil {
+		log.Fatalf("can't parse json (%s): %s", err, j)
```
benlaurie: URL.

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	ret := x509.NewCertPool()
+	for i := 0; i < len(certs.Certificates); i++ {
+		r, err := x509.ParseCertificate(certs.Certificates[i])
+		switch err.(type) {
+		case nil, x509.NonFatalErrors:
+			// ignore
+		default:
+			log.Fatalf("can't parse certificate: %s %#v", err, certs.Certificates[i])
```
benlaurie: URL.

go/fixchain/fix_and_log_test.go
taknira: It really does.  It is weird.

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+		ret.AddCert(r)
+		log.Printf("Root %d: %s", i, r.Subject.CommonName)
```
benlaurie: Squash this?

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+type toPost struct {
+	chain   []*x509.Certificate
+	retries uint8
+}
+
+func (l *Logger) postToLog(p *toPost) {
+	// Add post to toPost channel in a separate goroutine to avoid deadlock
+	// between postServers during retries.
```
benlaurie: Err? How can this deadlock arise?

go/fixchain/logger.go
```diff
@@ -0,0 +1,232 @@
...
+	l.wg.Add(2)
```
benlaurie: 2? Why?

go/fixchain/logger_test.go
```diff
@@ -0,0 +1,208 @@
...
+		{
+			testC,
+			false,
+		},
+	}
+
+	l := &Logger{postCertCache: newLockedMap()}
+	l.postCertCache.set(hash(GetTestCertificateFromPEM(t, googleLeaf)), true)
+	l.postCertCache.set(hash(GetTestCertificateFromPEM(t, megaLeaf)), true)
+	l.postCertCache.set(hash(GetTestCertificateFromPEM(t, testLeaf)), false)
```
benlaurie: A little weird - we never set anything to false in this map, really, do we?

go/fixchain/main/fixchain.go
```diff
@@ -0,0 +1,110 @@
...
+					err, derBytes)
+			}
+
+			chain = append(chain, cert)
+		}
+		fl.QueueAllCertsInChain(chain)
+	}
+}
+
+func contentStore(baseDir string, subDir string, content []byte) {
```
benlaurie: What is this?

go/fixchain/post.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/google/certificate-transparency/go/x509"
+)
+
+// Post attempts to post the given chain to the Certificate Transparency log at
+// the given url, using the given http client.  Post returns a FixError if it
+// is unable to post the chain either because client.Post() failed, or the http
+// response code returned was not 200.  It is up to the caller to handle such
+// errors appropriately.
+func Post(chain []*x509.Certificate, client *http.Client, url string) *FixError {
```
benlaurie: Perhaps a more descriptive name than just `Post`?

go/fixchain/test_round_trippers.go
```diff
@@ -0,0 +1,175 @@
...
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+type fixTestRoundTripper struct{}
+
+func (rt fixTestRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
```
benlaurie: This code looks like a subset of the other round tripper. Why is it not shared?

go/fixchain/url_cache.go
```diff
@@ -20,7 +43,7 @@ type urlCache struct {
 }
 
 func (u *urlCache) getURL(url string) ([]byte, error) {
-	r, ok := u.cache[url]
+	r, ok := u.cache.get(url)
```
benlaurie: FWIW, I did not make this locked in the first place because even if locked, a miss can still lead to a double fetch, unless you hold the lock for the whole fetch, which would be a bad idea.

That said, I suspect you need write locking to be thread safe (I forget!) - but it not, not sure there's much point in locking.

go/fixchain/fix_error.go
taknira: The way we currently use this field it is.  However, the whole FixError struct and its fields are exported, so if someone wanted to randomly use our struct and put something other than a cert in to these bytes, they could.

go/fixchain/fix_error.go
taknira: PostFailed is used when the call to http.Client.Post() returns an error.  LogPostFailed is used when the response from the log contains a status code other than 200.

go/fixchain/fix_error.go
pphaneuf: This smells like you're not applying [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)... :wink:

go/fixchain/logger.go
taknira: It has...

gofmt doesn't format vertically mind you, it keeps the given field groupings and just aligns the types in each group.

go/fixchain/fix_error.go
benlaurie: The idea is you modify the code to explain this (presumably in the comment where the error is defined).

go/fixchain/fix_error.go
benlaurie: Meh. Anyone can abuse any struct they want. Probably best to document it is a cert. Especially since we dump it as one...

benlaurie: @pphaneuf it is used. :-)

benlaurie: And, indeed, useful.

pphaneuf: What I mean is that "some day, someone could put something that's not a cert", *that* isn't used, and likely you aren't going to need it (YAGNI). So having it be a narrower interface (a cert) is a good idea, especially as it's usually easier to widen an interface than to retroactively try to narrow it back down!

go/fixchain/logger.go
taknira: Suppose you have just one postServer, PS, and it's blocked waiting for something to come in on the toPost chan.
A chain in toPost form, C, gets added to the toPost chan.  

Looking at the code in postServer() here's what happens next: 
-  PS takes that toPost from the chan.
-  PS calls l.postChain(toPost) AND DOES NOT RETURN AND WAIT TO TAKE ANOTHER toPost FROM THE CHAN UNTIL THIS HAS RETURNED (that part is important).
- l.postChain begins execution, but, oh no!, the post of C failed (for some reason).
- l.postChain decrements the number of retries and calls l.postToLog(C) to queue up the toPost to try again.
- l.postToLog tries to put C on the chan, and blocks until a postServer takes it off the chan (because that's unbuffered chans for you).

But oh no!  Our one and only postServer is still waiting for postChain to return, and is not taking any more toPosts off chan until that's done!

DEADLOCK.

taknira: I have a test that encountered this exact situation :)

go/fixchain/logger.go
taknira: You add one because you're putting another active request out there.
You add one because you're doing it in a new goroutine and you want to make sure that goroutine gets to complete its business.
Thinking about it, It may be that the second one is not necessary (as long as the goroutine adding the request does not call done either) as the one added for the new request will not be decremented until the goroutine adding it has successfully added it and it has been posted...

go/fixchain/logger_test.go
taknira: We don't in our code no, but that is how one would 'remove' an item from the postCertCache if they were going to.  So I was testing that IsPosted(item) would return false in the case that either the item is not in the cache at all, or it has been 'removed' by being set to false.

go/fixchain/main/fixchain.go
taknira: This is a function to save the errors that are spat out by the FixAndLog to a directory.  Note:  this is the simple open source main package.  People using this would most likely alter contentStore as they wish to store the errors wherever/however they need to.  Both logStringErrors() and logJSONErrors() use this function as a way of storing the FixErrors.

go/fixchain/url_cache.go
taknira: You do need locking to be thread safe, as multiple goroutines will be potentially reading from and writing to this one cache at the same time.  My tests went into a frenzy of panicking goroutines once or twice due to this :)

go/fixchain/hash.go
taknira: It's not a precondition, no.  The point of hashBag is just to hash a given set of certificates, whether it contains duplictates or not.  It was used for caching in the fixer, but I've since removed that, and am currently undecided about where and how it should be put back, if at all.  If we were to put it back it'd probably be at the FixAndLog level.

go/fixchain/fix_error.go
taknira: Either way, this is just a commenting issue, as this field is currently used specifically when a cert can't be parsed, so the data type will still have to be []byte

go/fixchain/post.go
taknira: I actually quite liked the simplicity of it.  The asynchronous Logger contains a lot of functions containing some combination of the words 'post', 'chain', and 'log', and this being the base synchronous function that actually does do the post, it seemed fitting to give it a simple name.  Also, I liked the duality of Fix and Post being our two exported synchronous functions.  But I can call it  'PostChainToLog' or something similar if you really insist :P

taknira: How would you feel about 'Log()'?  Then we'd have a Fixer and a Logger, and Fix() and Log() :D

benlaurie: I'd be OK with it being called Post if it were a member of Logger. Why isn't it?

taknira: To separate application logic from concurrency.  Post() contains the application logic, Logger contains the concurrency.  All the fixing application logic is separated from the Fixer too.

go/fixchain/fix.go
taknira: TODO added

go/fixchain/fix_error.go
taknira: Comment added

go/fixchain/fix_error.go
taknira: Commented to indicate they are certificate bytes

go/fixchain/fixer.go
taknira: No idea :)  I'd imagine I considered putting something else in there then backtracked but never made it all the way back.  Fixed now.

go/fixchain/hash.go
taknira: In the original code, hashBag was for caching in the Fixer.  It was used to hash a chain that had had every cert in it queued for fixing and logging.  I removed this because it didn't make sense for the new Fixer, but it does make sense for the FixAndLog.  I have just added caching to the FixAndLog using hashBag to calculate the chain hashes.  In our use case the chains should always be deduped, but I have not put this functionality in hashBag because in the one place it is called the chain is deduped right before calling hashBag (and the deduped chain is then used after calling hashBag too).  In general (and in the tests) it does not assume that hashBag is only for deduped chains, and hashing two chains containing the same certs, but where one has some additional duplicates, does not produce the same hash.

go/fixchain/logger.go
taknira: Done

go/fixchain/logger.go
taknira: Done

go/fixchain/logger.go
taknira: Done

go/fixchain/logger.go
taknira: Done

go/fixchain/logger.go
taknira: Done

go/fixchain/logger.go
taknira: Comment added to explain

go/fixchain/logger.go
taknira: After thinking this through, I have changed this back to Add(1) and removed the Done() inside the goroutine.

go/fixchain/main/fixchain.go
taknira: Comment added

go/fixchain/test_round_trippers.go
taknira: It is now shared :)

go/fixchain/fix_error.go
taknira: Removed

go/fixchain/fix_error.go
taknira: Commented.

go/fixchain/fix_error_test.go
taknira: Changed to marshal then unmarshal and check result matches the original

Note:  This involved adding Equal() and Unmarshal() functions in fix_error.go that need reviewing.

go/fixchain/fix_error_test.go
taknira: Changed.

go/fixchain/fix.go
```diff
@@ -71,45 +72,144 @@ func (fix *toFix) constructChain() ([][]*x509.Certificate, []*FixError) {
...
 	// implementation
 	r := urlReplacement(url)
 	if r != nil {
-		log.Printf("Replaced %s: %+v", url, r)
 		for _, c := range r {
-			fix.opts.Intermediates.AddCert(c)
+			icerts = append(icerts, c)
 		}
-		return nil
+		return icerts, nil
```
ThomasHabets: not just return r, nil?

go/fixchain/fix_error.go
```diff
@@ -47,3 +90,130 @@ func (e FixError) TypeString() string {
...
+	case "FixFailed":
+		ferr.Type = FixFailed
+	case "PostFailed":
+		ferr.Type = PostFailed
+	case "LogPostFailed":
+		ferr.Type = LogPostFailed
+	case "VerifyFailed":
+		ferr.Type = VerifyFailed
+	default:
+		return nil, errors.New("Cannot parse FixError Type")
```
ThomasHabets: Errors start with lower case, because they stack.

go/fixchain/fix_error.go
```diff
@@ -47,3 +90,130 @@ func (e FixError) TypeString() string {
...
+	}
+
+	if u.Cert != nil {
+		cert, err := x509.ParseCertificate(u.Cert)
+		if err != nil {
+			return nil, fmt.Errorf("Cannot parse FixError Cert: %s", err)
```
ThomasHabets: same

go/fixchain/fix_error.go
```diff
@@ -47,3 +90,130 @@ func (e FixError) TypeString() string {
...
+		}
+		ferr.Cert = cert
+	}
+
+	for _, c := range u.Chain {
+		cert, err := x509.ParseCertificate(c)
+		if err != nil {
+			return nil, fmt.Errorf("Cannot parse FixError Chain: %s", err)
```
ThomasHabets: same

go/fixchain/logger.go
```diff
@@ -0,0 +1,250 @@
...
+// attempt that is made to post a certificate chain to said log.
+type Logger struct {
+	url    string
+	client *http.Client
+	roots  *x509.CertPool
+	toPost chan *toPost
+	errors chan<- *FixError
+
+	active uint32
+	// Counters may not be entirely accurate due to non-atomicity.
```
ThomasHabets: Do you mean this code breaks when running with gotsan? Don't make gotsan complain.

go/fixchain/logger_test.go
ThomasHabets: you wouldn't use "delete"?

go/fixchain/main/fixchain.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/google/certificate-transparency/go/fixchain/ratelimiter"
+	"github.com/google/certificate-transparency/go/x509"
+)
+
+// Assumes chains to be stores in a file in JSON encoded with the certificates
+// in DER format.
+func processChains(file string, fl *fixchain.FixAndLog) {
+	f, err := os.Open(file)
+	if err != nil {
+		log.Fatalf("Can't open %s: %s", f, err)
```
ThomasHabets: Consider using %q for the filename

go/fixchain/post.go
```diff
@@ -0,0 +1,59 @@
...
+			Type:  PostFailed,
+			Chain: chain,
+			Error: fmt.Errorf("Can't post: %s", err),
+		}
+	}
+
+	defer resp.Body.Close()
+	jo, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatalf("Can't read response: %s", err)
```
ThomasHabets: while the previous log.Fatal looks like a "can't happen, panic!". This one does not. Returning error seems better, no?

go/fixchain/fix_and_log.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// NewFixAndLog creates an object that will asynchronously fix any chains that
+// are added to its queue, and then log them to the Certificate Transparency log
+// found at the given url.  Any errors encountered along the way are pushed to
+// the given errors channel.
+func NewFixAndLog(workerCount int, errors chan<- *FixError, client *http.Client, logURL string, limiter Limiter, logStats bool) *FixAndLog {
+	chains := make(chan []*x509.Certificate)
+	fl := &FixAndLog{
+		fixer:  NewFixer(workerCount, chains, errors, client, logStats),
```
ThomasHabets: Does this mean that the API has two levels of async workers? Will the `Fixer` be used outside of here?

This asyncness makes for complicated code, for example above in `Wait()` where it's impossible to remember what order the four lines need to be in, without spending a lot of effort reading and thinking. Async code tends to be much less quick to read that way.

taknira: It depends what you mean by 'levels'.  FixAndLog does not wrap more asyncness around the Fixer and Logger.

The Fixer has its async workers, and the Logger has its async workers.  FixAndLog starts them both up, and then provides a bridge between the two - it takes any fixed chains output from the Fixer and passes them to the Logger for logging.  FixAndLog is just a wrapper for this process.

ThomasHabets: And it makes more sense to glue these async things together than to make pairs of sync versions of them, thus moving the asyncness up a level?

I mean I can see that it could. But it'd certainly make for easier to read/debug code if they were simple calls, so coupling their "scaling" would have benefits if it's feasible.

go/fixchain/fix.go
taknira: Changed.

go/fixchain/fix_error.go
taknira: Done

go/fixchain/fix_error.go
taknira: Done

go/fixchain/fix_error.go
taknira: Done

go/fixchain/main/fixchain.go
taknira: Done

go/fixchain/post.go
taknira: Agreed and changed.

go/fixchain/fix_and_log.go
taknira: I have now added a second workerCount parameter to the NewFixAndLog() call, so there is one for each of the fixer and logger.

One of the reasons for keeping the Fixer and Logger so separate is that when a chain is verified, it can result in multiple chains.  We want to try to log each of these chains.  Thus, the workload for the logger could be significantly higher than the workload for the fixer.  Being able to specify separate worker counts for the two tools means that we can increase the number of log workers if required, without having extra fix workers sitting around idle.  This is one of the reasons for gluing them together in this way, rather than coupling their scaling by having one call to the synchronous Fix() followed by one or more calls to the synchronous Post() within each goroutine.

Another reason for building a separate Fixer and Logger and having them as part of the external API is that we may want to do something else with the chains after they have been fixed.  This way we can take the fixer and run it on its own, and do what we want with the output.  As we have the Fixer and Logger built, I thought we may as well use them, within (what I thought was) a relatively simple FixAndLog wrapper :)

go/fixchain/fix_and_log.go
```diff
@@ -0,0 +1,99 @@
...
+	logger *Logger
+	wg     sync.WaitGroup
+
+	done        *lockedMap // Cache of chains that QueueAllCertsInChain() has already been called on.
+	alreadyDone uint
+}
+
+// QueueAllCertsInChain adds every cert in the chain and the chain to the queue
+// to be fixed and logged.  Note: chain is expected to be in the order of
+// cert --> root.
```
benlaurie: Is it? Isn't the whole point that we don't care what order they are in?

go/fixchain/logger.go
taknira: It did break when I ran my tests with the -race flag.  I have now made all of the counters be incremented atomically.

go/fixchain/hash.go
benlaurie: At least comment that it would hash differently for chains with dupes, but not for chains with different orders (which is why it is a bag rather than a set).

Though I question why we would ever want a bag rather than a set!

go/fixchain/fix_and_log.go
taknira: You are right, for the QueueAllCertsInChain it doesn't matter.  For QueueChain it matters that the cert the chain is for comes first for the check to see if it has already been posted, but here as we are fix-then-logging every cert in the chain it doesn't matter.  Comment removed.

go/fixchain/post.go
benlaurie: Then I think it really does need a better name.

benlaurie:
LGTM apart from any outstanding minor comments.

go/fixchain/logger_test.go
taknira: Yeah I was thinking about that, that would be a better way to remove things.  But as we never remove anything from them we never implemented a method to remove things from the cache that uses delete.

go/fixchain/post.go
taknira: Changed to PostChainToLog()

go/fixchain/hash.go
taknira: Added comment.  But it's true, for our purposes I suppose we wouldn't.

go/fixchain/hash.go
```diff
@@ -25,6 +25,10 @@ func hashChain(ch []*x509.Certificate) (hash [hashSize]byte) {
 	return
 }
 
+// hashBag hashes all of the certs in the chain, irrespective of their order.
+// Two chains containing the same certs in different orders will result in the
+// same hash.  Two chains containing the same certs but of different lengths
```
benlaurie: This not a technically accurate description (i.e. they could be the same length, but still differ). But I won't make you change it, I get the drift.

taknira: The Travis build failed so needs running again anyway, so figured I might as well :)

PR #1129, opened 2016-03-10
[REFERENCE] Add include_scts support to get-entries.

AlCutter:
Adds support for the `include_scts` parameter of the get-entries handler of the reference CT implementation.  Needed for #1128.

### DIFF
--- a/cpp/server/handler.cc
+++ b/cpp/server/handler.cc
@@ -235,6 +235,18 @@ int GetIntParam(const multimap<string, string>& query, const string& param) {
 }
 
 
+// Returns true iff query includes a parameter named |param| whose value is
+// "true", and false otherwise.
+bool GetBoolParam(const multimap<string, string>& query, const string& param) {
+  string value;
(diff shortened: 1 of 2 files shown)


cpp/server/handler.cc
```diff
@@ -294,6 +306,8 @@ void HttpHandler::GetEntries(evhttp_request* req) const {
                      "Missing or invalid \"end\" parameter.");
   }
 
+  const bool include_scts(GetBoolParam(query, "include_scts"));
+
```
pphaneuf: Maybe put that right next to line 321? And lift the comment too?

https://github.com/google/certificate-transparency/blob/master/cpp/server/handler.cc#L193

pphaneuf:
LGTM, with just a minor adjustment.

cpp/server/handler.cc
AlCutter: Done, squashed it in.

PR #1133, opened 2016-03-20
Tweak signature algorithm mismatch testing certificate

estark37:
The certificate used for testing mismatched signature algorithms had a
ASN1 universal primitive with tag number 469794384. I'm not sure how
that got there in the first place, but large tag numbers are no longer
accepted in BoringSSL as of
https://boringssl-review.googlesource.com/#/c/7238/.

This changes the tag number to 100, since presumably all that's
important is that the signature algorithm doesn't match the signer.

### DIFF
--- a/cpp/log/cert_test.cc
+++ b/cpp/log/cert_test.cc
@@ -83,7 +83,45 @@ static const char kInvalidCertString[] =
     "-----BEGIN CERTIFICATE-----\ninvalid"
     "\n-----END CERTIFICATE-----\n";
 
+// This certificate is the same as |kMatchingSigAlgsCertString| below,
+// but with the unsigned signatureAlgorithm parameter changed so as to
+// not match the one in the signed part of the certificate.
 static const char kMismatchingSigAlgsCertString[] =
+    "-----BEGIN CERTIFICATE-----\n"
+    "MIIFVjCCBD2gAwIBAgIQGERW2P+5Xt15QiwP1NRmtTANBgkqhkiG9w0BAQsFADCB\n"
+    "kDELMAkGA1UEBhMCR0IxGzAZBgNVBAgTEkdyZWF0ZXIgTWFuY2hlc3RlcjEQMA4G\n"
+    "A1UEBxMHU2FsZm9yZDEaMBgGA1UEChMRQ09NT0RPIENBIExpbWl0ZWQxNjA0BgNV\n"
+    "BAMTLUNPTU9ETyBSU0EgRG9tYWluIFZhbGlkYXRpb24gU2VjdXJlIFNlcnZlciBD\n"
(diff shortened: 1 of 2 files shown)


eranmes:
How is BoringSSL going to reject such certificates? That particular certificate, IIRC was actually observed in the wild. I'm curious how BoringSSL will reject it.

eranmes:
That is, how existing code will behave differently.

estark37:
Oh, hmm, that was an actual certificate? (I assumed it was just manipulated in some weird artificial way to create a test input.) Previous BoringSSL will happily load it, but latest BoringSSL will refuse to parse it, and the test will fail here: https://github.com/google/certificate-transparency/blob/master/cpp/log/cert.cc#L357

cc @davidben in case I'm misunderstanding the BoringSSL behavior.

davidben:
Yeah, it will fail to parse. I'm not sure how that certificate could possibly be seen in the wild. It must have been an impressively buggy signer. :-) Universal tags are INTEGER, OCTET STRING, etc. You're not supposed to just go make them up.

estark37:
Friendly ping :)

cpp/log/cert_test.cc
```diff
@@ -85,6 +85,40 @@ static const char kInvalidCertString[] =
 
 static const char kMismatchingSigAlgsCertString[] =
```
eranmes: Please add a comment saying this certificate was derived from kIllegalSigAlgParameterCertString with the unsigned parts changed.

cpp/log/cert_test.cc
```diff
@@ -449,6 +484,13 @@ TEST_F(CertTest, SignatureAlgorithmMatches) {
   EXPECT_FALSE(mismatched_algs.IsSignedBy(issuer).ValueOrDie());
 }
 
+#if defined(OPENSSL_IS_BORINGSSL) && defined(BORINGSSL_201603)
```
eranmes: How about making this test always run, but EXPECT_FALSE with that particular BoringSSL and EXPECT_TRUE with all others versions of OpenSSL/BoringSSL?

cpp/log/cert_test.cc
estark37: I added some comments but it made more sense to me to say how kMismatchingSigAlgsCertString differs from kMatchingSigAlgsCertString and how kIllegalSigAlgParameterCertString differs from kMismatchingSigAlgsCertString... which is similar to but not exactly what you asked for. LMK what you think.

estark37:
Thanks @eranmes. Addressed your comments and fixed tests that were failing locally (sorry about that). Please hold off on re-reviewing until the travis build finishes; I'll ping you when it's green.

cpp/log/cert_test.cc
estark37: Done

eranmes:
LGTM

PR #1137, opened 2016-03-29
Update README URLs based on HTTP redirects

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein

### Corrected URLs 
Was | Now 
--- | --- 
https://code.google.com/p/googlemock/ | https://github.com/google/googlemock

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ released, so this means hand building the `OpenSSL_1_0_2-stable`
 branch from the
 [OpenSSL git repository](https://www.openssl.org/source/repos.html).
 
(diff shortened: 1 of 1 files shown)


pphaneuf:
LGTM

ReadmeCritic:
@pphaneuf actually maybe the link and text should update to https://github.com/google/googletest

pphaneuf:
Possibly, although we were already pulling from that, so this here is, strictly speaking, correct. :wink: 

We'll look into updating further.

PR #1138, opened 2016-03-30
Python: Fix domain name reversing in cert_desc

eranmes:
The cert_desc module reversed every domain name it found in any component in the subject and issuer fields. 
Only the CN should be reversed as it's indexed that way.


phad:
Looks good, thanks for fixing.

PR #1141, opened 2016-04-04
Preloader

taknira:
Two commits:  One to improve logging, and one to fix the excessive memory usage of the new preloader.

### DIFF
--- a/go/fixchain/fix_and_log.go
+++ b/go/fixchain/fix_and_log.go
@@ -4,6 +4,7 @@ import (
 	"log"
 	"net/http"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/certificate-transparency/go/x509"
@@ -16,19 +17,22 @@ type FixAndLog struct {
 	logger *Logger
 	wg     sync.WaitGroup
 
+	queued      uint32     // Whole chains queued - before checking cache.
 	done        *lockedMap // Cache of chains that QueueAllCertsInChain() has already been called on.
-	alreadyDone uint
+	alreadyDone uint32
+	chainsSent  uint32
 }
 
 // QueueAllCertsInChain adds every cert in the chain and the chain to the queue
 // to be fixed and logged.
 func (fl *FixAndLog) QueueAllCertsInChain(chain []*x509.Certificate) {
 	if chain != nil {
+		atomic.AddUint32(&fl.queued, 1)
 		dchain := newDedupedChain(chain)
 		// Caching check
 		h := hashBag(dchain.certs)
 		if fl.done.get(h) {
-			fl.alreadyDone++
+			atomic.AddUint32(&fl.alreadyDone, 1)
 			return
 		}
 		fl.done.set(h, true)
@@ -38,6 +42,7 @@ func (fl *FixAndLog) QueueAllCertsInChain(chain []*x509.Certificate) {
 				continue
 			}
 			fl.fixer.QueueChain(cert, dchain.certs, fl.logger.RootCerts())
+			atomic.AddUint32(&fl.chainsSent, 1)
 		}
 	}
 }
@@ -52,6 +57,7 @@ func (fl *FixAndLog) QueueChain(chain []*x509.Certificate) {
 			return
 		}
 		fl.fixer.QueueChain(chain[0], chain, fl.logger.RootCerts())
+		atomic.AddUint32(&fl.chainsSent, 1)
 	}
 }
 
@@ -89,7 +95,7 @@ func NewFixAndLog(fixerWorkerCount int, loggerWorkerCount int, errors chan<- *Fi
 		t := time.NewTicker(time.Second)
 		go func() {
 			for _ = range t.C {
-				log.Printf("fix-then-log whole chains: %d already done", fl.alreadyDone)
+				log.Printf("fix-then-log: %d whole chains queued, %d whole chains already done, %d chains sent", fl.queued, fl.alreadyDone, fl.chainsSent)
 			}
 		}()
 	}
--- a/go/fixchain/logger.go
+++ b/go/fixchain/logger.go
@@ -79,6 +79,15 @@ func (l *Logger) QueueChain(chain []*x509.Certificate) {
 	l.postChainCache.set(h, true)
 
 	p := &toPost{chain: chain, retries: 5}
+	// postToLog() is called during the initial queueing of chains, rather than
+	// its asynchronous couterpart, asyncPostToLog(), to avoid spinning up an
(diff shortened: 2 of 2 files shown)


go/fixchain/fix_and_log.go
```diff
@@ -89,7 +95,7 @@ func NewFixAndLog(fixerWorkerCount int, loggerWorkerCount int, errors chan<- *Fi
 		t := time.NewTicker(time.Second)
 		go func() {
 			for _ = range t.C {
-				log.Printf("fix-then-log whole chains: %d already done", fl.alreadyDone)
+				log.Printf("fix-then-log: %d whole chains queued, %d whole chains already done, %d chains sent", fl.queued, fl.alreadyDone, fl.chainsSent)
```
benlaurie: random use of commas.

Also, must be some way to avoid saying "chains" and "whole chains" every time.

go/fixchain/logger.go
```diff
@@ -136,32 +136,8 @@ type toPost struct {
...
-	//   try to post it again.
-	// - l.postToLog() tries to put the toPost on the toPost chan, and blocks
-	//   until a postServer takes it off the chan.
-	// But the one and only postServer is still waiting for l.postChain (and
-	// therefore l.postToLog) to return, and will not go to take another toPost
-	// off the toPost chan until that happens.
-	// Thus, deadlock.
-	//
-	// Similar situations with multiple postServers can easily be imagined.
 	l.wg.Add(1)
```
benlaurie: Surely this waitgroup add should go, too?

go/fixchain/fix_and_log.go
taknira: What about the use of commas is random?

go/fixchain/logger.go
taknira: No, this wg.Add() isn't because of the goroutine.  This Add() represents the addition of another active request.

go/fixchain/logger.go
```diff
@@ -182,7 +158,38 @@ func (l *Logger) postChain(p *toPost) {
...
+				// not go to take another toPost off the toPost chan until that
+				// happens.
+				// Thus, deadlock.
+				//
+				// Similar situations with multiple postServers can easily be
+				// imagined.
+				l.wg.Add(1)
+				go func() {
+					l.postToLog(p)
+					l.wg.Done()
```
benlaurie: Seems to me you should drop this l.wg.Done() and also the l.wg.Add(1) in postToLog().

Also, why does this make any difference to memory consumption? And if it does, you need to add a note to prevent refactoring back to what it was.

go/fixchain/fix_and_log.go
benlaurie: Oh, I see how I'm supposed to read it. Take it back, its fine.

go/fixchain/logger.go
taknira: The answers to both of your comments lie in the fact that logToPost is not only used in retries, but also when initially queuing chains for logging :)

You need the Add() in postToLog() so that the wg is incremented each time QueueChain() is used to add a chain to the queue of things to post.

This makes a difference to memory consumption because, as it was before, each time a chain was sent to the logger to be queued, QueueChain would call logToPost, which, instead of blocking while waiting for one of the workers to take the chain off the channel, would spin up a new goroutine to block, leaving logToPost and QueueChain to return immediately.  This was resulting in LOADS (10s to 100s of thousands) of new goroutines sitting waiting for a free logger,  each holding a chain.  The latest test run with these changes is showing only ~1.2GB of memory rather than the previous 70GB+ :)

go/fixchain/fix_and_log.go
taknira: :)

go/fixchain/logger.go
taknira: Done.

benlaurie: OK, it seems to me there's two things going on here:

1. Initial queuing doesn't need to be put into a goroutine, and doing so causes problems because you then get a huge backlog of goroutines.

2. There's a race condition, where the Add(1) in postToLog happens after someone has successfully Wait()ed for that wg, which causes you to Add(1) before the go routine, and then requires a Done() inside the go routine to cancel that Add() in postToLog().

This second point and the fact that you end up repeating this block of code strongly suggests that what you really want is postToLog() (sync) and queueToLog() or asyncPostToLog() or somesuch (async, obvs).

taknira: 1. Yes. Exactly.
2. Yes again, although I would say 'There's a race condition, where the Add(1) in postToLog happens after someone has successfully Wait()ed for that wg, which causes you to Add(1) before the go routine, and then requires a Done() inside the go routine to cancel the Add(1) before the goroutine.'  We need to add 1 to the wg as we are adding a new active request to the queue, and the Add(1) in postToLog() is what I consider to be that, as it's called right before we add the request to the queue.

But yes, if the repeated lines of code go in their own little async function, and we don't call postToLog in that function at all, then we can achieve all of these things in a much cleaner way!

Done, and it is much prettier!

go/fixchain/logger.go
```diff
@@ -182,7 +174,35 @@ func (l *Logger) postChain(p *toPost) {
...
+				// - l.postToLog() tries to put the toPost on the toPost chan,
+				//   and blocks until a postServer takes it off the chan.
+				// But the one and only postServer is still waiting for
+				// l.postChain (and therefore l.postToLog) to return, and will
+				// not go to take another toPost off the toPost chan until that
+				// happens.
+				// Thus, deadlock.
+				//
+				// Similar situations with multiple postServers can easily be
+				// imagined.
```
benlaurie: I would move this comment to the function.

taknira: Assumption: 'the function' = asyncPostToLog()

Even though it's the reason that, in this situation, that function should be used over another function?  What if we were using asyncPostToLog() somewhere else too, for some other purpose?  (I know we don't, but I'm mentioning it as an example of when having this comment tied to the function could be confusing.)

benlaurie: I would say that when you use it somewhere else you need to update the comment (note that it is local to this module).

benlaurie: Or just change that comment to be an example of why it might be used.

taknira: Okay.  Should the comment in QueueChain() about why we're using postToLog instead of asyncPostToLog in that situation be moved to postToLog instead too then?  It's the same type of comment, but with the functions reversed, and a different reason.

taknira: Moved both comments :)

PR #1142, opened 2016-04-05
Work with boringssl

AlCutter:


Martin2112:
Looks like this switches to BoringSSL by default? Not that we shouldn't do that but I was hoping we could build with both on Travis to help spot inconsistencies.

Should it default to OpenSSL not BoringSSL?

cpp/log/cert_test.cc
```diff
@@ -489,7 +489,8 @@ TEST_F(CertTest, SignatureAlgorithmMatches) {
 
 TEST_F(CertTest, IllegalSignatureAlgorithmParameter) {
   Cert cert(kIllegalSigAlgParameterCertString);
-  #if defined(OPENSSL_IS_BORINGSSL) && defined(BORINGSSL_201603)
+  #if defined(OPENSSL_IS_BORINGSSL) && \
+              (defined(BORINGSSL_201603) || defined(BORINGSSL_201512))
```
Martin2112: Not ideal as discussed but we can live with it for now.

PR #1145, opened 2016-04-06
Python: Print log freezing info.

eranmes:
In preparation for freezing Certly's log, support printing that information if it is present in the log list.

### DIFF
--- a/python/utilities/log_list/print_log_list.py
+++ b/python/utilities/log_list/print_log_list.py
@@ -6,6 +6,7 @@ import hashlib
 import json
 import os
 import sys
+import time
 
 import gflags
 import jsonschema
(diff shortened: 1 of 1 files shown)


python/utilities/log_list/print_log_list.py
```diff
@@ -39,9 +40,9 @@
 
 def is_log_list_valid(json_log_list, schema_file):
     try:
+        schema_data = json.load(open(schema_file, "rb"))
         jsonschema.validate(
-            json_log_list,
-            json.load(open(schema_file, "rb")))
+            json_log_list, schema_data)
```
pphaneuf: I kind of like the old version better here, especially considering Python's mad function scoping... Not a big deal, if you prefer this, can keep it like this.

pphaneuf:
LGTM as-is, or even better reverting that first diff chunk.

python/utilities/log_list/print_log_list.py
eranmes: Will remove, only did this for debugging purposes.

PR #1146, opened 2016-04-07
Add support for DSA signature checking under BoringSSL + tests

Martin2112:
BoringSSL does not support DSA signatures via x509_verify so this adds extra handling for that case if x509_verify fails with an unsupported algorithm.

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -71,15 +71,12 @@ StatusOr<bool> check_dsa_signature(const X509* cert,
                           "cert is missing a signature");
   }
 
-  // If the following breaks at some future point it might need to be
-  // replaced by EVP_PKEY_base_id(issuer_key)
-  if (EVP_PKEY_type(issuer_key->type) != EVP_PKEY_DSA ||
-      X509_get_signature_type(cert) != EVP_PKEY_DSA) {
+  if (EVP_PKEY_type(issuer_key->type) != EVP_PKEY_DSA) {
     return ::util::Status(Code::FAILED_PRECONDITION,
-                          "cert or issuer does not have a DSA public key");
+                          "issuer does not have a DSA public key");
   }
 
-  const int alg_nid = OBJ_obj2nid(sig->algorithm);
+  const int alg_nid = X509_get_signature_nid(cert);
 
   int digest_nid;
   // We need the nid of the digest so we can create an EVP_MD for it later.
@@ -89,16 +86,17 @@ StatusOr<bool> check_dsa_signature(const X509* cert,
   }
 
   // Get the DER encoded certificate info from the cert
-  std::unique_ptr<unsigned char*> der_buf_ptr(nullptr);
-  const int der_length = i2d_X509_CINF(cert_info, der_buf_ptr.get());
+  unsigned char* der_buf(nullptr);
+  const int der_length = i2d_X509_CINF(cert_info, &der_buf);
   if (der_length < 0) {
-    // What does this return value mean? Let's assume it means the cert
-    // is bad until proven otherwise.
+    // Failed to decode. Several possible reasons but we will just reject
+    // the input rather than trying to interpret the cause
     LOG(WARNING) << "Failed to serialize the CINF component";
     LOG_OPENSSL_ERRORS(WARNING);
     return ::util::Status(Code::INVALID_ARGUMENT,
                           "failed to serialize cert info");
   }
+  std::unique_ptr<unsigned char> der_buf_ptr(der_buf);
 
   // If the key is missing parameters we don't accept it. This is allowed
   // by RFC 3279 but we have not found any examples in the wild where it's
@@ -134,7 +132,7 @@ StatusOr<bool> check_dsa_signature(const X509* cert,
     return ::util::Status(Code::INTERNAL, "failed to check DSA signature");
   }
 
-  return out_valid != 0;
+  return out_valid == 1;
 }
 
 #endif
@@ -476,12 +474,16 @@ StatusOr<bool> Cert::IsSignedBy(const Cert& issuer) const {
   const StatusOr<bool> is_valid_dsa_sig =
       check_dsa_signature(x509_.get(), issuer_key.get());
 
-  if (is_valid_dsa_sig.ok() && is_valid_dsa_sig.ValueOrDie()) {
-    ClearOpenSSLErrors();
-    return true;
+  if (is_valid_dsa_sig.ok()) {
+    if (is_valid_dsa_sig.ValueOrDie()) {
+      ClearOpenSSLErrors();
+      return true;
+    } else {
+      // Ensure we return the same status as we'd have got from calling
+      // IsValidSignatureChain() under OpenSSL when the signature is not valid.
+      return util::Status(Code::INVALID_ARGUMENT, "invalid certificate chain");
+    }
   }
-
-  LOG(ERROR) << "check_dsa_signature returned: " << is_valid_dsa_sig.status();
 #endif
 
   unsigned long err = ERR_peek_last_error();
@@ -532,8 +534,8 @@ util::Status Cert::DerEncoding(string* result) const {
   int der_length = i2d_X509(x509_.get(), &der_buf);
 
   if (der_length < 0) {
-    // What does this return value mean? Let's assume it means the cert
-    // is bad until proven otherwise.
+    // Failed to decode. Several possible reasons but we will just reject
+    // the input rather than trying to interpret the cause
     LOG(WARNING) << "Failed to serialize cert";
     LOG_OPENSSL_ERRORS(WARNING);
     return util::Status(Code::INVALID_ARGUMENT, "DER decoding failed");
@@ -578,8 +580,8 @@ util::Status Cert::Sha256Digest(string* result) const {
   unsigned char digest[EVP_MAX_MD_SIZE];
   unsigned int len;
   if (X509_digest(x509_.get(), EVP_sha256(), digest, &len) != 1) {
-    // What does this return value mean? Let's assume it means the cert
-    // is bad until proven otherwise.
+    // Failed to digest. Several possible reasons but we will just reject
+    // the input rather than trying to interpret the cause
     LOG(WARNING) << "Failed to compute cert digest";
     LOG_OPENSSL_ERRORS(WARNING);
     return util::Status(Code::INVALID_ARGUMENT, "SHA256 digest failed");
@@ -599,8 +601,8 @@ util::Status Cert::DerEncodedTbsCertificate(string* result) const {
   unsigned char* der_buf(nullptr);
   int der_length = i2d_re_X509_tbs(x509_.get(), &der_buf);
   if (der_length < 0) {
-    // What does this return value mean? Let's assume it means the cert
-    // is bad until proven otherwise.
+    // Failed to serialize. Several possible reasons but we will just reject
+    // the input rather than trying to interpret the cause
     LOG(WARNING) << "Failed to serialize the TBS component";
     LOG_OPENSSL_ERRORS(WARNING);
     return util::Status(Code::INVALID_ARGUMENT, "TBS DER serialize failed");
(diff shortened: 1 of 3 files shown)


cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+// to do the work ourselves. Can be called with a non DSA issuer key but will
+// always return false in that case.
+//
+// cert is the certificate that is to be checked
+// issuer_key is the public key of the certificate issuer (should be DSA)
+// If dsa_key_with_params is not null it's a key that can be used to copy
+// the DSA parameters if they are absent from the issuer key.
+// Returns true if the DSA signature was correctly verified or false if
+// it did not or there was any error along the way.
+StatusOr<bool> check_dsa_signature(const X509* cert, EVP_PKEY* issuer_key,
```
ekasper: Why `StatusOr<bool>` rather than simply a `Status`?

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+                                   const EVP_PKEY* dsa_key_with_params) {
+  // Before we start rummaging about through pointers ensure everything we need
+  // is present
+  if (!cert || !issuer_key || !cert->cert_info || !cert->cert_info->signature ||
```
ekasper: Avoid reaching to struct internals where possible. I don't know what, if anything, BoringSSL will make opaque here but  nevertheless. Bssl has accessors for `X509_get_cert_info` etc.

cpp/log/cert.cc
ekasper: Can the issuer key be const? Perhaps it can't because the underlying API isn't correctly const?

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+      !cert->cert_info->signature->algorithm) {
+    return ::util::Status(Code::FAILED_PRECONDITION,
+                          "cert is null or missing required parameters");
+  }
+
+  if (issuer_key->type != EVP_PKEY_DSA) {
```
ekasper: Ditto, `EVP_PKEY_type`

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+    return ::util::Status(Code::FAILED_PRECONDITION,
+                          "cert does not have a DSA public key");
+  }
+
+  // The nid in the cert_info struct seems to be zero so we look it up
```
ekasper: Don't get this comment - which "nid" in the cert_info struct?

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+  const int alg_nid = OBJ_obj2nid(cert->cert_info->signature->algorithm);
+
+  int digest_nid;
+  // We need the nid of the digest so we can create an EVP_MD for it later.
+  // Should succeed as we know we have a DSA key
```
ekasper: know -> now

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+    // What does this return value mean? Let's assume it means the cert
+    // is bad until proven otherwise. Assume nothing to free.
+    LOG(WARNING) << "Failed to serialize the CINF component";
+    LOG_OPENSSL_ERRORS(WARNING);
+    return ::util::Status(Code::INVALID_ARGUMENT,
+                          "failed to serialize cert info");
+  }
+
+  // If the key is missing parameters try to copy them from the supplied
+  // parent key. TODO: This is not currently enabled as we have no examples
```
ekasper: I'd hope that you'll never need this...

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+  unsigned char md_buffer[EVP_MAX_MD_SIZE];
+  unsigned int md_size;
+
+  // Build the digest of the cert info
+  if (!EVP_Digest(der_buf, der_length, md_buffer, &md_size, md, nullptr)) {
+    OPENSSL_free(der_buf);
+    return ::util::Status(Code::INTERNAL, "digest failed");
+  }
+
+  // And we're done with the DER cert info and can check the result
```
ekasper: Use a scoped ptr?

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+  if (dsa == nullptr || md == nullptr) {
+    OPENSSL_free(der_buf);
+    return ::util::Status(Code::INTERNAL,
+                          "failed to create hasher or get DSA sig");
+  }
+
+  unsigned char md_buffer[EVP_MAX_MD_SIZE];
+  unsigned int md_size;
+
+  // Build the digest of the cert info
```
ekasper: It might be a little shorter to go through `EVP_Verify..`: https://boringssl.googlesource.com/boringssl/+/master/include/openssl/evp.h#321

Or is DSA gone from the EVP layer (in which case add a comment)?

cpp/log/cert.cc
```diff
@@ -43,6 +43,99 @@ static int X509_get_signature_nid(const X509* x) {
...
+  if (!EVP_Digest(der_buf, der_length, md_buffer, &md_size, md, nullptr)) {
+    OPENSSL_free(der_buf);
+    return ::util::Status(Code::INTERNAL, "digest failed");
+  }
+
+  // And we're done with the DER cert info and can check the result
+  OPENSSL_free(der_buf);
+
+  int out_valid;
+  if (DSA_check_signature(&out_valid, md_buffer, md_size, cert->signature->data,
```
ekasper: I know BoringSSL headers say that this is a "safer" version but I actually find this confusing. If you can't go through the EVP layer, maybe do

```
if (!DSA_check_signature(&out_valid, ...)) {
    return ::util::Status(Code::INTERNAL, ...);
}

return out_valid;
```

cpp/log/cert.cc
```diff
@@ -380,6 +475,26 @@ StatusOr<bool> Cert::IsSignedBy(const Cert& issuer) const {
     ClearOpenSSLErrors();
     return false;
   }
+
+  // With BoringSSL we might have a signature algorithm that is not supported
+  // by X509_verify but we still want to accept into a log. This is a weaker
+  // check than x509_verify but sufficient for our needs as we are rejecting
+  // spam rather than intending to trust the certificate.
+  if (lib == ERR_LIB_EVP && reason == EVP_R_UNSUPPORTED_ALGORITHM) {
```
ekasper: I'm not sure how much stability BoringSSL promises wrt error codes but it's generally a pain that applications rely on them and it might be more robust to just always attempt the DSA verify; it'll bail quickly if the key type is wrong.

cpp/log/cert.cc
```diff
@@ -1370,12 +1504,31 @@ util::Status CertChain::IsValidSignatureChain() const {
                         "certificate chain is not loaded");
   }
 
+  // Search up the chain from the leaf for a certificate with a DSA
```
ekasper: Seeing as you do all this only to throw the `|dsa_cert_with_params|` away unused, I'd suggest that you just delete this for now and only deal with it when you really have to. You can't even coerce OpenSSL to spit out such a cert without hacking into source so I think it's really unlikely that they exist.

cpp/log/cert.cc
Martin2112: StatusOr<bool> allows us to distinguish cases where we failed to verify from where we did the verify but found an invalid signature.

issuer_key can't be const unfortunately, I did try.

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc
Martin2112: Removed the comment.

cpp/log/cert.cc
Martin2112: Reworded.

cpp/log/cert.cc
Martin2112: We haven't found any certs in the wild that use this but it's in RFC 3279.

cpp/log/cert.cc

cpp/log/cert.cc
Martin2112: Can't use EVP api for this unfortunately. Comment added.

cpp/log/cert.cc

cpp/log/cert.cc

cpp/log/cert.cc
Martin2112: I removed all this code for the moment. We didn't have anything to test with as we couldn't find any examples.

Martin2112:
PTAL.

cpp/log/cert.cc
```diff
@@ -43,6 +43,100 @@ static int X509_get_signature_nid(const X509* x) {
...
+  if (!OBJ_find_sigid_algs(alg_nid, &digest_nid, nullptr)) {
+    return ::util::Status(Code::INTERNAL, "lookup sigid for algorithm failed");
+  }
+
+  // Get the DER encoded certificate info from the cert
+  unsigned char* der_buf(nullptr);
+  const int der_length = i2d_X509_CINF(cert_info, &der_buf);
+  if (der_length < 0) {
+    // What does this return value mean? Let's assume it means the cert
+    // is bad until proven otherwise.
```
benlaurie: a negative return means an error occurred, ERR_get_error() will say what error...

see, e.g., man i2d_X509.

Martin2112: This was cut and pasted from elsewhere in the file. It occurs quite a few times, shall I fix up all of them?

cpp/log/cert.cc
```diff
@@ -43,6 +43,100 @@ static int X509_get_signature_nid(const X509* x) {
...
+  }
+
+  int out_valid;
+  if (!DSA_check_signature(&out_valid, md_buffer, md_size,
+                           cert->signature->data, cert->signature->length,
+                           dsa)) {
+    return ::util::Status(Code::INTERNAL, "failed to check DSA signature");
+  }
+
+  return out_valid != 0;
```
benlaurie: Documentation says out_valid == 1 on success.

cpp/log/cert.cc
```diff
@@ -1375,7 +1491,8 @@ util::Status CertChain::IsValidSignatureChain() const {
     Cert* subject = *it;
     Cert* issuer = *(it + 1);
 
-    const StatusOr<bool> status = subject->IsSignedBy(*issuer);
+    const StatusOr<bool> status =
+        subject->IsSignedBy(*issuer);
```
benlaurie: Why the wrap?

cpp/log/cert.cc
benlaurie: Sure.

ekasper: I don't remember exactly but I suspect the comment was meant to say that you can't distinguish decoding errors and other errors such as malloc failures.

I would not advocate using ERR_get_error to make a distinction unless you really have to. (That is the case when reading PEM files and encountering EOF, but I don't think it's the case here.) It's kind of a historical tragedy that the error codes are public. They really should be used for debugging and you should not in general assume that they are part of the stable API.

cpp/log/cert.cc
ekasper: As I mentioned in the other thread I'd prefer endorsing sane behaviour, and not tolerating everything just because it's in the RFC. IIRC even OpenSSL doesn't let you generate such certs so I doubt anything will.

cpp/log/cert.cc
```diff
@@ -43,6 +43,100 @@ static int X509_get_signature_nid(const X509* x) {
...
+    return ::util::Status(Code::FAILED_PRECONDITION,
+                          "cert is missing a signature");
+  }
+
+  if (EVP_PKEY_type(issuer_key->type) != EVP_PKEY_DSA) {
+    return ::util::Status(Code::FAILED_PRECONDITION,
+                          "issuer does not have a DSA public key");
+  }
+
+  const int alg_nid = OBJ_obj2nid(sig->algorithm);
```
ekasper: `X509_get_signature_nid`

cpp/log/cert.cc
Martin2112: Fixed.

cpp/log/cert.cc
Martin2112: Fixed.

cpp/log/cert.cc
Martin2112: I've replaced all occurrences of this comment with a note that the decode / serialize failed and we're rejecting the input rather than trying to work out exactly what happened.

Martin2112:
I think I've fixed everything, assuming Travis agrees.

PR #1147, opened 2016-04-07
Use getaddrinfo instead of inet_aton in Client::Connect

RJPercival:
This following are now supported, in addition to IPv4 addresses:
- hostnames
- IPv6 addresses
- service names instead of port numbers

Also fixes a couple of deficiencies:
- Initializes the serializer, allowing the ct tool to deserialize v1 SCTs.
- Loads the system CA roots by default, which allows most certificates to be verified.

### DIFF
--- a/cpp/client/client.cc
+++ b/cpp/client/client.cc
@@ -2,6 +2,7 @@
 
 #include <arpa/inet.h>
 #include <glog/logging.h>
+#include <netdb.h>
 #include <netinet/in.h>
 #include <stddef.h>
 #include <stdint.h>
@@ -15,7 +16,7 @@
 
 using std::string;
 
-Client::Client(const string& server, uint16_t port)
+Client::Client(const string& server, const string& port)
     : server_(server), port_(port), fd_(-1) {
 }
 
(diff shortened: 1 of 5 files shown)


cpp/client/client.cc
```diff
@@ -26,24 +27,42 @@ Client::~Client() {
...
+    AI_ADDRCONFIG, /* ai_flags */
+    AF_UNSPEC,     /* ai_family */
+    SOCK_STREAM,   /* ai_socktype */
+    IPPROTO_TCP    /* ai_protocol */
+  };
+
+  struct addrinfo* server_addr;
+
+  int ret = getaddrinfo(server_.c_str(), port_.c_str(), &server_addr_hints,
+                        &server_addr);
```
pphaneuf: Without the `AI_NUMERICHOST` hint, this can becomes a blocking call where it wasn't before. Are we okay with that?

RJPercival: As far as I can see, this class is only ever used by SSLClient, which in turn is only used by cpp/client/ct.cc - a command-line tool. I think allowing that to block for DNS lookups is a reasonable trade-off in exchange for allowing hostnames.

cpp/client/ct.cc
```diff
@@ -996,6 +997,7 @@ int Monitor() {
 int main(int argc, char** argv) {
   google::SetUsageMessage(argv[0] + string(kUsage));
   util::InitCT(&argc, &argv);
+  ConfigureSerializerForV1CT();
```
pphaneuf: This is probably necessary to make the tool work, but isn't mentioned in the description.

cpp/client/ssl_client.cc
```diff
@@ -80,7 +80,8 @@ SSLClient::SSLClient(const string& server, uint16_t port, const string& ca_dir,
              SSL_CTX_load_verify_locations(ctx_.get(), NULL, ca_dir.c_str()))
         << "Unable to load trusted CA certificates.";
   } else {
-    LOG(WARNING) << "No trusted CA certificates given.";
+    SSL_CTX_set_default_verify_paths(ctx_.get());
```
pphaneuf: Also not mentioned in the description.

cpp/client/client.cc
```diff
@@ -26,24 +27,42 @@ Client::~Client() {
...
+  static const addrinfo server_addr_hints = {
+    AI_ADDRCONFIG, /* ai_flags */
+    AF_UNSPEC,     /* ai_family */
+    SOCK_STREAM,   /* ai_socktype */
+    IPPROTO_TCP    /* ai_protocol */
+  };
+
+  struct addrinfo* server_addr;
+
+  int ret = getaddrinfo(server_.c_str(), port_.c_str(), &server_addr_hints,
```
pphaneuf: Make this `const`.

pphaneuf:
Done with this round, just some changes to the description, really.

cpp/client/ct.cc
RJPercival: Description amended.

cpp/client/ssl_client.cc
RJPercival: Description amended.

pphaneuf:
LGTM

PR #1153, opened 2016-04-13
Add issuer cn matcher to scanner + print whole chains option

Martin2112:
Adds a flag to match certs by issuer CN to Go scanner. Also adds the option to output the whole chain for matches in base64.

### DIFF
--- a/go/scanner/main/scanner.go
+++ b/go/scanner/main/scanner.go
@@ -53,11 +53,7 @@ func chainToString(certs []ct.ASN1Cert) string {
 	return base64.StdEncoding.EncodeToString(output)
 }
 
-func logCertChain(entry *ct.LogEntry) {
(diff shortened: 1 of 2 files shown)


go/scanner/main/scanner.go
```diff
@@ -40,7 +43,45 @@ func logPrecertInfo(entry *ct.LogEntry) {
...
+	}
+
+	return base64.StdEncoding.EncodeToString(output)
+}
+
+func logCertChain(entry *ct.LogEntry) {
+	log.Printf("Index %d: Chain: %s", entry.Index, chainToString(entry.Chain))
+}
+
+func logPrecertChain(entry *ct.LogEntry) {
```
AlCutter: These two methods seem to be the same, could you delete one?

Martin2112: Done.

PR #1155, opened 2016-04-15
Preloader

taknira:
Getting roots during log creation and adding retries to getting roots so that logs (ahem, rocketeer) don't cause the whole preloader run to fail by randomly send a non-200 response just one time.


pphaneuf:
LGTM

PR #1163, opened 2016-04-27
Python: Parse flags in more tests.

eranmes:
The travis CI either fails or has its log filled with Python errors
about accessing flags prior to parsing them (this is an error in
gflags 3.0.4 and above).

This commit fixes the tests by parsing the flags.
Note that for twisted-based tests running using trial the test is
imported so it is not effective to rely on the module name being __main__.


pphaneuf:
LGTM

PR #1172, opened 2016-05-11
Python: Remove cert_analysis

eranmes:
It's not used and getting in the way of clean-ups.


eranmes:
Paul, PTAL - it should make cleaning up the Python logobserver code slightly easier.

python/ct/proto/certificate.proto
```diff
@@ -24,6 +24,7 @@ message DNAttribute {
   optional string value = 2;
 }
 
+// TODO(eran): Remove this message.
```
phad: You should mark the place where it is used, in X509Description, as deprecated.

python/utilities/logobserver/logobserver.py
```diff
@@ -6,7 +6,6 @@
 import sys
 import requests
 
-from ct.cert_analysis import tld_list
```
phad: did you also delete this source file?

python/ct/proto/certificate.proto
eranmes: Done.

python/utilities/logobserver/logobserver.py
eranmes: Yes.

phad:
Thanks, LGTM modulo Travis.

PR #1173, opened 2016-05-12
Python: Separate AsyncLogClient from the LogClient

eranmes:
There really is no reason for those separate classes to live in the same
package.


eranmes:
To be clear, there are no functional changes: I've simply moved the AsyncLogClient from log_client.py to a module of its own.

python/ct/client/log_client.py
```diff
@@ -2,57 +2,23 @@
...
 
 gflags.DEFINE_integer("entry_fetch_batch_size", 1000, "Maximum number of "
                       "entries to attempt to fetch in one request.")
 
-gflags.DEFINE_integer("max_fetchers_in_parallel", 100, "Maximum number of "
-                      "concurrent fetches.")
-
-gflags.DEFINE_integer("get_entries_retry_delay", 1, "Number of seconds after "
-                      "which get-entries will be retried if it encountered "
-                      "an error.")
```
pphaneuf: Should remove this line, for consistency.

python/ct/client/log_client.py
```diff
@@ -571,426 +527,3 @@ def add_chain(self, certs_list):
             {'chain': [base64.b64encode(certificate) for certificate in certs_list]})
         return self._parse_sct(sct_data)
```
pphaneuf: Should delete this trailing line.

python/ct/client/async_log_client.py
```diff
@@ -0,0 +1,482 @@
+"""RFC 6962 client API."""
+import base64
+import json
```
pphaneuf: I believe these two imports are unused?

pphaneuf:
Done with this round of review.

There's now a fair bit of digging into `log_client`, including for some underscore-prefixed things... Might need a follow up sort of cleanup?

python/ct/client/async_log_client.py
eranmes: Correct, removed.

python/ct/client/log_client.py
eranmes: Done.

python/ct/client/log_client.py
eranmes: Done.

eranmes:
Thanks for the review, will submit shortly.
@pphaneuf re digging into the log_client, I agree, some clean-up may be in order, do you have a concrete suggestion on how to address that?
I could extract common constants/functions to log_client_base or something.

pphaneuf:
LGTM.

Yes, something along those lines? Maybe more `log_client_common` (or just `common.py`, as this is already in the `ct/client/` directory, provides enough context, no "stuttering" on import), as I believe the blocking and non-blocking clients are quite independent (they don't derive from a common base class, say).

PR #1178, opened 2016-05-16
Make SerialHasher::Create return a unique_ptr.

pphaneuf:
This enforces the ownership transfer at compile-time.

### DIFF
--- a/cpp/merkletree/serial_hasher.h
+++ b/cpp/merkletree/serial_hasher.h
@@ -3,6 +3,7 @@
 
 #include <openssl/sha.h>
 #include <stddef.h>
+#include <memory>
 #include <string>
 
 #include "base/macros.h"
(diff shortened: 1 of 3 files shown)


cpp/merkletree/serial_hasher.h
```diff
@@ -26,8 +27,8 @@ class SerialHasher {
   // Finalize the hash context and return the binary digest blob.
   virtual std::string Final() = 0;
 
-  // A virtual constructor.  The caller gets ownership of the returned object.
-  virtual SerialHasher* Create() const = 0;
+  // A virtual constructor.
```
Martin2112: This comment doesn't add much value now.

pphaneuf: Expanded it a little, better?

Martin2112: Yes, looks good.

Martin

PR #1182, opened 2016-05-17
Add a non-const version of StatusOr<>::ValueOrDie.

pphaneuf:


cpp/util/statusor.h
```diff
@@ -68,6 +68,10 @@ class StatusOr {
     CHECK(ok()) << "Attempting to fetch value of non-OK StatusOr";
     return value_;
   }
+  inline T& ValueOrDie() {
```
Martin2112: Might need a blank line before that, style guide?

PR #1184, opened 2016-05-17
Convert HTTPLogClient::UploadSubmission to use StatusOr.

pphaneuf:


cpp/client/ct.cc
```diff
@@ -335,13 +335,12 @@ static int Upload() {
...
-  AsyncLogClient::Status ret =
-      client.UploadSubmission(contents, FLAGS_precert, &sct);
+  const StatusOr<SignedCertificateTimestamp> sct(
+      client.UploadSubmission(contents, FLAGS_precert));
 
-  if (ret != AsyncLogClient::OK) {
-    LOG(ERROR) << "Submission failed, error = " << ret;
+  if (!sct.status().ok()) {
+    LOG(ERROR) << "Submission failed: " << sct.status();
     return 1;
```
Martin2112: Not sure why this API was defined to return an int but hey ho.

pphaneuf: This ends up being the exit status of the tool, returned by the main, below.

PR #1186, opened 2016-05-18
Python: Remove the Observations message.

eranmes:
Now that the observation gathering code is gone, this message is
redundant.

### DIFF
--- a/python/ct/proto/certificate.proto
+++ b/python/ct/proto/certificate.proto
@@ -24,15 +24,6 @@ message DNAttribute {
   optional string value = 2;
 }
 
-// TODO(eran): Remove this message.
-message Observation {
-  optional string description = 1;
-
-  optional string reason = 2;
-
-  optional bytes details = 3;
-}
-
 message X509Description {
   optional string version = 1;
 
@@ -58,8 +49,7 @@ message X509Description {
   // SHA256 hash from DER form.
   optional bytes sha256_hash = 10;
 
-  // Deprecated - The code to fill this field in was removed.
-  repeated Observation observations = 11;
+  // Field 11 was Observation, no longer produced.
 
   repeated DNAttribute root_issuer = 12;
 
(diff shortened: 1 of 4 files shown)


python/ct/proto/certificate.proto
```diff
@@ -58,8 +49,7 @@ message X509Description {
   // SHA256 hash from DER form.
   optional bytes sha256_hash = 10;
 
-  // Deprecated - The code to fill this field in was removed.
-  repeated Observation observations = 11;
+  // Field 11 was Observation, no longer produced.
```
pphaneuf: Could use the `reserved` feature: https://developers.google.com/protocol-buffers/docs/proto#reserved

pphaneuf:
There are some mentions of `observations` in `cert_desc.py`.

Also, in a docstring of `CertificateReport.scan_der_certs`, in `reporter.py` (maybe just need rewording the docstring, or maybe the code needs to go as well? not sure).

eranmes:
Good catch, thanks, though that code didn't do anything because, as far as I can tell, nothing provides observations to the cert_desc constructor anymore.

eranmes:
@pphaneuf PTAL.

pphaneuf:
LGTM

python/ct/proto/certificate.proto
eranmes: Done.

eranmes:
FYI I've reverted the change to use 'reserved' as it does not seem to be supported by the protocol buffer compiler used in travis or on my workstation (2.5.0) and I'd rather not force everyone using this project to upgrade their protocol buffer compiler.

phad:
I think there's an alternative method you can use based on proto extensions
that achieves the same goal as 'reserved' in preventing re-use of the ID.

pphaneuf:
If it's choking up on it, I'm fine with just a comment.

eranmes:
A comment it is.

PR #1193, opened 2016-05-18
Pass a non-const  MerkleTree as a pointer.

pphaneuf:
To avoid surprises, as recommended by the style guide:

https://google.github.io/styleguide/cppguide.html#Reference_Arguments


eranmes:
For the record, this is done to comply with the style guide.

pphaneuf:
I'll add a `CHECK_NOTNULL` before merging.

PR #1198, opened 2016-05-19
Parse flags in more tests.

pphaneuf:
This is a backport of commit 1d463480dd967f0054bc9e59e8eff1256888d993.

Does not fix everything, but makes it better...


pphaneuf:
Note that Travis is failing, but that's expected. It's failing "less badly" than before, and hopefully a follow up PR should get it passing. 😉

PR #1208, opened 2016-05-23
Fork the ct.proto file for Java.

pphaneuf:
The Java side (almost) doesn't read or write those, using them as in-memory data structures. Splitting it off allows refactorings to be done more easily, on both sides.

### DIFF
--- a/java/ct.proto
+++ b/java/ct.proto
@@ -1,0 +1,124 @@
+syntax = "proto2";
+
+package ct;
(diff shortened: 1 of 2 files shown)


java/ct.proto
```diff
@@ -0,0 +1,124 @@
+syntax = "proto2";
+
+package ct;
+option java_package = "org.certificatetransparency.ctlog.proto";
```
eranmes: You'd want to remove the java_package from proto/ct.proto now...

PR #1209, opened 2016-05-23
Converted MerkleAuditProof to a POJO.

pphaneuf:


java/src/org/certificatetransparency/ctlog/MerkleAuditProof.java
```diff
@@ -0,0 +1,18 @@
+package org.certificatetransparency.ctlog;
+
+import org.certificatetransparency.ctlog.proto.Ct;
+
+
+public class MerkleAuditProof {
+  public Ct.Version version;
+  public long treeSize;
+  public long leafIndex;
+  public java.util.List<byte[]> pathNode;
```
phad: erm, wouldn't it be more readable to import java.util.List at the top?

java/src/org/certificatetransparency/ctlog/MerkleAuditProof.java
```diff
@@ -0,0 +1,18 @@
...
+
+  public MerkleAuditProof(Ct.Version version_, long treeSize_, long leafIndex_) {
```
phad: this is odd style.  trailing underscores I mean.  you might remove them, and have the lines below read 'this.version = version' which is pretty much how I've seen Java done.

java/src/org/certificatetransparency/ctlog/MerkleAuditProof.java
pphaneuf: I think I stole this from somewhere which had somehow decided that this was okay, but sure.

java/src/org/certificatetransparency/ctlog/MerkleAuditProof.java
```diff
@@ -0,0 +1,18 @@
...
+    version = version_;
+    treeSize = treeSize_;
+    leafIndex = leafIndex_;
+    pathNode = new java.util.ArrayList<byte[]>();
```
pphaneuf: Is that a reasonable implementation of `List<>` to use? I just looked around and poked in the dark until it compiled and the tests passed, really... :wink:

java/src/org/certificatetransparency/ctlog/MerkleAuditProof.java
pphaneuf: Done.

java/src/org/certificatetransparency/ctlog/MerkleAuditProof.java
phad: Yes, I think it's ok.

java/src/org/certificatetransparency/ctlog/MerkleAuditProof.java
```diff
@@ -0,0 +1,20 @@
...
+  public Ct.Version version;
+  public long treeSize;
+  public long leafIndex;
+  public List<byte[]> pathNode;
+
+  public MerkleAuditProof(Ct.Version version, long treeSize, long leafIndex) {
+    this.version = version;
+    this.treeSize = treeSize;
+    this.leafIndex = leafIndex;
+    pathNode = new ArrayList<byte[]>();
```
pphaneuf: Is that a reasonable implementation of `List<>` to use? I just looked around and poked in the dark until it compiled and the tests passed, really... :wink:

PR #1210, opened 2016-05-23
Make sure Travis notices failing Java tests.

pphaneuf:
Tested it by adding an intentional failure, works great, thanks for the tip!


eranmes:
Yeah, that's a good idea.

PR #1211, opened 2016-05-23
Converted MerkleTreeLeaf to a POJO.

pphaneuf:


java/src/org/certificatetransparency/ctlog/MerkleTreeLeaf.java
```diff
@@ -0,0 +1,16 @@
+package org.certificatetransparency.ctlog;
+
+import org.certificatetransparency.ctlog.proto.Ct;
+
+
+public class MerkleTreeLeaf {
```
Martin2112: It looks to me that instances of this could be made immutable.

pphaneuf: I'm not sure what that means, but I like the sound of it? :smiley:

Martin2112: If you declare them as public final they can only be set once at construction. In this case though the objects are presumably modifiable so it's unclear as to whether this is a great help.

pphaneuf: In general, I prefer more `const`/immutable/`final`, but since I don't really know what's consuming this downstream all that well, I think I'll leave it like this...

PR #1214, opened 2016-05-23
De-template some internal methods of EtcdConsistentStore.

pphaneuf:


pphaneuf:
After quite a bit of struggle with flaky builds, Travis is finally green. 😛

eranmes:
Quick! Commit before it's red again!

cpp/log/consistent_store.h
```diff
@@ -79,6 +81,10 @@ class EntryHandle : public EntryHandleBase {
     return &entry_;
   }
 
+  bool SerializeToString(std::string* output) const override {
```
eranmes: One may wonder why this method is needed given you could simply call Entry().SerializeToString ?

pphaneuf: This is to implement the pure virtual in the parent class, which allows method that only need to write the entry to etcd (without needing anything else from the entry) to take the non-templated parent class (avoiding the need to be templated themselves).

PR #1216, opened 2016-05-24
Add new field accessors to Cert.

phad:
* PrintVersion() - the X509 Version in readable form, ie. V3 is returned as '3'
* SubjectAltNames() - returns the dNSName values from the SAN field
* SPKI() - returns the subject public key.

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -16,6 +16,7 @@
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 #include <time.h>
+#include <algorithm>
 #include <memory>
 #include <string>
 #include <vector>
@@ -233,6 +234,34 @@ util::Status Cert::LoadFromDerBio(BIO* bio_in) {
     return util::Status(Code::INVALID_ARGUMENT, "Not a valid encoded cert");
   }
   return util::Status::OK;
+}
+
+
+string Cert::PrintVersion() const {
+  if (!IsLoaded()) {
+    LOG(ERROR) << "Cert not loaded";
+    return string();
+  }
+
+  const long version(X509_get_version(x509_.get()));
+  return to_string(1 + version);
+}
+
+
+string Cert::PrintSerialNumber() const {
+  if (!IsLoaded()) {
+    LOG(ERROR) << "Cert not loaded";
+    return string();
+  }
+
+  ScopedBIGNUM serial_number_BN(
+      ASN1_INTEGER_to_BN(X509_get_serialNumber(x509_.get()), NULL));
+
+  ScopedOpenSSLString serial_number_hex(BN_bn2hex(serial_number_BN.get()));
+
+  string serial(serial_number_hex.get());
+  std::transform(serial.begin(), serial.end(), serial.begin(), ::toupper);
+  return serial;
 }
 
 
@@ -671,7 +700,7 @@ util::Status Cert::PublicKeySha256Digest(string* result) const {
 }
 
 
-util::Status Cert::SPKISha256Digest(string* result) const {
+StatusOr<string> Cert::SPKI() const {
   if (!IsLoaded()) {
     LOG(ERROR) << "Cert not loaded";
     return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
@@ -681,19 +710,29 @@ util::Status Cert::SPKISha256Digest(string* result) const {
   int der_length =
       i2d_X509_PUBKEY(X509_get_X509_PUBKEY(x509_.get()), &der_buf);
   if (der_length < 0) {
-    // Failed to serialize. Several possible reasons but we will just reject
-    // the input rather than trying to interpret the cause
+    // What does this return value mean? Let's assume it means the cert
+    // is bad until proven otherwise.
     LOG(WARNING) << "Failed to serialize the Subject Public Key Info";
     LOG_OPENSSL_ERRORS(WARNING);
-    return util::Status(Code::INVALID_ARGUMENT, "SPKI SHA256 digest failed");
+    return util::Status(Code::INVALID_ARGUMENT, "Cert::SPKI() failed");
   }
 
-  string sha256_digest = Sha256Hasher::Sha256Digest(
-      string(reinterpret_cast<char*>(der_buf), der_length));
+  string result;
+  result.assign(
+      string(reinterpret_cast<char*>(CHECK_NOTNULL(der_buf)), der_length));
 
-  result->assign(sha256_digest);
   OPENSSL_free(der_buf);
-  return util::Status::OK;
+  return result;
+}
+
+
+util::Status Cert::SPKISha256Digest(string* result) const {
+  const util::StatusOr<string> spki(SPKI());
+  if (spki.ok()) {
+    string sha256_digest = Sha256Hasher::Sha256Digest(spki.ValueOrDie());
+    CHECK_NOTNULL(result)->assign(sha256_digest);
+  }
+  return spki.status();
 }
 
 util::Status Cert::OctetStringExtensionData(int extension_nid,
@@ -837,6 +876,34 @@ bool IsValidRedactedHost(const string& hostname) {
 
 namespace {
 
+util::Status ExtractSubjectAltNames(STACK_OF(GENERAL_NAME)* subject_alt_names,
+                                    vector<string>* dns_alt_names) {
+  CHECK_NOTNULL(subject_alt_names);
+  CHECK_NOTNULL(dns_alt_names);
+
+  dns_alt_names->clear();
+  const int subject_alt_name_count = sk_GENERAL_NAME_num(subject_alt_names);
+
+  for (int i = 0; i < subject_alt_name_count; ++i) {
+    GENERAL_NAME* const name(sk_GENERAL_NAME_value(subject_alt_names, i));
+
+    util::Status name_status;
+    if (name->type == GEN_DNS) {
+      const string dns_name =
+          ASN1ToStringAndCheckForNulls(name->d.dNSName, "DNS name",
+                                         &name_status);
+
+      if (!name_status.ok()) {
+        return name_status;
+      }
+
+      dns_alt_names->push_back(dns_name);
+    }
+  }
+  return util::Status::OK;
+}
+
+
 
 bool ValidateRedactionSubjectAltNames(STACK_OF(GENERAL_NAME) *
                                           subject_alt_names,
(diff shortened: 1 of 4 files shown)


cpp/log/cert.cc
```diff
@@ -236,6 +236,31 @@ util::Status Cert::LoadFromDerBio(BIO* bio_in) {
...
 
 
+string Cert::PrintVersion() const {
+  if (!IsLoaded()) {
+    LOG(ERROR) << "Cert not loaded";
+    return string();
+  }
+
+  const long version(X509_get_version(x509_.get()));
+  return std::to_string(1 + version);
```
pphaneuf: There's a `using std::to_string` at the top, so you can drop the `std::` qualifier.

cpp/log/cert.h
```diff
@@ -148,12 +150,21 @@ class Cert {
...
+  // Returns ?? if the cert is not loaded.
+  util::Status SubjectAltNames(std::vector<std::string>* dns_alt_names) const;
+
   // Sets the SHA256 digest of the cert's subjectPublicKeyInfo in |result|.
   // Returns TRUE if computing the digest succeeded.
   // Returns FALSE if computing the digest failed.
   // Returns ERROR if the cert is not loaded.
   util::Status SPKISha256Digest(std::string* result) const;
 
+  // Returns the cert's subkectPublicKeyInfo, or an error.
```
pphaneuf: Typo in subkectPublicKeyInfo?

cpp/log/cert.cc
```diff
@@ -671,6 +696,31 @@ util::Status Cert::PublicKeySha256Digest(string* result) const {
...
 
+StatusOr<string> Cert::SPKI() const {
+  if (!IsLoaded()) {
+    LOG(ERROR) << "Cert not loaded";
+    return util::Status(Code::FAILED_PRECONDITION, "Cert not loaded");
+  }
+
+  unsigned char* der_buf(nullptr);
+  int der_length =
+      i2d_X509_PUBKEY(X509_get_X509_PUBKEY(x509_.get()), &der_buf);
```
pphaneuf: `X509_get_X509_PUBKEY`, that never returns null? Or, if it does, `i2d_X509_PUBKEY` can handle it?

cpp/log/cert.h
phad: done

cpp/log/cert.cc
phad: oops, didn't notice that.

cpp/log/cert.cc
```diff
@@ -671,6 +696,31 @@ util::Status Cert::PublicKeySha256Digest(string* result) const {
...
+  if (der_length < 0) {
+    // What does this return value mean? Let's assume it means the cert
+    // is bad until proven otherwise.
+    LOG(WARNING) << "Failed to serialize the Subject Public Key Info";
+    LOG_OPENSSL_ERRORS(WARNING);
+    return util::Status(Code::INVALID_ARGUMENT, "SPKI() failed");
+  }
+
+  string result;
+  result.assign(string(reinterpret_cast<char*>(der_buf), der_length));
```
pphaneuf: Could you put a `CHECK_NOTNULL` around `der_buf`?

cpp/log/cert.cc
```diff
@@ -671,6 +696,31 @@ util::Status Cert::PublicKeySha256Digest(string* result) const {
...
+
+  unsigned char* der_buf(nullptr);
+  int der_length =
+      i2d_X509_PUBKEY(X509_get_X509_PUBKEY(x509_.get()), &der_buf);
+  if (der_length < 0) {
+    // What does this return value mean? Let's assume it means the cert
+    // is bad until proven otherwise.
+    LOG(WARNING) << "Failed to serialize the Subject Public Key Info";
+    LOG_OPENSSL_ERRORS(WARNING);
+    return util::Status(Code::INVALID_ARGUMENT, "SPKI() failed");
```
pphaneuf: Add "Cert::" to the beginning of the message.

cpp/log/cert.cc
```diff
@@ -671,6 +696,31 @@ util::Status Cert::PublicKeySha256Digest(string* result) const {
...
+
+  string result;
+  result.assign(string(reinterpret_cast<char*>(der_buf), der_length));
+
+  OPENSSL_free(der_buf);
+  return result;
+}
+
+
 util::Status Cert::SPKISha256Digest(string* result) const {
```
pphaneuf: It'd be nice if this method was updated to rely on the one you just added, instead of just duplicating most of it?

cpp/log/cert.h
```diff
@@ -148,12 +150,21 @@ class Cert {
   // Returns ERROR if the cert is not loaded.
   util::Status PublicKeySha256Digest(std::string* result) const;
 
+  // Sets the Subject Alternative Name dNSNames in |dns_alt_names|.
+  // Returns util::Status::OK if the SAN dNSNames were extracted.
+  // Returns ?? if the DAN dNSNames could not be extracted.
+  // Returns ?? if the cert is not loaded.
+  util::Status SubjectAltNames(std::vector<std::string>* dns_alt_names) const;
```
pphaneuf: Would be nicer if it just returned `util::StatusOr<std::vector<std::string>>>`?

cpp/log/cert.cc
phad: `X509_get_X509_PUBKEY` can return null. If that happens, `i2d_X509_PUBKEY` returns a zero-length and doesn't modify der_buf.  The function then returns (status::OK, "")

cpp/log/cert.cc
```diff
@@ -887,10 +963,23 @@ bool ValidateRedactionSubjectAltNames(STACK_OF(GENERAL_NAME) *
 }  // namespace
 
 
+util::Status Cert::SubjectAltNames(vector<string>* dns_alt_names) const {
+  ScopedGENERAL_NAMEStack subject_alt_names(
+      static_cast<STACK_OF(GENERAL_NAME)*>(X509_get_ext_d2i(
+          x509_.get(), NID_subject_alt_name, nullptr, nullptr)));
+
+  if (subject_alt_names) {
```
pphaneuf: Could skip the conditional, since it duplicates the check.

Also, that helper method isn't very big, could just be inlined? Or is it going to be used again in some upcoming PR?

cpp/log/cert.cc
```diff
@@ -887,10 +963,23 @@ bool ValidateRedactionSubjectAltNames(STACK_OF(GENERAL_NAME) *
...
+    return ExtractSubjectAltNames(subject_alt_names.get(), dns_alt_names);
+  }
+
+  return util::Status::OK;
```
pphaneuf: Unless... it should return `NOT_FOUND`when there `subject_alt_names` is null?

cpp/log/cert.cc
```diff
@@ -837,6 +887,32 @@ bool IsValidRedactedHost(const string& hostname) {
...
+  if (subject_alt_names) {
+    const int subject_alt_name_count = sk_GENERAL_NAME_num(subject_alt_names);
+
+    for (int i = 0; i < subject_alt_name_count; ++i) {
+      GENERAL_NAME* const name(sk_GENERAL_NAME_value(subject_alt_names, i));
+
+      util::Status name_status;
+      if (name->type == GEN_DNS) {
+        const string dns_name =
+            ASN1ToStringAndCheckForNulls(name->d.dNSName, "DNS name",
```
pphaneuf: Wow, thanks for bringing this function to my attention, it needs a bit of love (returning a `util::Status` as an out parameter, backward from the rest?!?). Leave it to me, though. :wink:

cpp/log/cert.cc
```diff
@@ -887,10 +963,23 @@ bool ValidateRedactionSubjectAltNames(STACK_OF(GENERAL_NAME) *
 }  // namespace
 
 
+util::Status Cert::SubjectAltNames(vector<string>* dns_alt_names) const {
+  ScopedGENERAL_NAMEStack subject_alt_names(
+      static_cast<STACK_OF(GENERAL_NAME)*>(X509_get_ext_d2i(
+          x509_.get(), NID_subject_alt_name, nullptr, nullptr)));
+
+  if (subject_alt_names) {
+    return ExtractSubjectAltNames(subject_alt_names.get(), dns_alt_names);
```
pphaneuf: If this stays like this, use `CHECK_NOTNULL` on `dns_alt_names`.

pphaneuf:
OK, done for this round of review.

cpp/log/cert.cc
phad: done

pphaneuf:
Except this little detail:

```
[ RUN      ] CertTest.PrintSerialNumber
cpp/log/cert_test.cc:390: Failure
Value of: Cert(google_pem_).PrintSerialNumber()
  Actual: "605381f50001000088bd"
Expected: "605381F50001000088BD"
[  FAILED  ] CertTest.PrintSerialNumber (1 ms)
```

cpp/log/cert.cc
phad: I'm sure my original intentions were honorable...

phad: I'll add a bit of test coverage for the (currently entirely untested) SPKISha256Digest first.

cpp/log/cert.h
phad: I was in two minds about util::Status Fn(stuff, Thing* out)  and util::StatusOr<Thing> Fn(stuff)
Actually I might make the SPKI method be the former to be consistent with everything else first.

cpp/log/cert.cc
phad: done

cpp/log/cert.cc
phad: Quite a bit of that fn was duplicated in the new ExtractSubjectAltNames() so I've re-implemented it to use ExtractSubjectAltNames() instead.

cpp/log/cert.cc
phad: I think it's quite valid for a cert to have no SANs so returning OK with the vector set empty should be good, no?

phad:
I've normalized the PrintSerialNumber() method to up-case the result.  Should address the BORING_SSL/INTERESTING_SSL difference.

cpp/log/cert.cc
phad: The helper is shared with ValidateRedactionSubjectAltNames()

cpp/log/cert.cc
```diff
@@ -846,35 +913,19 @@ bool ValidateRedactionSubjectAltNames(STACK_OF(GENERAL_NAME) *
...
-          if (!IsValidRedactedHost(dns_name)) {
-            LOG(WARNING) << "Invalid redacted host: " << dns_name;
-            *status = util::Status(Code::INVALID_ARGUMENT,
-                                   "Invalid redacted hostname");
-            return true;
-          }
-
-          redacted_name_count++;
-        }
+        redacted_name_count++;
```
phad: ooh, a fun bug!

cpp/log/cert.h
pphaneuf: I was thinking of refactoring the other methods to use `StatusOr<>` more, actually, so doing it like this saves me work. :wink:

cpp/log/cert.cc
pphaneuf: Is it? I mean, I see a resemblance, but they're not sharing right now?

phad: The helper was extracted from ValidateRedactionSubjectAltNames() originally.  I intended to then go back and refactor, but forgot.

You're right, there's a difference: the original ValidateRedactionSubjectAltNames() did extra stuff at the back end of its 'for' loop.

I figured that it would be acceptable to do those extra things as a second phase for-loop in order to promote the re-use of the extracted method.  The behaviour in the error case will be a little different (the vector<string> will have more content in) but I thought that probably wouldn't matter.

cpp/log/cert.cc
```diff
@@ -710,40 +714,25 @@ StatusOr<string> Cert::SPKI() const {
...
-  if (der_length < 0) {
-    // Failed to serialize. Several possible reasons but we will just reject
-    // the input rather than trying to interpret the cause
-    LOG(WARNING) << "Failed to serialize the Subject Public Key Info";
-    LOG_OPENSSL_ERRORS(WARNING);
-    return util::Status(Code::INVALID_ARGUMENT, "SPKI SHA256 digest failed");
+  const util::StatusOr<string> spki(SPKI());
+  if (spki.ok()) {
+    string sha256_digest = Sha256Hasher::Sha256Digest(spki.ValueOrDie());
+    result->assign(sha256_digest);
```
pphaneuf: Why don't people just do `*result = Sha256Hasher::Sha256Digest(spki.ValueOrDie());`? Or, if you like the temporary (to give it a name, as a form of documentation?), `*result = sha256_digest`?

Not your doing, but could you at least add a `CHECK_NOTNULL`, while you're at it?

cpp/log/cert.cc
pphaneuf: Are you going to do it in this PR, or in a separate one? Just so I know to hold off my review until you're done with that change, I don't really mind either way.

phad: No, it's already done here.

cpp/log/cert.cc
pphaneuf: SGTM.

cpp/log/cert.cc
```diff
@@ -887,10 +938,24 @@ bool ValidateRedactionSubjectAltNames(STACK_OF(GENERAL_NAME) *
...
+
+  return util::Status::OK;
+}
+
+
 // Helper method for validating V2 redaction rules. If it returns true
 // then the result in status is final.
 bool Cert::ValidateRedactionSubjectAltNameAndCN(int* dns_alt_name_count,
-                                                util::Status* status) const {
+                                              util::Status* status) const {
```
pphaneuf: Spacing was changed here?

cpp/log/cert.cc
pphaneuf: Ah, damn, it was collapsed in some funny way, I see it now. :stuck_out_tongue:

pphaneuf:
Just a tiny whitespace issue, LGTM otherwise.

cpp/log/cert.cc
phad: oops, fixed.

phad:
I'll squash and re-push, then merge.

PR #1217, opened 2016-05-24
Converted LogEntry to a POJO.

pphaneuf:


java/src/org/certificatetransparency/ctlog/LogEntry.java
```diff
@@ -0,0 +1,12 @@
+package org.certificatetransparency.ctlog;
+
+import org.certificatetransparency.ctlog.proto.Ct;
+
+
+public class LogEntry {
+  public Ct.X509ChainEntry x509Entry;
```
phad: Do you plan next to POJOfy X509ChainEntry and PrecertChainEntry?

java/src/org/certificatetransparency/ctlog/LogEntry.java
```diff
@@ -0,0 +1,12 @@
...
+  public Ct.PrecertChainEntry preCertEntry;
```
phad: case nit - it's 'precert' normally in our code, no?

java/src/org/certificatetransparency/ctlog/LogEntry.java
```diff
@@ -0,0 +1,12 @@
...
+
+  public LogEntry() {
```
phad: perhaps this ctor can be omitted?

java/src/org/certificatetransparency/ctlog/LogEntry.java
pphaneuf: Yes, much POJOing. I'm momentarily stuck with `SignedCertificateTimestamp` and `DigitallySigned`, but I should be able to sort it out at some point...

pphaneuf: The end state is to not require `protobuf.jar` or `protoc` to build the Java side.

java/src/org/certificatetransparency/ctlog/LogEntry.java
pphaneuf: Except for `PreCert`?

The former protobuf setter was called `setPrecertEntry`, so you mean that this should be called `precertEntry`?

java/src/org/certificatetransparency/ctlog/LogEntry.java
pphaneuf: You can omit constructors? :smiley:

phad: It was a punt on my behalf, but this confirms it:
http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.8.9

java/src/org/certificatetransparency/ctlog/LogEntry.java
phad: yes, it should be 'precertEntry'

java/src/org/certificatetransparency/ctlog/LogEntry.java
phad: Cool, sgtm

java/src/org/certificatetransparency/ctlog/LogEntry.java
pphaneuf: Oh, I remember, I was looking at the local variable that it was being set from. Changed it.

Of note, this field is write-only. :stuck_out_tongue:

PR #1222, opened 2016-05-26
Add add-json function

gdbelvin:


go/client/logclient.go
```diff
@@ -57,6 +58,12 @@ type addChainResponse struct {
 	Signature  string     `json:"signature"`   // Log signature for this SCT
 }
 
+// addJSONRequest represents the JSON request bodoy sent ot the add-json CT
```
Martin2112: typo: body

go/client/logclient.go
```diff
@@ -57,6 +58,12 @@ type addChainResponse struct {
...
+// method.
+type addJSONRequest struct {
+	Data string `json:"data"`
```
pphaneuf: If you change `string` for `interface{}` here...

go/client/logclient.go
```diff
@@ -277,6 +284,37 @@ func (c *LogClient) AddChainWithContext(ctx context.Context, chain []ct.ASN1Cert
 	return c.addChainWithRetry(ctx, AddChainPath, chain)
 }
 
+func (c *LogClient) AddJSON(data []byte) (*ct.SignedCertificateTimestamp, error) {
```
pphaneuf: ... `[]byte` for `interface{}`...

go/client/logclient.go
```diff
@@ -277,6 +284,37 @@ func (c *LogClient) AddChainWithContext(ctx context.Context, chain []ct.ASN1Cert
...
+	req := addJSONRequest{
+		Data: base64.StdEncoding.EncodeToString(data),
```
pphaneuf: And `base64.StdEncoding.EncodeToString(data)` for just `data`, it will then take anything, and the JSON encoder already base64-encodes a `byte[]`, so you'll get exactly the same result as now without changing anything to the calling code, while being more general (accepting anything that will encode to JSON).

gdbelvin:
Addressed comments. PTAL

pphaneuf:
LGTM, although a test would be lovely? Too much to ask? :wink:

gdbelvin:
As you wish!

pphaneuf:
Nice, thanks! Waiting on the Travis build to be green, and I'll merge it, shouldn't be too long (the Mac build needs a bit of TLC, I'm taking care of it!).

PR #1231, opened 2016-05-27
Converted SignedEntry to a POJO.

pphaneuf:


java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
```diff
@@ -114,12 +115,12 @@ public static ParsedLogEntry parseLogEntry(InputStream merkleTreeLeaf, InputStre
...
       Ct.X509ChainEntry x509EntryChain = parseX509ChainEntry(extraData,
-          treeLeaf.timestampedEntry.signedEntry.getX509());
+          treeLeaf.timestampedEntry.signedEntry.x509);
       logEntry.x509Entry = x509EntryChain;
     } else if (entryType == Ct.LogEntryType.PRECERT_ENTRY) {
       Ct.PrecertChainEntry preCertChain = parsePrecertChainEntry(extraData,
-          treeLeaf.timestampedEntry.signedEntry.getPrecert());
-       logEntry.precertEntry = preCertChain;
+          treeLeaf.timestampedEntry.signedEntry.preCert);
+      logEntry.precertEntry = preCertChain;
```
pphaneuf: Am I the only one to find this "get a thing into a local, assign local to member variable of same name, never use local ever again" pattern kind of odd?

phad: it is to some extent an act of documentation.  here, preCertChain's type is clear for all to see.
folding those two lines into one, and it is left for the reader to go dig out (or hover with their mouse in their fancy IDE in Java land)

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
```diff
@@ -197,9 +194,9 @@ public static TimestampedEntry parseTimestampedEntry(InputStream in) {
    * @throws SerializationException if an I/O error occurs.
    * @return {@link Ct.X509ChainEntry} proto object.
    */
-  public static Ct.X509ChainEntry parseX509ChainEntry(InputStream in, ByteString x509Cert) {
+  public static Ct.X509ChainEntry parseX509ChainEntry(InputStream in, byte[] x509Cert) {
```
phad: this looks like a public API change.  is that OK?

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
pphaneuf: In C++, I'd have been sorely tempted to use `auto`. In Go, you'd do `localThing := getThing()`. Considering Java is the one where users are most likely to have a fancy IDE (and so, need this the least), as well as typically having the longest type names, I find this especially masochistic?

I guess that's why they invented the fancy IDEs that can rename all the things for you? :wink:

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
pphaneuf: It's only called internally. I'm not sure why it's marked `public`?

But this whole series of PRs is changing the API, as the data structures are what end-users get from this...

pphaneuf:
Is this one okay?

phad:
ah yes, sorry I forgot to close this one.  LGTM.

PR #1234, opened 2016-05-31
Make Cert more robust.

pphaneuf:
The idea here is to reduce the state space of the `Cert` object by making it absolutely certain that `x509_` is always non-null. At the moment, `Cert` behaves a bit like a smart pointer for `X509` (with "extra features"), while being also often put inside a smart pointer, so it's easy to check that one isn't null, and then just go on, while forgetting to check the other level.

I *strongly* recommend reviewing this one commit at a time! Each commit makes sense, and could be merged individually (as long as they're kept in sequence, anyway), I could merge this either as a merge commit or fast-forwarded, let me know if you have a preference?

### DIFF
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -177,24 +177,27 @@ Cert::Cert(ScopedX509 x509) : x509_(move(x509)) {
 }
 
 
-Cert::Cert(const string& pem_string) {
+unique_ptr<Cert> Cert::FromPemString(const std::string& pem_string) {
   // A read-only bio.
   ScopedBIO bio_in(BIO_new_mem_buf(const_cast<char*>(pem_string.data()),
                                    pem_string.length()));
   if (!bio_in) {
     LOG_OPENSSL_ERRORS(ERROR);
-    return;
+    return nullptr;
   }
 
-  x509_.reset(PEM_read_bio_X509(bio_in.get(), nullptr, nullptr, nullptr));
+  ScopedX509 x509(PEM_read_bio_X509(bio_in.get(), nullptr, nullptr, nullptr));
 
-  if (!x509_) {
+  if (!x509) {
     // At this point most likely the input was just corrupt. There are a few
     // real errors that may have happened (a malloc failure is one) and it is
     // virtually impossible to fish them out.
     LOG(WARNING) << "Input is not a valid PEM-encoded certificate";
     LOG_OPENSSL_ERRORS(WARNING);
+    return nullptr;
   }
+
+  return unique_ptr<Cert>(new Cert(move(x509)));
 }
 
 
--- a/cpp/log/cert.cc
+++ b/cpp/log/cert.cc
@@ -173,7 +173,13 @@ string ASN1ToStringAndCheckForNulls(ASN1_STRING* asn1_string,
 }
 
 
+unique_ptr<Cert> Cert::FromX509(ScopedX509 x509) {
+  return x509 ? unique_ptr<Cert>(new Cert(move(x509))) : nullptr;
+}
+
+
 Cert::Cert(ScopedX509 x509) : x509_(move(x509)) {
+  CHECK(x509_);
 }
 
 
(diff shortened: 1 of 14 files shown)


cpp/log/cert.cc
```diff
@@ -173,28 +173,35 @@ string ASN1ToStringAndCheckForNulls(ASN1_STRING* asn1_string,
 }
 
 
+unique_ptr<Cert> Cert::FromX509(ScopedX509 x509) {
+  return x509 ? unique_ptr<Cert>(new Cert(move(x509))) : nullptr;
+}
+
+
 Cert::Cert(ScopedX509 x509) : x509_(move(x509)) {
+  CHECK(x509_);
```
pphaneuf: I verified that I had caught all call sites of this by deleting this constructor entirely, and replacing the body of `Cert::FromX509` (the only expected call site, which avoids the `CHECK` by checking `x509` first) with `return nullptr;`, which compiled just fine.

cpp/log/cms_verifier.cc
```diff
@@ -228,15 +228,15 @@ Cert* CmsVerifier::UnpackCmsSignedCertificate(const string& cms_object) {
...
   if (UnpackCmsDerBio(source_bio.get(), unpacked_bio.get()).ok()) {
     // The unpacked data should be a valid DER certificate.
     // TODO: The RFC does not yet define this as the format so this may
     // need to change.
-    const Status status = cert->LoadFromDerBio(unpacked_bio.get());
+    cert = Cert::FromDerBio(unpacked_bio.get());
 
-    if (!status.ok()) {
+    if (!cert) {
       LOG(WARNING) << "Could not unpack cert from CMS DER encoded data";
```
Martin2112: I'm suspicious that it doesn't return explicitly here but I wrote that code.

cpp/log/cert.cc
```diff
@@ -243,24 +243,15 @@ unique_ptr<Cert> Cert::FromDerBio(BIO* bio_in) {
 
 
 string Cert::PrintVersion() const {
-  if (!IsLoaded()) {
-    LOG(ERROR) << "Cert not loaded";
-    return string();
-  }
-
-  const long version(X509_get_version(x509_.get()));
+  const long version(X509_get_version(CHECK_NOTNULL(x509_.get())));
```
Martin2112: Not sure I'd expect a print version method to check fail.

cpp/log/cert.cc
```diff
@@ -271,22 +262,12 @@ string Cert::PrintSerialNumber() const {
 
 
 string Cert::PrintIssuerName() const {
-  if (!IsLoaded()) {
-    LOG(ERROR) << "Cert not loaded";
-    return string();
-  }
-
-  return PrintName(X509_get_issuer_name(x509_.get()));
+  return PrintName(X509_get_issuer_name(CHECK_NOTNULL(x509_.get())));
```
Martin2112: I'm worried about all these checks as it may make it impossible for tools to print out a cert that's not valid.

Martin2112:
I've read this through pretty carefully and it seems to be ok. I'm kind of worried that I didn't find anything wrong!

cpp/log/cms_verifier.cc
pphaneuf: Yeah, with the previous version, it would have returned a non-null `Cert*`, which would have returned false from `IsLoaded()`? But would have returned null if `UnpackCmsDerBio` had an error? Seems a bit dodgy indeed!

The new version will at least return null in both error cases, and non-null only in case of success.

cpp/log/cert.cc
pphaneuf: Well, to be fair, if it doesn't `CHECK`-fail here, it'll just segfault with a slightly less clear error message. :wink:

pphaneuf: Hopefully, the commit before that one should make it impossible for `x509_` to ever be null, as well, so this is sheer paranoia, really.

PR #1235, opened 2016-05-31
Clean better.

pphaneuf:


phad:
Did it in fact previously do no cleaning at all?

Makefile.am
```diff
@@ -19,6 +19,10 @@ BUILT_SOURCES = \
 	proto/ct.pb.cc \
 	proto/ct.pb.h
 
+CLEANFILES = \
```
phad: Is this a standard Makefile target?

pphaneuf: This is an Automake variable: https://www.gnu.org/software/automake/manual/automake.html#Clean

phad: ok thx

pphaneuf:
There was cleaning, but only of what Automake knew about (`.o` files, binaries, that sort of things). Protocol buffer generated output (`protoc` stuff) isn't known by Automake, so it wasn't cleaned, and the version.cc file that I generated with a rule nearby.

Not a big deal, we rarely do `make clean`, but if we do it, we'd expect all generated files to get deleted.

PR #1236, opened 2016-06-01
Converted PrecertChainEntry to a POJO.

pphaneuf:


java/ct.proto
```diff
@@ -51,16 +51,6 @@ message PreCert {
   optional bytes tbs_certificate = 2;
```
phad: I'm surprised you're not ditching this message at the same time, as it's so tightly coupled.

java/src/org/certificatetransparency/ctlog/PrecertChainEntry.java
```diff
@@ -0,0 +1,18 @@
+package org.certificatetransparency.ctlog;
+
+import org.certificatetransparency.ctlog.proto.Ct;
+
+
+public class PrecertChainEntry {
+  // The chain certifying the precertificate, as submitted by the CA.
+  public java.util.List<byte[]> precertificateChain;
```
phad: import java.util.List and make this 'public List'

phad:
Changes look OK but check what's up with Travis.

java/ct.proto
pphaneuf: Coming right up in 4e11c1866676edc7cb2a270309dfcfddd697b3fd. :wink:

java/src/org/certificatetransparency/ctlog/PrecertChainEntry.java
pphaneuf: Done.

pphaneuf:
Travis is hosed for us at the moment because of travis-ci/apt-source-whitelist#279... This is another case where separate repositories would help.

pphaneuf:
(I did run `ant build test` by hand, and it passed)

phad:
OK, LGTM and "LGTTravis" on its behalf.

PR #1237, opened 2016-06-01
Fix (most) warnings found with GCC.

pphaneuf:
There's still signed/unsigned comparisons in gtest/gmock themselves, so it cannot really be all fixed, you would still have to use `-Wno-sign-compare`.


pphaneuf:
For what it's worth, the Mac part of Travis still works (well, no worse than usual, it does need the occasional poke).

PR #1238, opened 2016-06-02
Simplified SCT list serialization

rolandshoemaker:
Builds off the work of #1124 but trims the public interface a bit. Since the SCT list is a mix of TLS and ASN.1 encoding styles (thanks to `cloudflare/cfssl` for reminding me of this) we can't just use `encoding/asn1`, in the places where TLS style encoding is required I've just reused existing functions (i.e. `writeVarBytes`/`writeUint`).

This also updates the test case as the previous one did not parse correctly under other SCT implementations (namely OpenSSL, the easiest way to test this is to generate a OCSP response/cert that contains the SCT extension which OpenSSL will happily parse + print). Since there are no explicit test vectors for SCT/list serialization I've just used one I generated that parsed properly with OpenSSL.

### DIFF
--- a/go/serialization.go
+++ b/go/serialization.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"container/list"
 	"crypto"
+	"encoding/asn1"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -82,7 +83,7 @@ func readVarBytes(r io.Reader, numLenBytes int) ([]byte, error) {
 		return nil, err
 	}
 	data := make([]byte, l)
-	if n, err := io.ReadFull(r, data);  err != nil {
+	if n, err := io.ReadFull(r, data); err != nil {
 		if err == io.EOF || err == io.ErrUnexpectedEOF {
 			return nil, fmt.Errorf("short read: expected %d but got %d", l, n)
 		}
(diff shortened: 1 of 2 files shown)


mehmooda:
Sorry about not finishing this. Real life got in the way. Anyway I agree to my commits being used

go/serialization.go
```diff
@@ -509,3 +512,51 @@ func SerializeSTHSignatureInput(sth SignedTreeHead) ([]byte, error) {
...
+	if sctListLen > MaxSCTListLength+2 {
+		return 0, fmt.Errorf("SCT List too large to serialize: %d", sctListLen)
+	}
+	return sctListLen, nil
+}
+
+// SerializeSCTList serializes the passed-in slice of SignedCertificateTimestamp into a
+// byte slice as a SignedCertificateTimestampList (see RFC6962 Section 3.3)
+func SerializeSCTList(scts []SignedCertificateTimestamp) ([]byte, error) {
+	buf := new(bytes.Buffer)
```
jsha: Move this declaration closer to the `Grow()` call.

go/serialization.go
```diff
@@ -509,3 +512,51 @@ func SerializeSTHSignatureInput(sth SignedTreeHead) ([]byte, error) {
 		return nil, fmt.Errorf("unsupported STH version %d", sth.Version)
 	}
 }
+
+// SCTListSerializedLength determines the length of the required buffer should a SCT List need to be serialized
```
jsha: You should either (a) specify this will always return >= 2 if error is nil, or (b) return the size of the list without the length prefix (i.e. >= 0), and add the length prefix in the calling function.

I mention this because the code below takes `size - 2`, which makes the assumption size is always >= 2.

go/serialization_test.go
```diff
@@ -436,6 +438,16 @@ func TestSerializeSCT(t *testing.T) {
 	}
 }
 
+func TestSerializeSCTList(t *testing.T) {
```
jsha: This test should test more of the error cases, like list too large, SCT to large, empty list, and maybe anything that could cause `SerializedLength` or `SerializeSCT` to fail, if it's relatively straightforward to cause those to fail (i.e. doesn't require extensive mocking).

jsha:
Generally looks good. @AlCutter could you take a look, and also maybe fix up the CLA tag? Roland tells me that both he and @mehmooda have signed the CLA.

go/serialization.go
rolandshoemaker: Fixed with (b) and added relevant comments.

AlCutter:
Looks good, but the CLA bot is choking on the merge-commit (for future reference it's better to rebase onto master); I suspect it wants me to add everyone who ever committed anything here and get their approval to (re)commit their commits :/.

I've re-built the string of commits based on `master` at [AlCutter/sctlist-trim](github.com/google/AlCutter/sctlist-trim), so I could merge that and we'll just close this with a "merged as ..." comment, or you can bump your branch to that commit (or recreate it yourself) and re-push?

If you want to bump your branch, it'll be something like:
```bash
# assuming you're on your sctlist-trim branch
git fetch github.com/google/AlCutter/sctlist-trim
git reset --hard FETCH_HEAD
```
Alternatively, you can just `rebase -i` your branch onto `master` and just `ignore` the merge commit.

rolandshoemaker:
Bumped.

pphaneuf:
@googlebot? Are you happy? 😃

pphaneuf:
So it appears there's some issues with @googlebot, where it can't always identify the authors on GitHub properly, because some of the APIs it uses only give the GitHub username? And it gets confused, due to some bugs... I checked the CLAs manually, though, and it's all fine.

@rolandshoemaker, I think it might help in the future if you went to https://cla.developers.google.com/ and updated your contact information to have your GitHub username?

PR #1240, opened 2016-06-04
Use a no-spaces JSON encoding for hashing.

gdbelvin:
Standard JSON implementations do not emit spaces when marshing objects.
This change allows other langues and libraries to compute matching
ToString values and thus properly perform hash based lookups.

### DIFF
--- a/cpp/util/json_wrapper.h
+++ b/cpp/util/json_wrapper.h
@@ -101,7 +101,7 @@ class JsonObject {
   }
 
   const char* ToString() const {
-    return json_object_to_json_string(obj_);
+    return json_object_to_json_string_ext(obj_, JSON_C_TO_STRING_PLAIN);
   }
 
   std::string DebugString() const {


pphaneuf:
"Standard" JSON implementations? Hashing JSON requires a canonical representation, and that doesn't seem to exist for JSON (spaces are just the tip of the iceberg here). That's why the XJSON server has to use the objecthash library to achieve this (which in turn has to use libICU to canonicalise the UTF-8, which can cause you to lose your mind if you try to do it yourself).

pphaneuf:
How about closing this PR, and we'll do something with #1243?

gdbelvin:
I'll update this PR to support ObjectHash, unless you'd like to do it of course :-)

pphaneuf:
Oh, feel free! :smiley: 

I'd also say that if this small tweak happens to make this particular JSON implementation compatible with another particular JSON implementation that you're using for your development, you should feel free to use your branch for your client development (just rebase it on top of `master` once in a while, shouldn't be a big deal).

So everyone can win: you don't have to do a bunch of work to make a fully-correct PR that we'll accept (and still get your own other work done), and we don't have to accept a not-really-correct hack PR. :wink:

pphaneuf:
As a reminder of potential pain, two identical JSON messages:

```json
{
   "error_message" : "Missing or invalid \"second\" parameter.",
   "success" : false
}
```

```json
{
   "error_message" : "Missing or invalid &#39;second&#39; parameter.",
   "success" : false
}
```

If you get a UTF-8 accented character that's in Latin1, there's no less than **three** different encodings for the same glyph (that I know of!), and when you think what you need is canonicalisation, it turns out there's **four** ways to do it.

This is an adventure I don't recommend getting into. :wink:

PR #1242, opened 2016-06-08
Converted PreCert to a POJO.

pphaneuf:
With the empty constructor removed. :smiley:

### DIFF
--- a/java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
+++ b/java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
@@ -9,6 +9,7 @@ import org.certificatetransparency.ctlog.MerkleAuditProof;
 import org.certificatetransparency.ctlog.MerkleTreeLeaf;
 import org.certificatetransparency.ctlog.ParsedLogEntry;
 import org.certificatetransparency.ctlog.ParsedLogEntryWithProof;
+import org.certificatetransparency.ctlog.PreCert;
 import org.certificatetransparency.ctlog.PrecertChainEntry;
 import org.certificatetransparency.ctlog.SignedEntry;
 import org.certificatetransparency.ctlog.TimestampedEntry;
@@ -168,18 +169,17 @@ public class Deserializer {
       int length = (int) readNumber(in, 3);
       signedEntry.x509 = readFixedLength(in, length);
     } else if (entryType == Ct.LogEntryType.PRECERT_ENTRY_VALUE) {
-      Ct.PreCert.Builder preCertBuilder = Ct.PreCert.newBuilder();
+      PreCert preCert = new PreCert();
 
-      byte[] arr = readFixedLength(in, 32);
-      preCertBuilder.setIssuerKeyHash(ByteString.copyFrom(arr));
+      preCert.issuerKeyHash = readFixedLength(in, 32);
 
       // set tbs certificate
-      arr = readFixedLength(in, 2);
+      byte[] arr = readFixedLength(in, 2);
       int length = (int) readNumber(in, 2);
 
-      preCertBuilder.setTbsCertificate(ByteString.copyFrom(readFixedLength(in, length)));
+      preCert.tbsCertificate = readFixedLength(in, length);
 
-      signedEntry.preCert = preCertBuilder.build();
+      signedEntry.preCert = preCert;
     } else {
       throw new SerializationException(String.format("Unknown entry type: %d", entryType));
     }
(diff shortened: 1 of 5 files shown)


java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
```diff
@@ -168,18 +169,17 @@ public static TimestampedEntry parseTimestampedEntry(InputStream in) {
...
-      Ct.PreCert.Builder preCertBuilder = Ct.PreCert.newBuilder();
+      PreCert preCert = new PreCert();
 
-      byte[] arr = readFixedLength(in, 32);
-      preCertBuilder.setIssuerKeyHash(ByteString.copyFrom(arr));
+      preCert.issuerKeyHash = readFixedLength(in, 32);
 
       // set tbs certificate
-      arr = readFixedLength(in, 2);
+      byte[] arr = readFixedLength(in, 2);
```
phad: you seem to be not using 'arr' after this line.

phad: are you deliberate skipping over 2 bytes?  if so, a better name than 'arr' could make that clear, plus a comment to say what is being skipped.

pphaneuf: I'll be honest with you here, I don't really know, I only turned it from an assignment to a declaration, as I deleted the declaration that was at line 173 before.

I think it's indeed skipping 2 bytes, no idea why. I think it might work if I just call `readFixedLength` and ignore the return value (I'd add a comment to explain what's happening, of course!)? @eranmes should know.

phad: Yeah, I think ignoring the return ought to be ok here, with a comment.
In c++ I'd do this:
  (void)readFixedLength(in, 2);
Although that style of indicating the return gets discarded seems to provoke a lot of 'opinion'.

pphaneuf: Wait a second, we're not barbarians here, `in` is an `InputStream`, and I see a, er, `skip` method? How about we use that? :wink:

phad: Sounds ideal!

pphaneuf: Hmm, that method seems to raise an exception if the underlying stream doesn't support seek, and the `readFixedLength` also handles short reads, which isn't the worst, but is also more work here...

eranmes: Seems like a bug - the TBSCertificate is immediately after the issuer key hash, so there shouldn't be any need to skip 2 bytes.
This code is from an external contributor and there's no test data as far as I can see.
Feel free to not skip these 2 bytes and I'll add test data at some point.

pphaneuf: Eep! Done.

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
```diff
@@ -168,18 +169,15 @@ public static TimestampedEntry parseTimestampedEntry(InputStream in) {
...
     } else if (entryType == Ct.LogEntryType.PRECERT_ENTRY_VALUE) {
-      Ct.PreCert.Builder preCertBuilder = Ct.PreCert.newBuilder();
+      PreCert preCert = new PreCert();
 
-      byte[] arr = readFixedLength(in, 32);
-      preCertBuilder.setIssuerKeyHash(ByteString.copyFrom(arr));
+      preCert.issuerKeyHash = readFixedLength(in, 32);
 
       // set tbs certificate
-      arr = readFixedLength(in, 2);
```
phad: You'd better not mix a bug-fix in with the refactors.  Add a TODO if you like so someone else can investigate properly (and add tests!)

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
pphaneuf: On second thought (and a bit of discussion here), I'll leave this PR to be neutral in terms of what it does (stick to what's in the description), rather than bury a possible bugfix in it. I'll send a separate PR for that.

java/src/org/certificatetransparency/ctlog/serialization/Deserializer.java
pphaneuf: I just moved my branch's head back one commit (but I've kept that commit aside for a later PR).

pphaneuf:
I think you were okay with this, only the rubber stamp needed? :wink:

PR #1248, opened 2016-06-10
Objecthash

### LINKED ISSUE #1243: Support Object Hash in XJSON server

gdbelvin:
Closes #1243

### DIFF
--- a/cpp/proto/xjson_serializer.cc
+++ b/cpp/proto/xjson_serializer.cc
@@ -41,7 +41,12 @@ SerializeResult CheckJsonFormat(const string& json) {
 
 string V1LeafData(const LogEntry& entry) {
   CHECK(entry.has_x_json_entry());
-  return entry.x_json_entry().json();
+  hash h;
+  if (!common_json_hash(entry.x_json_entry().json().c_str(), h)) {
+	LOG(ERROR) << "Faild to compute object hash: "
+	           << entry.x_json_entry().json() << ".";
+  }
+  return string(reinterpret_cast<const char*>(&h[0]), sizeof(h));
 }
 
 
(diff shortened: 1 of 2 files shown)


pphaneuf:
Hmm, this pulls in `origin/tostring`? Seems like a mistake, you'd want the other commits here rebased on top of `master`?

gdbelvin:
Finished ObjectHash integration with XJSON.  I have a client that successfully queries the log by hash in #1249

cpp/util/json_wrapper.h
```diff
@@ -101,7 +101,7 @@ class JsonObject {
   }
 
   const char* ToString() const {
-    return json_object_to_json_string(obj_);
+    return json_object_to_json_string_ext(obj_, JSON_C_TO_STRING_PLAIN);
```
pphaneuf: This change isn't going to be necessary, should be reverted.

cpp/proto/xjson_serializer.cc
```diff
@@ -41,7 +41,12 @@ SerializeResult CheckJsonFormat(const string& json) {
 
 string V1LeafData(const LogEntry& entry) {
   CHECK(entry.has_x_json_entry());
-  return entry.x_json_entry().json();
+  hash h;
+  if (!common_json_hash(entry.x_json_entry().json().c_str(), h)) {
+	LOG(ERROR) << "Faild to compute object hash: "
+	           << entry.x_json_entry().json() << ".";
+  }
+  return string(reinterpret_cast<const char*>(&h[0]), sizeof(h));
```
pphaneuf: That's a fairly different implementation than the other ones, no? So the leaf hash preimage for XJSON is... the hash of the object? I guess objecthash doesn't give you access to its normalised representation, so you have to go with what you've got?

cpp/util/json_wrapper.h
pphaneuf: Ah, it's the first commit that's sticking around. This will sort yourself out:

```
git rebase --onto ceb1244 a30ef155
```

gdbelvin: I suggest we keep the change as a step toward better cleanliness, and keeping in line with what other libraries do (I'm thinking of the Go library).

cpp/proto/xjson_serializer.cc
gdbelvin: Indeed it is quite different than the others.  I couldn't find a better place in the libraries to use object hash.  The typical pattern is Sha256(LeafData(entry)).  Where LeafData is the function you can swap out.   There's no current support for defining some function LeafHash that does this differently.

gdbelvin:
PTAL

Issue #1159, opened 2016-04-20
Golang: LogClient does not support fetching or verifying consistency proofs


RJPercival:
This should be implemented.

RJPercival:
Consistency proof fetching was added by https://github.com/google/certificate-transparency/pull/1285 and verification was added by https://github.com/google/certificate-transparency/pull/1161. Thanks @AlCutter and @gdbelvin!

PR #1249, opened 2016-06-14
[Go] Implement GetProofByHash

### LINKED ISSUE #1278: [C++, Go] Inclusion Proof fails for leaf_index = 0
Returning a leaf index of 0 is a valid CT response, per the [RFC](https://tools.ietf.org/html/draft-ietf-trans-rfc6962-bis-17#section-10.4.1), and is the index of the first entry.  

SCT:
`{"tree_size":1,"timestamp":1469742447178,"sha256_root_hash":"4OEMvu0pQwiIBmNv8x+7ivc10k6j0yTuSsmRA3ZPmes=","tree_head_signature":"BAMARTBDAiAiVLelb8WoTpu4JnCaiAVFBZyPrVuNrdXJzlHr2qy49QIfbHdVQRYnOjFsBtduv+7NubX5xhLTPEy9nKtRyueqwA=="}`

Proof By Hash:
`/ct/v1/get-proof-by-hash?hash=4OEMvu0pQwiIBmNv8x%2B7ivc10k6j0yTuSsmRA3ZPmes%3D&tree_size=1`
`{"leaf_index":0,"audit_path":[]}`

The python implementation follows the RFC algorithm and verifies, but the Go and C++ implementations do not.

gdbelvin:
This PR includes the code required to fetch inclusion proofs

Contributes to #1159

### DIFF
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -6,11 +6,13 @@ package client
 import (
 	"bytes"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"strconv"
 	"time"
 
@@ -25,6 +27,7 @@ const (
 	AddJSONPath           = "/ct/v1/add-json"
 	GetSTHPath            = "/ct/v1/get-sth"
 	GetEntriesPath        = "/ct/v1/get-entries"
+	GetProofByHashPath    = "/ct/v1/get-proof-by-hash"
 	GetSTHConsistencyPath = "/ct/v1/get-sth-consistency"
 )
 
@@ -91,6 +94,12 @@ type getEntryAndProofResponse struct {
 	LeafInput string   `json:"leaf_input"` // the entry itself
 	ExtraData string   `json:"extra_data"` // any chain provided when the entry was added to the log
 	AuditPath []string `json:"audit_path"` // the corresponding proof
+}
+
+// GetProofByHashResponse represents the JSON response to the CT get-proof-by-hash method.
+type GetProofByHashResponse struct {
+	LeafIndex int64    `json:"leaf_index"` // The 0-based index of the end entity corresponding to the "hash" parameter.
+	AuditPath [][]byte `json:"audit_path"` // An array of base64-encoded Merkle Tree nodes proving the inclusion of the chosen certificate.
 }
 
 // New constructs a new LogClient instance.
@@ -318,4 +327,15 @@ func (c *LogClient) GetSTHConsistency(ctx context.Context, first, second uint64)
 		return nil, err
 	}
 	return resp.Consistency, nil
+}
+
+// GetProofByHash returns an audit path for the hash of an SCT.
+func (c *LogClient) GetProofByHash(ctx context.Context, hash []byte, treeSize uint64) (*GetProofByHashResponse, error) {
+	b64Hash := url.QueryEscape(base64.StdEncoding.EncodeToString(hash))
+	u := fmt.Sprintf("%s%s?tree_size=%d&hash=%v", c.uri, GetProofByHashPath, treeSize, b64Hash)
+	var resp GetProofByHashResponse
+	if err := fetchAndParse(ctx, c.httpClient, u, &resp); err != nil {
+		return nil, err
+	}
+	return &resp, nil
 }
--- a/go/client/logclient_test.go
+++ b/go/client/logclient_test.go
@@ -3,6 +3,7 @@ package client
 import (
 	"bytes"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"math"
 	"net/http"
@@ -16,6 +17,14 @@ import (
 	ct "github.com/google/certificate-transparency/go"
 	"golang.org/x/net/context"
 )
+
+func dh(s string) []byte {
+	b, err := hex.DecodeString(s)
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
 
 const (
 	ValidSTHResponse = `{"tree_size":3721782,"timestamp":1396609800587,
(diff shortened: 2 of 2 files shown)


gdbelvin:
Thoughts on the best way to test this are welcome.

gdbelvin:
Added tests

pphaneuf:
Another merge crept up, these muddle up the PRs a bit, you should use something along the lines of `git fetch https://github.com/google/certificate-transparency.git master && git rebase FETCH_BASE` (could be simplified a bit, but the exact command-line would depend on what the remotes on your local repo are named).

gdbelvin:
Rebased as requested. PTAL

gdbelvin:
Rebased again. PTAL 
@pphaneuf

go/client/logclient.go
```diff
@@ -105,6 +101,19 @@ type getEntryAndProofResponse struct {
 	AuditPath []string `json:"audit_path"` // the corresponding proof
 }
 
+type getProofByHashResponse struct {
+	LeafIndex    int      `json:"leaf_index"` // The 0-based index of the end entity corresponding to the "hash" parameter.
+	AuditPath    []string `json:"audit_path"` // An array of base64-encoded Merkle Tree nodes proving the inclusion of the chosen certificate.
```
pphaneuf: Does it work if you make this field a `[][]byte` instead (and strip out the base64 decoding at line 423)?

go/client/logclient.go
```diff
@@ -105,6 +101,19 @@ type getEntryAndProofResponse struct {
...
+	Success      bool     `json:"success"`
+	ErrorMessage string   `json:"error_message"`
+}
+
+// addJSONRequest represents the JSON request body sent ot the add-json CT
```
pphaneuf: Why did this move?

go/client/logclient.go
```diff
@@ -105,6 +101,19 @@ type getEntryAndProofResponse struct {
 	AuditPath []string `json:"audit_path"` // the corresponding proof
 }
 
+type getProofByHashResponse struct {
+	LeafIndex    int      `json:"leaf_index"` // The 0-based index of the end entity corresponding to the "hash" parameter.
```
pphaneuf: We're using `int64` for leaf indices in general, is there a problem with the JSON parser that prevents using that?

go/client/logclient.go
```diff
@@ -386,3 +403,28 @@ func (c *LogClient) GetEntries(start, end int64) ([]ct.LogEntry, error) {
...
+func (c *LogClient) GetProofByHash(hash []byte, treeSize uint64) (ct.AuditPath, error) {
+	var resp getProofByHashResponse
+	b64Hash := url.QueryEscape(base64.StdEncoding.EncodeToString(hash))
+	u := fmt.Sprintf("%s%s?tree_size=%d&hash=%v", c.uri, GetProofByHashPath, treeSize, b64Hash)
+	log.Printf("%v", u)
+	err := c.fetchAndParse(u, &resp)
+	if err != nil {
+		return nil, err
+	}
+	if resp.ErrorMessage != "" {
```
pphaneuf: I think you might be disappointed by some recent developments in `fetchAndParse` (and might need a further rebase, at the minimum)...

go/serialization.go
```diff
@@ -126,10 +130,10 @@ func readASN1CertList(r io.Reader, totalLenBytes int, elementLenBytes int) ([]AS
 // Returns a non-nil error if there was a problem.
 func ReadTimestampedEntryInto(r io.Reader, t *TimestampedEntry) error {
 	var err error
```
pphaneuf: That line might not be needed anymore, with your other changes? (I like your way better!)

go/serialization.go
```diff
@@ -144,13 +148,48 @@ func ReadTimestampedEntryInto(r io.Reader, t *TimestampedEntry) error {
 		if t.PrecertEntry.TBSCertificate, err = readVarBytes(r, PreCertificateLengthBytes); err != nil {
 			return err
 		}
+	case XJSONLogEntryType:
+		if t.JSONData, err = readVarBytes(r, JSONLengthBytes); err != nil {
```
pphaneuf: But if you do that, might have to make the `=` into a `:=` as well?

pphaneuf:
Hmm... Am I missing some reference from `logclient.go`, or are the changes in `serialization.go`, `types.go`, and `signatures.go` are not related at all to what's in the description of this PR? Other change got mixed in?

go/serialization.go
gdbelvin: Kept the original scheme.  Has to do with `struct.field` not being something you can `:=` assign.

gdbelvin:
The changes in `types.go` and `serialization.go` allow a client to compute what the right hash of a given `SCT` and `json` so that the client can use the `GetByHash` function.   I decided to separate the compute hash and get-by-hash functions in case the client has the hash available directly. 

Cleaned up, rebased, updated. PTAL.

gdbelvin:
PTAL

gdbelvin:
Fixes #1278

go/client/logclient.go
```diff
@@ -85,13 +82,21 @@ type getAcceptedRootsResponse struct {
...
 	LeafInput string   `json:"leaf_input"` // the entry itself
 	ExtraData string   `json:"extra_data"` // any chain provided when the entry was added to the log
 	AuditPath []string `json:"audit_path"` // the corresponding proof
 }
 
+// GetProofByHashResponse represents the JSON response to the CT get-proof-by-hash method.
+type GetProofByHashResponse struct {
+	LeafIndex    int64    `json:"leaf_index"` // The 0-based index of the end entity corresponding to the "hash" parameter.
+	AuditPath    [][]byte `json:"audit_path"` // An array of base64-encoded Merkle Tree nodes proving the inclusion of the chosen certificate.
+	Success      bool     `json:"success"`
```
eranmes: I suggest getting rid of this and the ErrorMessage because the GetProofByHashResponse should not be created if there was a failure.

go/merkletree/merkle_verifier_test.go
```diff
@@ -299,9 +310,30 @@ func getConsistencyProofs() []consistencyTestVector {
...
+	googleLeaf := new(bytes.Buffer)
+	// Insert leaf cert into local CT instance:
+	// curl -X POST localhost:8088/ct/v1/add-chain -d '{"chain": [...]}'
+	// {"sct_version":0,"id":"3xwuwRUAlFJHqWFoMl3cXHlZ6PfG04j8AC4LvT9012Q=","timestamp":1469742100562,"extensions":"","signature":"BAMARjBEAiBCJXqOLwGlXdh749\/dUiWjp8AlwQUrdg2+PCqEn9lFTAIgfw8o4Q4b0YJQfpemrN4SoBNVZ5IJMAmehk\/TSZexenc="}
+	// Query by hash for the inclusion proof:
+	// curl localhost:8088/ct/v1/get-proof-by-hash?hash=4OEMvu0pQwiIBmNv8x%2B7ivc10k6j0yTuSsmRA3ZPmes%3D&tree_size=1
+	// {"leaf_index":0,"audit_path":[]}
+
+	// Compute leaf value from timestamp
+	if err := ct.SerializeV1SCTMerkleTreeLeafX509(googleLeaf, uint64(1469742100562), googlePEM); err != nil {
```
eranmes: I believe you could use test data to load certificate, SCT from here:
https://github.com/google/certificate-transparency/tree/master/test/testdata
The test will then be much simpler and will only test that SerializeV1SCTMerkleTreeLeafX509 correctly serializes.

gdbelvin:
@pphaneuf @taknira 

Cleaned up this PR, and made it exclusively about fetching the inclusion proof.

go/types.go
```diff
@@ -319,6 +319,7 @@ type TimestampedEntry struct {
 	X509Entry    ASN1Cert
 	JSONData     []byte
 	PrecertEntry PreCert
+	JSONData     []byte
```
pphaneuf: Same as the one at line 320? This causes a build error. :wink:

gdbelvin:
Fixed

go/client/logclient.go
```diff
@@ -308,3 +317,12 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
 	sth.TreeHeadSignature = *ds
 	return
 }
+
+// GetProofByHash returns an audit path for the hash of an SCT.
+func (c *LogClient) GetProofByHash(hash []byte, treeSize uint64) (GetProofByHashResponse, error) {
```
pphaneuf: Big enough that `GetProofByHashResponse` should be returned as a pointer?

pphaneuf: That would also allow returning `nil` in the case of an error.

go/client/logclient.go
```diff
@@ -308,3 +317,12 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
...
+	var resp GetProofByHashResponse
```
pphaneuf: I'd move that to right before it's needed (before `fetchAndParse`).

go/client/logclient.go
pphaneuf: We have some historical baggage, but might as well take a `context.Context` as the first parameter on new methods?

go/types.go
```diff
@@ -317,8 +317,8 @@ type TimestampedEntry struct {
 	Timestamp    uint64
 	EntryType    LogEntryType
 	X509Entry    ASN1Cert
-	JSONData     []byte
 	PrecertEntry PreCert
+	JSONData     []byte
```
pphaneuf: Maybe you could move it up to where it was, to avoid the noise?

pphaneuf:
Done with this round of review, should be straightforward... :smiley:

go/client/logclient.go
```diff
@@ -319,10 +319,12 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
 }
 
 // GetProofByHash returns an audit path for the hash of an SCT.
-func (c *LogClient) GetProofByHash(hash []byte, treeSize uint64) (GetProofByHashResponse, error) {
-	var resp GetProofByHashResponse
+func (c *LogClient) GetProofByHash(ctx context.Context, hash []byte, treeSize uint64) (*GetProofByHashResponse, error) {
+	resp := new(GetProofByHashResponse)
```
pphaneuf: Oops, you've seen my comment in #1285? I like `var` better than `new`: it seems there's no case where `new` does anything that you can't do with `var` and `&`, so I'd rather have just one way of doing things, for consistency, and let the compiler do the escape analysis...

And either way, I'd still want it declared "at the last moment". :wink:

go/client/logclient_test.go
```diff
@@ -275,7 +275,7 @@ func TestGetProofByHash(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		resp, err := c.GetProofByHash(tc.hash, tc.treesize)
+		resp, err := c.GetProofByHash(context.TODO(), tc.hash, tc.treesize)
```
pphaneuf: I guess that in a test, it should be `context.Background()`? Not like we'll refactor in the future to get the context from somewhere else, right?

go/serialization.go
```diff
@@ -147,6 +147,10 @@ func ReadTimestampedEntryInto(r io.Reader, t *TimestampedEntry) error {
 		if t.PrecertEntry.TBSCertificate, err = readVarBytes(r, PreCertificateLengthBytes); err != nil {
 			return err
 		}
+	case XJSONLogEntryType:
```
pphaneuf: I kind of missed this in the previous round, but, er, this has nothing to do with "implementing GetProofByHash", doesn't it? :wink:

gdbelvin: Would you like it moved to a different PR? It was so nice to sneak it into this one !

pphaneuf: Yeah, this is making this PR all icky! :laughing:

pphaneuf:
LGTM, waiting on Travis to merge.

pphaneuf:
Will need a rebase (and you can squash into one commit, if you want).

gdbelvin:
Rebased on master

PR #1251, opened 2016-06-16
Workaround LLVM's feeble APT mirror being down.

AlCutter:


AlCutter:
The build still breaks because of `-Wall -Werror` barfing in gflags, but I believe @pphaneuf is sorting that out.

PR #1253, opened 2016-06-16
Ignore clang warnings due to unused stuff caused by glog

AlCutter:


pphaneuf:
Looks... tolerable to me?

pphaneuf:
Makes me think of Windows code.

PR #1258, opened 2016-06-23
6962-bis: Implement OID parsing/serialization.

eranmes:
For implementing 6962-bis we'll need a way to decode DER-encoded OIDs and encode them (for later TLS serialization).

I had some hand-rolled code to do this, but using OpenSSL is probably safer. I hope.

### DIFF
--- a/cpp/proto/serializer_v2.cc
+++ b/cpp/proto/serializer_v2.cc
@@ -1,0 +1,117 @@
+/* -*- indent-tabs-mode: nil -*- */
+#include "proto/serializer_v2.h"
+
+#include <glog/logging.h>
+
+#include "util/status.h"
+#include "util/openssl_util.h"
+#include "util/openssl_scoped_types.h"
+
+#if defined(OPENSSL_IS_BORINGSSL)
+#include <openssl/asn1.h>
+#endif
+
+namespace rfc6962_bis {
+using util::StatusOr;
+using util::Status;
+using cert_trans::ScopedOpenSSLBytes;
+
+namespace {
+const char kDerOIDTag = 0x06;
+// OpenSSL documentation recommends 80, see the BUGS section in:
+// https://www.openssl.org/docs/manmaster/crypto/OBJ_obj2txt.html
+const size_t kTextOIDMaxSize = 80;
+}
+
+OID::OID() : oid_(nullptr) {
+}
+
+OID::OID(ASN1_OBJECT* oid) : oid_(oid) {
+  CHECK_NOTNULL(oid_);
+}
+
+OID::OID(const OID& other) : oid_(nullptr) {
+  if (other.oid_ != nullptr) {
+    oid_ = OBJ_dup(other.oid_);
+  }
+}
+
+OID::~OID() {
+  if (oid_ != nullptr) {
+    ASN1_OBJECT_free(oid_);
+  }
+}
+
+util::StatusOr<std::string> OID::ToTagMissingDER() const {
+  if(oid_ == nullptr) {
+    // Uninitialized.
+    return Status(util::error::INVALID_ARGUMENT,
+                  std::string("OID not initialized."));
+  }
+  int encoded_length = i2d_ASN1_OBJECT(oid_, NULL);
+  if (encoded_length <= 0) {
+    return Status(util::error::INVALID_ARGUMENT,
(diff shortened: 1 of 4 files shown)


cpp/proto/serializer_v2.cc
```diff
@@ -0,0 +1,108 @@
...
+OID::~OID() {
+  if (oid_ != nullptr) {
+    ASN1_OBJECT_free(oid_);
+  }
+}
+
+std::string OID::ToTagMissingDER() const {
+  if(oid_ == nullptr) {
+    // Uninitialized.
+    return "";
```
Martin2112: I'm suspicious if this should be a StatusOr<string>. Is it only used to display values? If we did get an erroneous value we wouldn't be able to display it.

cpp/proto/serializer_v2.cc
```diff
@@ -0,0 +1,108 @@
...
+
+  unsigned char* encoded_oid =
+      static_cast<unsigned char*>(OPENSSL_malloc(encoded_length));
+  // i2d_ASN1_OBJECT will change the pointer, so have to use a temporary.
+  unsigned char* tmp_ptr = encoded_oid;
+  encoded_length = i2d_ASN1_OBJECT(oid_, &tmp_ptr);
+  if (encoded_length <= 0) {
+    LOG(WARNING) << "Cannot encode OID, OpenSSL error: "
+                 << util::DumpOpenSSLErrorStack();
+    return "";
```
Martin2112: leaks encoded_oid

cpp/proto/serializer_v2.cc
```diff
@@ -0,0 +1,108 @@
...
+  OPENSSL_free(encoded_oid);
+  return std::move(out);
+}
+
+std::string OID::ToString() const {
+  if (oid_ == nullptr) {
+    return "";
+  }
+
+  const size_t kOutputSize = 100;
```
Martin2112: Where did 100 come from? Is that a safe limit? Move up to top level?

cpp/proto/serializer_v2.cc
eranmes: No, that will be used to encode an OID to the form that'll later be used to TLS serialize - in a V2 SCT, for example.
However since these are never supposed to be empty the caller can just check if the string is empty(). Do you think that's good enough?

cpp/proto/serializer_v2.cc
eranmes: Oh, haven't realized we've got ScopedOpenSSLString. Nice.

cpp/proto/serializer_v2.cc
eranmes: OpenSSL recommends 80 - so I've changed and added a reference to the manpage recommending the size (https://www.openssl.org/docs/manmaster/crypto/OBJ_obj2txt.html). Also moved to the top level.

eranmes:
@Martin2112 PTAL.

cpp/proto/serializer_v2.cc
Martin2112: The concern was that we're hiding the error if the encode fails. I'm not sure it can fail but if it does we're possibly doing the wrong thing.

cpp/proto/serializer_v2.cc
```diff
@@ -0,0 +1,115 @@
...
+    const std::string& missing_tag_oid_der) {
+  std::string oid_der = std::string(1, kDerOIDTag) + missing_tag_oid_der;
+  const unsigned char* data_ptr =
+      reinterpret_cast<const unsigned char*>(oid_der.data());
+
+  ASN1_OBJECT *oid = d2i_ASN1_OBJECT(NULL, &data_ptr, oid_der.size());
+
+  if (oid == nullptr) {
+    return Status(util::error::INVALID_ARGUMENT,
+                  std::string("Bad DER: ") + util::DumpOpenSSLErrorStack());
```
Martin2112: "Bad DER in OID" would be more specific.

eranmes: Done.

Martin2112:
The remaining issue is FromTagMissingDER. There's two ways it can fail but it doesn't return an error if they happen. Does that matter?

eranmes:
As discussed switched to StatusOr for ToTagMissingDer, @Martin2112 PTAL.

PR #1259, opened 2016-06-24
Add additonal stats to outer FixAndLog wrapper.

taknira:


go/fixchain/fix_and_log.go
```diff
@@ -17,17 +17,20 @@ type FixAndLog struct {
...
+	alreadyPosted uint32
+	chainsSent    uint32
 }
 
 // QueueAllCertsInChain adds every cert in the chain and the chain to the queue
 // to be fixed and logged.
 func (fl *FixAndLog) QueueAllCertsInChain(chain []*x509.Certificate) {
 	if chain != nil {
 		atomic.AddUint32(&fl.queued, 1)
+		atomic.AddUint32(&fl.chainsQueued, uint32(len(chain)))
```
phad: len(chain) is the number of certs in the chain you've been passed.
the docs for chainsQueued say it counts the original chain and sub-chains from intermediate certs
will that always be == len(chain) ?  If so, fine. (but make it clear in comments, since the code here looks like a mistake)

taknira: Yes it will be, because it submits each cert in the chain.  I have altered the comments to reflect this :)

pphaneuf:
LGTM

PR #1260, opened 2016-06-24
Do not use std::move in "return" statements.

pphaneuf:


Martin2112:
C++ eh? LGTM if travis happy.

PR #1264, opened 2016-07-05
Have the caller supply the RoundTripper used in logclient.go.

phad:
This permits callers to use a RoundTripper implementation different
to github.com/mreiferson/go-httpclient if they want to, and removes
the import dependency on that library.

### DIFF
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -17,7 +17,6 @@ import (
 	"time"
 
 	"github.com/google/certificate-transparency/go"
-	"github.com/mreiferson/go-httpclient"
 	"golang.org/x/net/context"
 )
 
@@ -109,18 +108,9 @@ type getEntryAndProofResponse struct {
 // New constructs a new LogClient instance.
 // |uri| is the base URI of the CT log instance to interact with, e.g.
 // http://ct.googleapis.com/pilot
-func New(uri string) *LogClient {
-	var c LogClient
-	c.uri = uri
-	transport := &httpclient.Transport{
-		ConnectTimeout:        10 * time.Second,
-		RequestTimeout:        30 * time.Second,
-		ResponseHeaderTimeout: 30 * time.Second,
-		MaxIdleConnsPerHost:   10,
-		DisableKeepAlives:     false,
-	}
-	c.httpClient = &http.Client{Transport: transport}
-	return &c
+// |hc| is the underlying client to be used for HTTP requests to the CT log.
+func New(uri string, hc *http.Client) *LogClient {
+	return &LogClient{uri: uri, httpClient: hc}
 }
 
 // Makes a HTTP call to |uri|, and attempts to parse the response as a JSON
--- a/go/preload/main/preload.go
+++ b/go/preload/main/preload.go
@@ -7,14 +7,17 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
 	"regexp"
 	"sync"
+	"time"
 
 	"github.com/google/certificate-transparency/go"
 	"github.com/google/certificate-transparency/go/client"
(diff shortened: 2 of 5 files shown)


phad:
@rolandshoemaker I believe you use the Go logclient, I'd appreciate your thoughts on this PR too. Thanks.

go/client/logclient.go
```diff
@@ -109,17 +108,12 @@ type getEntryAndProofResponse struct {
 // New constructs a new LogClient instance.
 // |uri| is the base URI of the CT log instance to interact with, e.g.
 // http://ct.googleapis.com/pilot
-func New(uri string) *LogClient {
+// |rt| is the underlying RoundTripper to be used as transport for HTTP
+// requests and responses.
+func New(uri string, rt http.RoundTripper) *LogClient {
```
pphaneuf: Why not just pass in a `*http.Client`?

rolandshoemaker: Agreed, this seems like a better approach.

phad: Yep, good idea, I'll do that instead.

go/client/logclient.go
```diff
@@ -109,18 +108,12 @@ type getEntryAndProofResponse struct {
...
-		DisableKeepAlives:     false,
-	}
-	c.httpClient = &http.Client{Transport: transport}
-	return &c
+// |hc| is the underlying client to be used for HTTP requests to the CT log.
+func New(uri string, hc *http.Client) *LogClient {
+	var lc LogClient
+	lc.uri = uri
+	lc.httpClient = hc
+	return &lc
```
pphaneuf: How about initializing `lc` with a literal:

```go
var lc LogClient{
        uri: uri
        httpClient: hc
}
```

You could even skip `lc` altogether, just `return &LogClient{...}"?

phad: yes, that would be more concise now the function has so little to do.

go/preload/main/preload.go
```diff
@@ -147,7 +148,7 @@ func main() {
 		}
 	}()
 
-	fetchLogClient := client.New(*sourceLogUri)
+	fetchLogClient := client.New(*sourceLogUri, &http.Client{})
```
pphaneuf: To preserve the semantics, this should now set up the transport in a similar way to what was done before?

go/scanner/main/scanner.go
```diff
@@ -96,7 +97,7 @@ func createMatcherFromFlags() (scanner.Matcher, error) {
 
 func main() {
 	flag.Parse()
-	logClient := client.New(*logUri)
+	logClient := client.New(*logUri, &http.Client{})
```
pphaneuf: Here too, set up the transport? I think this particular call site is why we came to use that helper library, IIRC...

go/preload/main/preload.go
phad: Done in 5c71eea

go/scanner/main/scanner.go
phad: OK, good to know: Done in 5c71eea

go/preload/main/preload.go
```diff
@@ -171,7 +182,7 @@ func main() {
 	sctWriterWG.Add(1)
 	go sctWriterJob(addedCerts, sctWriter, &sctWriterWG)
 
-	submitLogClient := client.New(*targetLogUri)
+	submitLogClient := client.New(*targetLogUri, &http.Client{})
```
pphaneuf: There's another here, but I *think* you might be okay if you reused the `httpclient.Transport`?

pphaneuf:
Just the one thing left.

phad:
Thanks, I missed that one.

pphaneuf:
LGTM

PR #1265, opened 2016-07-08
Golang: Initialize http.Client if passed nil in client.New

rolandshoemaker:
If the `http.Client` passed to `go/client.New` is nil initialize it before passing `client.LogClient` back to the user (general Golang stdlib behavior as far as I can tell).


rolandshoemaker:
cc @pphaneuf / @phad: super short follow-up to #1264.

pphaneuf:
LGTM

PR #1267, opened 2016-07-12
Run everything through gofmt.

pphaneuf:
Using the "simplify code" option, because more simple is more better!


taknira:
It's weird that the fixchain stuff has changes when run through gofmt, as I ran it all through gofmt when it was written.  In fact, when I run gofmt on it even now, none of these changes appear... are you using some swanky version of gofmt?

pphaneuf:
As mentioned in the description, I used the -s ("simplify code") option. Without it, the changes were much more limited, but I really like simpler code. 😃

taknira:
My bad for not thoroughly reading the description.  All good :)

PR #1268, opened 2016-07-13
Check for HTTP status 200 in fetchAndParse.

pphaneuf:
Also use a streaming JSON decoder for `fetchAndParse`, and ensure that the body is read to completion, so the connection can be reused.

### DIFF
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -125,20 +125,21 @@ func (c *LogClient) fetchAndParse(uri string, res interface{}) error {
 		return err
 	}
 	resp, err := c.httpClient.Do(req)
-	var body []byte
-	if resp != nil {
-		body, err = ioutil.ReadAll(resp.Body)
-		resp.Body.Close()
-		if err != nil {
-			return err
-		}
-	}
 	if err != nil {
 		return err
 	}
-	if err = json.Unmarshal(body, &res); err != nil {
+	defer resp.Body.Close()
+	// Make sure everything is read, so http.Client can reuse the connection.
+	defer ioutil.ReadAll(resp.Body)
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("got HTTP Status %s", resp.Status)
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(res); err != nil {
 		return err
 	}
+
 	return nil
 }
 


go/client/logclient.go
```diff
@@ -125,20 +125,21 @@ func (c *LogClient) fetchAndParse(uri string, res interface{}) error {
...
 	if err != nil {
 		return err
 	}
-	if err = json.Unmarshal(body, &res); err != nil {
+	defer resp.Body.Close()
+	// Make sure everything is read, so http.Client can reuse the connection.
+	defer ioutil.ReadAll(resp.Body)
+
+	if resp.StatusCode != 200 {
+		return errors.New("status code not 200")
```
phad: a better message might say what the status code actually was

go/client/logclient.go
```diff
@@ -125,20 +125,21 @@ func (c *LogClient) fetchAndParse(uri string, res interface{}) error {
...
-			return err
-		}
-	}
 	if err != nil {
 		return err
 	}
-	if err = json.Unmarshal(body, &res); err != nil {
+	defer resp.Body.Close()
+	// Make sure everything is read, so http.Client can reuse the connection.
+	defer ioutil.ReadAll(resp.Body)
```
phad: is the point here that if json.NewDecoder...Decode doesn't run (or terminates prematurely) that the deferred 'ReadAll' gets to the end of the input regardless?

is that a noop in the case of JSON parsing out successfully?

go/client/logclient.go
pphaneuf: Good point, done.

go/client/logclient.go
pphaneuf: That's correct. In the case of the HTTP status not being 200, the JSON decoding won't even be attempted, so if we want to reuse the connection, we need to slurp whatever body came in. And when it is attempted, there might be trailing cruft at the end, and if the JSON decoding got everything, it just won't get anything.

go/client/logclient.go
```diff
@@ -133,7 +133,7 @@ func (c *LogClient) fetchAndParse(uri string, res interface{}) error {
 	defer ioutil.ReadAll(resp.Body)
 
 	if resp.StatusCode != 200 {
-		return errors.New("status code not 200")
+		return errors.New("status code not 200: " + strconv.Itoa(resp.StatusCode))
```
phad: might fmt.Sprintf() be a bit shorter?

pphaneuf: `fmt.Sprintf()` was exactly the same, but there's a `fmt.Errorf()` that seems to fit perfectly. I made the message similar to another one at line 237, while I was there, for consistency.

go/client/logclient.go
phad: Ah yes, that's ideal.

go/client/logclient.go
phad: Yeah, that's better.

PR #1269, opened 2016-07-13
refactoring of the Go log client

pphaneuf:
Best looked at one commit at a time. :wink:

### DIFF
--- a/go/client/getentries.go
+++ b/go/client/getentries.go
@@ -5,6 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
 
 	ct "github.com/google/certificate-transparency/go"
 	"golang.org/x/net/context"
@@ -28,8 +31,19 @@ func getRawEntries(ctx context.Context, httpClient *http.Client, logURL string, 
 	if end < start {
 		return nil, errors.New("start should be <= end")
 	}
+
+	baseURL, err := url.Parse(strings.TrimRight(logURL, "/") + GetEntriesPath)
+	if err != nil {
+		return nil, err
+	}
+
+	baseURL.RawQuery = url.Values{
+		"start": []string{strconv.FormatInt(start, 10)},
+		"end":   []string{strconv.FormatInt(end, 10)},
+	}.Encode()
+
 	var resp getEntriesResponse
-	err := fetchAndParse(context.TODO(), httpClient, fmt.Sprintf("%s%s?start=%d&end=%d", logURL, GetEntriesPath, start, end), &resp)
+	err = fetchAndParse(context.TODO(), httpClient, baseURL.String(), &resp)
 	if err != nil {
 		return nil, err
 	}
--- a/go/client/getentries.go
+++ b/go/client/getentries.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	ct "github.com/google/certificate-transparency/go"
+	"golang.org/x/net/context"
 )
 
 // leafEntry respresents a JSON leaf entry
@@ -29,7 +30,7 @@ func (c *LogClient) GetEntries(start, end int64) ([]ct.LogEntry, error) {
 		return nil, errors.New("start should be <= end")
 	}
 	var resp getEntriesResponse
-	err := fetchAndParse(c.httpClient, fmt.Sprintf("%s%s?start=%d&end=%d", c.uri, GetEntriesPath, start, end), &resp)
+	err := fetchAndParse(context.TODO(), c.httpClient, fmt.Sprintf("%s%s?start=%d&end=%d", c.uri, GetEntriesPath, start, end), &resp)
 	if err != nil {
 		return nil, err
 	}
--- a/go/client/getentries.go
+++ b/go/client/getentries.go
@@ -1,0 +1,65 @@
+package client
+
+import (
+	"bytes"
+	"encoding/base64"
+	"errors"
+	"fmt"
+
+	ct "github.com/google/certificate-transparency/go"
+)
+
+// base64LeafEntry respresents a Base64 encoded leaf entry
+type base64LeafEntry struct {
+	LeafInput string `json:"leaf_input"`
+	ExtraData string `json:"extra_data"`
+}
+
+// getEntriesReponse respresents the JSON response to the CT get-entries method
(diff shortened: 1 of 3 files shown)


go/client/getentries.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// getEntriesReponse respresents the JSON response to the CT get-entries method
+type getEntriesResponse struct {
+	Entries []base64LeafEntry `json:"entries"` // the list of returned entries
+}
+
+// GetEntries attempts to retrieve the entries in the sequence [|start|, |end|] from the CT log server. (see section 4.6.)
+// Returns a slice of LeafInputs or a non-nil error.
+func (c *LogClient) GetEntries(start, end int64) ([]ct.LogEntry, error) {
```
phad: I'm going to assume this block is unchanged from the original in logclient.go -- github doesn't make checking that easy.

go/client/getentries.go
```diff
@@ -36,21 +35,19 @@ func (c *LogClient) GetEntries(start, end int64) ([]ct.LogEntry, error) {
 	}
 	entries := make([]ct.LogEntry, len(resp.Entries))
 	for index, entry := range resp.Entries {
-		leafBytes, err := base64.StdEncoding.DecodeString(entry.LeafInput)
-		leaf, err := ct.ReadMerkleTreeLeaf(bytes.NewBuffer(leafBytes))
+		leaf, err := ct.ReadMerkleTreeLeaf(bytes.NewBuffer(entry.LeafInput))
```
phad: this is a nice change.

go/client/logclient.go
```diff
@@ -108,11 +108,12 @@ func New(uri string, hc *http.Client) *LogClient {
 // Makes a HTTP call to |uri|, and attempts to parse the response as a JSON
```
phad: mention the new argument |ctx| in the docs.

go/client/getentries.go
```diff
@@ -29,7 +30,7 @@ func (c *LogClient) GetEntries(start, end int64) ([]ct.LogEntry, error) {
 		return nil, errors.New("start should be <= end")
 	}
 	var resp getEntriesResponse
-	err := fetchAndParse(c.httpClient, fmt.Sprintf("%s%s?start=%d&end=%d", c.uri, GetEntriesPath, start, end), &resp)
+	err := fetchAndParse(context.TODO(), c.httpClient, fmt.Sprintf("%s%s?start=%d&end=%d", c.uri, GetEntriesPath, start, end), &resp)
```
phad: I like it!  Presumably one should continue from here by propagating the context back through the call stack until you reach a frame where the caller has a context to hand, or can create one, then the TODO goes away?

go/client/getentries.go
```diff
@@ -20,17 +21,26 @@ type getEntriesResponse struct {
 	Entries []leafEntry `json:"entries"` // the list of returned entries
 }
 
-// GetEntries attempts to retrieve the entries in the sequence [|start|, |end|] from the CT log server. (see section 4.6.)
-// Returns a slice of LeafInputs or a non-nil error.
-func (c *LogClient) GetEntries(start, end int64) ([]ct.LogEntry, error) {
+func rawGetEntries(ctx context.Context, httpClient *http.Client, logURL string, start, end int64) (*getEntriesResponse, error) {
```
phad: 'rawGetEntries' is a slightly odd name.  getRawEntries,  getEntriesRaw?
'getEntriesResponse' would be nice but it's the name of a type so that'd be confusing.

go/client/getentries.go
```diff
@@ -28,8 +32,21 @@ func rawGetEntries(ctx context.Context, httpClient *http.Client, logURL string,
...
 		return nil, errors.New("start should be <= end")
 	}
+
+	baseURL, err := url.Parse(strings.TrimRight(logURL, "/") + GetEntriesPath)
+	if err != nil {
+		return nil, err
+	}
+
+	baseURL.RawQuery = url.Values{
+		"start": []string{strconv.FormatInt(start, 10)},
```
phad: is it generally known that FormatInt's 2nd arg is the number base?  I just had to look it up.

go/client/getentries.go
```diff
@@ -28,8 +32,21 @@ func rawGetEntries(ctx context.Context, httpClient *http.Client, logURL string,
...
+		"end":   []string{strconv.FormatInt(end, 10)},
+	}.Encode()
+
+	log.Printf("baseURL: %v", baseURL.String())
```
phad: %v -> %s since you know it's a string.

go/client/getentries.go
pphaneuf: I did reformat the comment at line 23, but that's it, no other changes (in that commit).

go/client/getentries.go
pphaneuf: It seems like it's one of those things like `strtoul` that you have to look up every damn time you see it... There was a straightforward-looking `Itoa`, but it takes an `int`, not an `int64`. :stuck_out_tongue_closed_eyes:

go/client/getentries.go
pphaneuf: Oops, that's actually a log I had in there while debugging, I should have removed it!

go/client/getentries.go
pphaneuf: This is the one I was telling you it might be possible to do more of it, but I only did it for the part I was using.

go/client/getentries.go
pphaneuf: Sort of done (albeit in a commit at the end of the branch, so it might look funny in GitHub?). I'll squash as appropriate before submitting.

go/client/getentries.go
pphaneuf: That's the idea. They're not created, normally, but always handed to you, and there's two "root" contexts, the "background" context, and the "TODO" context. As far as I can tell, the only difference between the two is that the TODO one identifies a site where we'd like to refactor in order to get a context as a parameter.

pphaneuf:
Having a small problem, as I'm using some feature of the standard library that was added in Go 1.6, but the Linux builds on Travis still have 1.5... Amusingly, the Mac OS X build has 1.6, so it passes.

I'll look into updating the Go on Travis.

pphaneuf:
OK, I've got Travis sorted out! PTAL...

PR #1271, opened 2016-07-14
Extract interface LogClient and rename existing struct logClientImpl.

phad:
Using an interface makes it easier to write a mock LogClient for unit
tests.

### DIFF
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -29,8 +29,18 @@ const (
 	GetEntriesPath  = "/ct/v1/get-entries"
 )
 
-// LogClient represents a client for a given CT Log instance
-type LogClient struct {
+// LogClient defines an interface for Log Clients.  This aids testing.
+type LogClient interface {
+	AddChain(chain []ct.ASN1Cert) (*ct.SignedCertificateTimestamp, error)
+	AddPreChain(chain []ct.ASN1Cert) (*ct.SignedCertificateTimestamp, error)
+	AddChainWithContext(ctx context.Context, chain []ct.ASN1Cert) (*ct.SignedCertificateTimestamp, error)
+	AddJSON(data interface{}) (*ct.SignedCertificateTimestamp, error)
+	GetSTH() (sth *ct.SignedTreeHead, err error)
+	GetEntries(start, end int64) ([]ct.LogEntry, error)
+}
+
+// logClientImpl conforms to LogClient and represents a client for a given CT Log instance
+type logClientImpl struct {
 	uri        string       // the base URI of the log. e.g. http://ct.googleapis/pilot
 	httpClient *http.Client // used to interact with the log via HTTP
 }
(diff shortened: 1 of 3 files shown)


go/client/logclient.go
```diff
@@ -29,8 +29,18 @@ const (
...
+	AddChain(chain []ct.ASN1Cert) (*ct.SignedCertificateTimestamp, error)
+	AddPreChain(chain []ct.ASN1Cert) (*ct.SignedCertificateTimestamp, error)
+	AddChainWithContext(ctx context.Context, chain []ct.ASN1Cert) (*ct.SignedCertificateTimestamp, error)
+	AddJSON(data interface{}) (*ct.SignedCertificateTimestamp, error)
+	GetSTH() (sth *ct.SignedTreeHead, err error)
+	GetEntries(start, end int64) ([]ct.LogEntry, error)
+}
+
+// logClientImpl conforms to LogClient and represents a client for a given CT Log instance
+type logClientImpl struct {
```
pphaneuf: Go with the shorter `logClient`?

phad: I'm not massively in favour of having type names that differ only the case of a single character.

go/client/logclient.go
```diff
@@ -105,21 +115,21 @@ type getEntryAndProofResponse struct {
...
 
-// New constructs a new LogClient instance.
+// New constructs a new logClientImpl instance.
 // |uri| is the base URI of the CT log instance to interact with, e.g.
 // http://ct.googleapis.com/pilot
 // |hc| is the underlying client to be used for HTTP requests to the CT log.
-func New(uri string, hc *http.Client) *LogClient {
+func New(uri string, hc *http.Client) LogClient {
 	if hc == nil {
 		hc = new(http.Client)
```
pphaneuf: Unrelated question: I've never been clear on the difference between `new(http.Client)` and `&http.Client{}`, do you know something I don't?

go/client/logclient.go
phad: But, maybe that's the Go way.  Do you know?

go/scanner/scanner.go
```diff
@@ -150,7 +150,7 @@ func DefaultScannerOptions() *ScannerOptions {
 // Scanner is a tool to scan all the entries in a CT Log.
 type Scanner struct {
 	// Client used to talk to the CT log instance
-	logClient *client.LogClient
+	logClient client.LogClient
```
pphaneuf: How did you spot those? With `go vet`? Compiler errors?

go/client/logclient.go
phad: The Effective Go sections on naming don't really go into this btw.  A quick strawpoll of folk nearby suggests that having logClient and LogClient in one file is asking for trouble.

go/client/logclient.go
phad: I think there isn't any difference, and it comes down to individual preference.  I've seen &{...} used more though.

go/scanner/scanner.go
phad: There was nothing wrong with the original code, but the consequence of the interface extraction is that we now pass interfaces around, and this is not done using pointers in Go from what I've read.

go/client/logclient.go
phad: I checked https://golang.org/doc/effective_go.html#allocation_new and https://golang.org/doc/effective_go.html#composite_literals - the two sections read together suggest there is no difference between new and &{}

go/scanner/scanner.go
pphaneuf: Yeah, interfaces are a pointer type, I understand (when you call something that takes an `interface{}`, for example, you have to do `&foo`), but I wasn't sure if `*LogClient` would be flagged as an error by the compiler, or whether it's allowed, but `go vet` would frown a bit, or if you just searched around by hand to fix those?

PR #1279, opened 2016-07-29
Simplify TLS deserialization.

eranmes:
The TLSDeserializer class contained a mix of low-level TLS deserializing
and deserialization of application-specific data (SCTs, etc).

The TLSDeserializer was moved to tls_encoding.h/cc and was left only with
the basic TLS deserialization abilities.

De-serialization of application-specific structures remains in serializer.h
as stand-alone functions (rather than member functions).

### DIFF
--- a/cpp/proto/serializer.cc
+++ b/cpp/proto/serializer.cc
@@ -396,60 +396,112 @@ SerializeResult CheckSctExtensionsFormat(
 }
 
 
-const size_t TLSDeserializer::kV2ExtensionCountLengthInBytes = 2;
-const size_t TLSDeserializer::kV2ExtensionTypeLengthInBytes = 2;
+namespace {
 
-
-TLSDeserializer::TLSDeserializer(const string& input)
-    : current_pos_(input.data()), bytes_remaining_(input.size()) {
-}
-
-
-DeserializeResult TLSDeserializer::ReadSCTV1(SignedCertificateTimestamp* sct) {
+DeserializeResult ReadSCTV1(TLSDeserializer* deserializer,
+                            SignedCertificateTimestamp* sct) {
   sct->set_version(ct::V1);
-  if (!ReadFixedBytes(Serializer::kKeyIDLengthInBytes,
-                      sct->mutable_id()->mutable_key_id())) {
+  if (!deserializer->ReadFixedBytes(Serializer::kKeyIDLengthInBytes,
+                                    sct->mutable_id()->mutable_key_id())) {
     return DeserializeResult::INPUT_TOO_SHORT;
   }
   // V1 encoding.
   uint64_t timestamp = 0;
-  if (!ReadUint(Serializer::kTimestampLengthInBytes, &timestamp)) {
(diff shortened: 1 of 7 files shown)


cpp/proto/serializer.cc
```diff
@@ -396,60 +400,115 @@ SerializeResult CheckSctExtensionsFormat(
...
+
+namespace serialization {
+
+namespace internal {
 
-DeserializeResult TLSDeserializer::ReadSCT(SignedCertificateTimestamp* sct) {
+DeserializeResult ReadExtensionsV1(TLSDeserializer* deserializer,
+                                   ct::TimestampedEntry* entry) {
+  string extensions;
+  if (!deserializer->ReadVarBytes(Serializer::kMaxExtensionsLength,
```
pphaneuf: Maybe `CHECK_NOTNULL(deserializer)` for consistency?

eranmes: Done.

cpp/proto/serializer.h
```diff
@@ -26,71 +26,25 @@ cert_trans::serialization::SerializeResult CheckSctExtensionsFormat(
...
-  // We do not make a copy, so input must remain valid.
-  // TODO(pphaneuf): And so we should take a string *, not a string &
-  // (which could be to a temporary, and not valid once the
-  // constructor returns).
-  explicit TLSDeserializer(const std::string& input);
-
-  bool ReachedEnd() const {
-    return bytes_remaining_ == 0;
-  }
+namespace serialization {
```
pphaneuf: This should be inside of `namespace cert_trans` as well, which would remove the need for much qualifying (for example, `cert_trans::serialization::DeserializeResult` a few lines below becomes just `DeserializeResult`).

cpp/proto/serializer.h
```diff
@@ -26,71 +26,25 @@ cert_trans::serialization::SerializeResult CheckSctExtensionsFormat(
...
 
-  cert_trans::serialization::DeserializeResult ReadSCT(
-      ct::SignedCertificateTimestamp* sct);
+namespace internal {
```
pphaneuf: You can skip whitespace between consecutive `namespace` openings/closings (really, it'd be nicer if we could just go `namespace serialization::internal {`, but here we are).

cpp/proto/serializer.cc
```diff
@@ -106,6 +107,9 @@ SerializeResult Serializer::SerializeV1STHSignatureInput(
 }
 
 
+////////////////////////////////////////////////////////////////////////////////
+//                                SERIALIZATION //
```
pphaneuf: If you're going to make funny boxes like that, you should do them properly? :wink: 

(the `//` on the right side should go further? I don't really care much for those kinds of markers, makes me think things should be in a different file when I see them)

eranmes:
All namespace comments do not seem to be applicable anymore now that I've merged your commit. PTAL.

pphaneuf:
LGTM

eranmes:
Thanks, merging now (as you noted, will have to rebase).

PR #1281, opened 2016-08-02
Add X509 Serialization functions

gdbelvin:
This PR also includes the functions needed to
serialize a MerkleTreeLeaf for X509 leaf entries.

### DIFF
--- a/go/serialization.go
+++ b/go/serialization.go
@@ -151,6 +151,33 @@ func ReadTimestampedEntryInto(r io.Reader, t *TimestampedEntry) error {
 	return nil
 }
 
+// SerializeTimestampedEntry writes timestamped entry to Writer.
+func SerializeTimestampedEntry(w io.Writer, t *TimestampedEntry) error {
+	if err := binary.Write(w, binary.BigEndian, t.Timestamp); err != nil {
+		return err
+	}
+	if err := binary.Write(w, binary.BigEndian, t.EntryType); err != nil {
+		return err
+	}
+	switch t.EntryType {
+	case X509LogEntryType:
+		if err := writeVarBytes(w, t.X509Entry, CertificateLengthBytes); err != nil {
+			return err
+		}
+	case PrecertLogEntryType:
+		if err := binary.Write(w, binary.BigEndian, t.PrecertEntry.IssuerKeyHash); err != nil {
+			return err
+		}
+		if err := writeVarBytes(w, t.PrecertEntry.TBSCertificate, PreCertificateLengthBytes); err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unknown EntryType: %d", t.EntryType)
+	}
+	writeVarBytes(w, t.Extensions, ExtensionsLengthBytes)
+	return nil
+}
+
 // ReadMerkleTreeLeaf parses the byte-stream representation of a MerkleTreeLeaf
 // and returns a pointer to a new MerkleTreeLeaf structure containing the
 // parsed data.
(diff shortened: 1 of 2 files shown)


go/merkletree/merkle_verifier_test.go
```diff
@@ -299,9 +299,18 @@ func getConsistencyProofs() []consistencyTestVector {
 	}
 }
 
-func TestVerifyInclusionProof(t *testing.T) {
+func TestVerifyInclusionProofAgainstRealData(t *testing.T) {
 	v := getVerifier()
+	// Serialized MerkleTreeLeaf from test-cert.pem and test-cert.proof
```
eranmes: Could you load these from the testdata directory?

eranmes:
Assigning to Martin for Go review.
My observations:
- The fix to the test & algorithm looks good.
- Serialization code looks good.

I'd suggest splitting this CL into two:
- First, submit the leaf-building code.
- Second, use the leaf-building code in the fix to the inclusion verification code. So rather than hard-coding the leaf in the tests twice, the serialization code will have it hard-coded for comparison, and the inclusion check test will load the SCT and cert from the testdata directory and will use the (now-tested) leaf building code to get the leaf.

go/merkletree/merkle_verifier.go
```diff
@@ -47,14 +47,14 @@ func (m MerkleVerifier) VerifyInclusionProof(leafIndex, treeSize int64, proof []
 
 // RootFromInclusionProof calculates the expected tree root given the proof and leaf.
```
eranmes: Please add documentation that this code expects 1-based leafIndex.

go/merkletree/merkle_verifier_test.go
```diff
@@ -299,9 +299,18 @@ func getConsistencyProofs() []consistencyTestVector {
 	}
 }
 
-func TestVerifyInclusionProof(t *testing.T) {
+func TestVerifyInclusionProofAgainstRealData(t *testing.T) {
```
eranmes: Nit: Test name: Seems to me that what's important in this test is not that it's using real data but that it tests the inclusion proof checking code for the case where there's exactly one entry in the tree. Rename the test accordingly.

go/serialization.go
```diff
@@ -151,6 +151,33 @@ func ReadTimestampedEntryInto(r io.Reader, t *TimestampedEntry) error {
 	return nil
 }
 
+// SerializeTimestampedEntry writes timestamped entry to Writer.
+func SerializeTimestampedEntry(w io.Writer, t *TimestampedEntry) error {
+	if err := binary.Write(w, binary.BigEndian, t.Timestamp); err != nil {
```
eranmes: Does go allow you to concatenate the condition here and in line 159? i.e.
if err:=...; err != nill || err:=...; err != nill {
  return err
}

?

go/serialization.go
```diff
@@ -561,3 +588,38 @@ func SerializeSCTList(scts []SignedCertificateTimestamp) ([]byte, error) {
...
+	}
+	if err := SerializeTimestampedEntry(w, &m.TimestampedEntry); err != nil {
+		return err
+	}
+	return nil
+}
+
+// SerializeV1SCTMerkleTreeLeafX509 write the merkle tree leaf for cert
+// to Writer.
+func SerializeV1SCTMerkleTreeLeafX509(w io.Writer, timestamp uint64, cert ASN1Cert) error {
```
eranmes: Nit: naming: The X509 at the end is redundant. There'll ever only be one type of MerkleTreeLeaf (we've removed that level of indirection in 6962-bis because we found it not to be necessary), so it's not useful to designate the type of tree leaf in the function name.

go/serialization.go
gdbelvin: The assignment in the If statement happens first. You can't `||` multiple assignment statements together.

gdbelvin:
Reduced PR to just include X509 Serialization functions.
@taknira @pphaneuf

go/serialization.go
```diff
@@ -561,3 +588,36 @@ func SerializeSCTList(scts []SignedCertificateTimestamp) ([]byte, error) {
 	}
 	return asn1.Marshal(buf.Bytes()) // transform to Octet String
 }
+
+// SerializeMerkleTreeLeaf writes MerkleTreeLeaf to Writer.
```
pphaneuf: Same thing here, I believe?

pphaneuf:
LGTM, merging.

go/serialization.go
```diff
@@ -151,6 +151,33 @@ func ReadTimestampedEntryInto(r io.Reader, t *TimestampedEntry) error {
 	return nil
 }
 
+// SerializeTimestampedEntry writes timestamped entry to Writer.
```
pphaneuf: Could you add a comment to mention that in case of error, `w` might have garbage on it?

PR #1284, opened 2016-08-03
Add JSON Serialization functions

gdbelvin:
Serialize XJSON Merkel Tree Leaf mimics the same encoding style as the json-c library.  This is a temporary implementation until either ObjectHash is used by the XJSON server or Trillian is finished.

### DIFF
--- a/go/serialization.go
+++ b/go/serialization.go
@@ -6,9 +6,11 @@ import (
 	"crypto"
 	"encoding/asn1"
 	"encoding/binary"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 )
 
 // Variable size structure prefix-header byte lengths
@@ -18,6 +20,7 @@ const (
 	ExtensionsLengthBytes       = 2
 	CertificateChainLengthBytes = 3
 	SignatureLengthBytes        = 2
+	JSONLengthBytes             = 3
 )
 
 // Max lengths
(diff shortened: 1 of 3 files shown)


gdbelvin:
@taknira @pphaneuf

go/serialization.go
```diff
@@ -622,3 +632,29 @@ func CreateX509MerkleTreeLeaf(cert ASN1Cert, timestamp uint64) *MerkleTreeLeaf {
...
 	}
 }
+
+// CreateJSONMerkleTreeLeaf creates the merkle tree leaf for json data.
+func CreateJSONMerkleTreeLeaf(data interface{}, timestamp uint64) *MerkleTreeLeaf {
+	jsonData, err := json.Marshal(AddJSONRequest{Data: data})
+	if err != nil {
+		return nil
+	}
+	// Match the JSON serialization implemented by json-c
```
pphaneuf: Engage cowboy mode! :laughing:

go/types.go
```diff
@@ -238,6 +241,7 @@ type LogEntry struct {
 	Leaf     MerkleTreeLeaf
 	X509Cert *x509.Certificate
 	Precert  *Precertificate
+	JSONData []byte
```
pphaneuf: Wouldn't that be better as an `interface{}` here? Your call, I'm not really sure where this gets used...

go/serialization.go
gdbelvin: Yep! This obviously doesn't handle the potentially non-deterministic way JSON entries can be ordered, but I've found that this works for now.

go/types.go
gdbelvin: This is the object that holds the serialized JSON, so it's not an `interface{}` at this point.

pphaneuf: OK.

pphaneuf:
LGTM

PR #1285, opened 2016-08-03
[Go] Add GetSTHConsistency

gdbelvin:
Adds GetSTHConsistency to Go client.

### DIFF
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -70,7 +70,7 @@ type getSTHResponse struct {
 	TreeHeadSignature []byte `json:"tree_head_signature"` // Log signature for this STH
 }
 
-// getConsistencyProofResponse represents the JSON response to the CT get-consistency-proof method
+// getConsistencyProofResponse represents the JSON response to the get-consistency-proof CT method
 type getConsistencyProofResponse struct {
 	Consistency [][]byte `json:"consistency"`
 }
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -311,9 +311,11 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
 }
 
 // GetSTHConsistency retrieves the consistency proof between two snapshots.
-func (c *LogClient) GetSTHConsistency(first, second uint64) ([][]byte, error) {
-	resp := new(getConsistencyProofResponse)
+func (c *LogClient) GetSTHConsistency(ctx context.Context, first, second uint64) ([][]byte, error) {
 	u := fmt.Sprintf("%s%s?first=%d&second=%d", c.uri, GetSTHConsistencyPath, first, second)
-	err := fetchAndParse(context.TODO(), c.httpClient, u, &resp)
-	return resp.Consistency, err
+	resp := new(getConsistencyProofResponse)
+	if err := fetchAndParse(ctx, c.httpClient, u, resp); err != nil {
+		return [][]byte{}, nil
+	}
+	return resp.Consistency, nil
 }
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -313,9 +313,9 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
 // GetSTHConsistency retrieves the consistency proof between two snapshots.
 func (c *LogClient) GetSTHConsistency(ctx context.Context, first, second uint64) ([][]byte, error) {
 	u := fmt.Sprintf("%s%s?first=%d&second=%d", c.uri, GetSTHConsistencyPath, first, second)
-	resp := new(getConsistencyProofResponse)
-	if err := fetchAndParse(ctx, c.httpClient, u, resp); err != nil {
-		return [][]byte{}, nil
+	var resp getConsistencyProofResponse
+	if err := fetchAndParse(ctx, c.httpClient, u, &resp); err != nil {
+		return nil, err
 	}
 	return resp.Consistency, nil
 }
--- a/go/client/logclient.go
+++ b/go/client/logclient.go
@@ -20,11 +20,12 @@ import (
 
 // URI paths for CT Log endpoints
 const (
-	AddChainPath    = "/ct/v1/add-chain"
-	AddPreChainPath = "/ct/v1/add-pre-chain"
-	AddJSONPath     = "/ct/v1/add-json"
-	GetSTHPath      = "/ct/v1/get-sth"
-	GetEntriesPath  = "/ct/v1/get-entries"
+	AddChainPath          = "/ct/v1/add-chain"
+	AddPreChainPath       = "/ct/v1/add-pre-chain"
+	AddJSONPath           = "/ct/v1/add-json"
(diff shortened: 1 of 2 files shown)


gdbelvin:
@pphaneuf @taknira

ChaosChoseME:
Pull Request #1285 [Go] Add GetSTHConsistency

go/client/logclient.go
```diff
@@ -71,7 +72,7 @@ type getSTHResponse struct {
 
 // getConsistencyProofResponse represents the JSON response to the CT get-consistency-proof method
```
phad: comment nit:  "to the get-consistency-proof CT method" would be consistent (ahem) with comments for other getFooResponse calls above.

go/client/logclient.go
```diff
@@ -308,3 +309,11 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
 	sth.TreeHeadSignature = *ds
 	return
 }
+
+// GetSTHConsistency retrieves the consistency proof between two snapshots.
+func (c *LogClient) GetSTHConsistency(first, second uint64) ([][]byte, error) {
```
phad: The array of byte-array return is probably fine, but did you consider instead defining a protocol buffer  containing 'repeated bytes proof' to represent the response, and returning a pointer-to-PB instead?  That would be better aligned with the other parts of the API.

go/client/logclient_test.go
```diff
@@ -34,6 +35,14 @@ const (
...
 )
 
+func b64(s string) []byte {
```
phad: nit: maybe name this func decB64 to be clearer that it decodes B64?

go/client/logclient_test.go
```diff
@@ -214,3 +223,34 @@ func TestAddJSON(t *testing.T) {
 		}
 	}
 }
+
+func TestGetSTHConsistency(t *testing.T) {
+	hs := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := w.Write([]byte(`{ "consistency": [ "IqlrapPQKtmCY1jCr8+lpCtscRyjjZAA7nyadtFPRFQ=", "ytf6K2GnSRZ3Au+YkivCb7N1DygfKyZmE4aEs9OXl\/8=" ] }`))
```
phad: you could inline the test and assignment here since the main return value from w.Write is discarded

if _, err:= w.Write( ...) ; err != nil {
}

go/client/logclient_test.go
```diff
@@ -214,3 +223,34 @@ func TestAddJSON(t *testing.T) {
...
+			b64("ytf6K2GnSRZ3Au+YkivCb7N1DygfKyZmE4aEs9OXl/8=")}},
+	}
+
+	for _, tc := range tests {
+		proof, err := c.GetSTHConsistency(tc.first, tc.second)
+		if err != nil {
+			t.Fatalf("Failed to get consistency proof: %v", err)
+		}
+		if !reflect.DeepEqual(proof, tc.proof) {
+			t.Errorf("GetSTHConsistency(): got\n%v, want\n%v", proof, tc.proof)
```
phad: I was advised in one of my reviews to reproduce the input in the error message to make the test failure messages more useful when writing tabulated tests.

t.Errorf("GetSTHConsistency(%d, %d): ..... ", tc.first, tc.second, ....)

phad:
Drive-by comments from a future user of this API :)

go/client/logclient.go
gdbelvin: I did. That would mean returning the `getConsistencyProofResponse` object, which would also mean exporting it. This seemed cleaner.

go/client/logclient_test.go
gdbelvin: I'm optimizing for fewer characters here so it's not a pain to use in table driven tests.   I think it's pretty clear what it does in context since it's clear that the function is passed base64 strings.

go/client/logclient.go
```diff
@@ -308,3 +309,11 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
...
+	resp := new(getConsistencyProofResponse)
+	u := fmt.Sprintf("%s%s?first=%d&second=%d", c.uri, GetSTHConsistencyPath, first, second)
+	err := fetchAndParse(context.TODO(), c.httpClient, u, &resp)
```
pphaneuf: For a new method, we should take the context as the first parameter.

go/client/logclient.go
```diff
@@ -308,3 +309,11 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
 	sth.TreeHeadSignature = *ds
 	return
 }
+
+// GetSTHConsistency retrieves the consistency proof between two snapshots.
+func (c *LogClient) GetSTHConsistency(first, second uint64) ([][]byte, error) {
+	resp := new(getConsistencyProofResponse)
```
pphaneuf: I'd prefer `var resp getConsistencyProofResponse` (and declaring it at the "last moment"), especially since it's not returned (even if it was returned as a pointer, you could always have done `return &resp, err`, say).

go/client/logclient.go
pphaneuf: The others aren't protocol buffers, aren't they? They're just structures declared in a common package? I think this is okay, since proofs aren't liable to be kept or passed around much...

pphaneuf:
Done with this round of review, should be all right next round.

go/client/logclient.go
```diff
@@ -311,9 +311,11 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
...
 
 // GetSTHConsistency retrieves the consistency proof between two snapshots.
-func (c *LogClient) GetSTHConsistency(first, second uint64) ([][]byte, error) {
-	resp := new(getConsistencyProofResponse)
+func (c *LogClient) GetSTHConsistency(ctx context.Context, first, second uint64) ([][]byte, error) {
 	u := fmt.Sprintf("%s%s?first=%d&second=%d", c.uri, GetSTHConsistencyPath, first, second)
-	err := fetchAndParse(context.TODO(), c.httpClient, u, &resp)
-	return resp.Consistency, err
+	resp := new(getConsistencyProofResponse)
+	if err := fetchAndParse(ctx, c.httpClient, u, resp); err != nil {
```
pphaneuf: Heh, I just caught on with the double pointer, in the previous version. Did that break anything?

I'd still prefer `var resp getConsistencyProofResponse`, but the location of the declaration is good now.

go/client/logclient.go
```diff
@@ -311,9 +311,11 @@ func (c *LogClient) GetSTH() (sth *ct.SignedTreeHead, err error) {
...
+		return [][]byte{}, nil
```
pphaneuf: Huh? Shouldn't you be returning `err` here?

Also, not sure on that one, but can't you return `nil` for a slice? This is the error case, after all, shouldn't be dereferenced...

pphaneuf:
LGTM, will merge once Travis is happy.

Issue #1273, opened 2016-07-19
limit HTTP headers and URL size.


ytrezq:
Sorry but I couldn’t found informations about this *(as the directory layout in undocumented)*

What’s the purpose of the evhtp web server in the project ? What is it used for ?

pphaneuf:
The Certificate Transparency log server protocol is that of an HTTP service (see RFC 6962 for more information), so there *is* an HTTP server, but we still use the one that's included with the `libevent` stable releases.

We do use `evhtp` in the HTTP client:

https://github.com/google/certificate-transparency/blob/master/cpp/net/url_fetcher.cc

We had to use it because of (at least) these problems with the HTTP client included in `libevent`:

https://github.com/libevent/libevent/issues/216
https://github.com/libevent/libevent/issues/217

ytrezq:
@pphaneuf : last question : what is the maximum value of the ᴜʀʟ length for the ʜᴛᴛᴘ server ?

pphaneuf:
From some reading of the current version of the `libevent` code, "very large". Large enough that I'm thinking we should limit it (using `evhttp_set_max_headers_size`).

ytrezq:
@pphaneuf : Please let me do it !

The only question is what should the limit values for ʜᴛᴛᴘ 413 and ʜᴛᴛᴘ 414 return codes.

pphaneuf:
Sure, feel free to send in a pull request! Make sure you have the appropriate CLA signed, as explained in the [CONTRIBUTING.md](https://github.com/google/certificate-transparency/blob/master/CONTRIBUTING.md) file.

The request line should never be too long (there are not many possible parameters), but that setting in `libevent` is shared with the maximum length of each header, so it should still be fairly generous (some cookies and user-agent strings are remarkably long!), maybe something like 4 kilobytes?

For the POST body, 32 kilobytes should be sufficient, without allowing much abuse?

It should be fine to put this in some central location, like `libevent_wrapper.cc`.

ytrezq:
However, as a student, it might take a week or two. Please stay patient.

pphaneuf:
Sure.

ytrezq:
@pphaneuf : why did you not assign to me ?

pphaneuf:
I tried that first, but it wouldn't let me assign it to someone who's not a member of the project. So I took it, that way it doesn't show up as unassigned, and it'll remind me to keep an eye on this in a few weeks, if I haven't heard back. 😉

[...]

PR #1289, opened 2016-08-09
Denial of service fix : limit HTTP requests size.

ytrezq:
fix issue #1273.

As far I can see, it seems the`CHECK_NOTNULL()`macro also stops the process properly isn’t it ?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,6 +14,7 @@ Fiaz Hossain <fiaz.hossain@salesforce.com>
 Google Inc.
 Jeff Trawick <trawick@gmail.com>
 Katriel Cohn-Gordon <katriel.cohn-gordon@cybersecurity.ox.ac.uk>
+Laël Cellier <lael.cellier@gmail.com>
 Mark Schloesser <ms@mwcollect.org>
 NORDUnet A/S
 Nicholas Galbreath <nickg@client9.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -38,6 +38,7 @@ Joe Tsai <joetsai@digital-static.net>
 Kat Joyce <katjoyce@google.com>
 Katriel Cohn-Gordon <katriel.cohn-gordon@cybersecurity.ox.ac.uk>
 Konrad Kraszewski <kraszewski@google.com> <laiquendir@gmail.com>
+Laël Cellier <lael.cellier@gmail.com>
 Linus Nordberg <linus@nordu.net>
 Mark Schloesser <ms@mwcollect.org>
 Nicholas Galbreath <nickg@client9.com>
--- a/cpp/util/libevent_wrapper.cc
+++ b/cpp/util/libevent_wrapper.cc
@@ -272,7 +272,12 @@ event* Base::EventNew(evutil_socket_t& sock, short events,
 
 
 evhttp* Base::HttpNew() const {
-  return CHECK_NOTNULL(evhttp_new(base_.get()));
+  const ev_ssize_t max_http_header_size = 4096;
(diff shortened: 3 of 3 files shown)


ytrezq:
The fact that the osx build failed doesn’t seems to be because of my commits.

AUTHORS
```diff
@@ -22,3 +22,4 @@ Ruslan Kovalov <ruslan.kovalyov@gmail.com>
 Venafi, Inc.
 Vladimir Rutsky <vladimir@rutsky.org>
 Ximin Luo <infinity0@gmx.com>
+Laël Cellier
```
pphaneuf: As per the instructions above, you should have your email address, and keep this list sorted. Also, if I'm deciphering these funny characters on GitHub properly, I think you might be missing a newline?

CONTRIBUTORS
```diff
@@ -50,3 +50,4 @@ Rob Stradling <rob@comodo.com>
 Ruslan Kovalov <ruslan.kovalyov@gmail.com>
 Vladimir Rutsky <vladimir@rutsky.org>
 Ximin Luo <infinity0@gmx.com>
+Laël Cellier
```
pphaneuf: Same comment as for `AUTHORS`.

cpp/util/libevent_wrapper.h
```diff
@@ -20,6 +20,11 @@
 #include "util/executor.h"
 #include "util/task.h"
 
+#define MAX_HTTP_HEADER_SIZE (ev_ssize_t)4096
+#define MAX_HTTP_POST_BODY_SIZE (ev_ssize_t)32768
```
pphaneuf: Should move those to `libevent_wrapper.cc`, inside the anonymous namespace, since they are not needed anywhere else. Also, you should make them C++ variables rather than C preprocessor macros.

Take note of there relevant part of the style guide:

https://google.github.io/styleguide/cppguide.html#Constant_Names

So it could be something like `const ev_ssize_t kMaxHttpHeaderSize = 4096;`, say?

cpp/util/libevent_wrapper.cc
```diff
@@ -272,7 +272,10 @@ event* Base::EventNew(evutil_socket_t& sock, short events,
 
 
 evhttp* Base::HttpNew() const {
-  return CHECK_NOTNULL(evhttp_new(base_.get()));
+  evhttp* vhtp=CHECK_NOTNULL(evhttp_new(base_.get()));
```
pphaneuf: Not sure that `vhtp` is that great of a name? I'm thinking something simple, like `http` (since the method is called `HttpNew`), ought to do it?

Also, mind the use of whitespace: there should be a single space on both sides of a `=`, and a single space after a `,` (below).

I personally prefer "constructor syntax", such as at line 380 of this file.

pphaneuf:
It's been a little while (sorry, my fault!), so it might be nice if you rebased your changes? First, make sure your `master` branch is up to date with ours, then run `git rebase master` while on your `denial_of_service_fix` branch (you might want to do a `git tag savepoint` before if you're not confident about this, just in case!). That will require you to force your next push to GitHub (`-f` flag), this is normal.

AUTHORS
ytrezq: I’m unwilling to provide a valid e‑mail address that can receive span. However, I’d be glad to share my irc or xmpp address.
Sorry for the new line. I’ll add it

cpp/util/libevent_wrapper.h
ytrezq: Sorry, I’ll change this.

ytrezq:
@pphaneuf : done.

pphaneuf:
Thanks!

PR #1293, opened 2016-08-16
Update expected fluentd hash

AlCutter:
Hash and URL from: https://cloud.google.com/logging/docs/agent/installation


pphaneuf:
Looks good enough to me?!? :laughing:

PR #1296, opened 2016-08-17
Python: Ensure repeated fixed_length bytes fields work.

eranmes:
The documentation in tls_options.proto claimed this feature did not work.
Prove that it does by adding a 'repeated bytes' field to the test message
where the length of each element is fixed (using fixed_length).

### DIFF
--- a/python/ct/serialization/tls_message_test.py
+++ b/python/ct/serialization/tls_message_test.py
@@ -26,6 +26,8 @@ valid_test_message.select_uint32 = 2
 valid_test_message.embedded_message.uint_32 = 3
 valid_test_message.repeated_message.add().uint_32 = 4
 valid_test_message.repeated_message.add().uint_32 = 256
+valid_test_message.vector_fixed_bytes.append("\xff\x00\xee\x01")
+valid_test_message.vector_fixed_bytes.append("\xfe\x00\xa0\x03")
 
 
 # Test vectors are given as a list of serialized, hex-encoded components.
@@ -46,6 +48,7 @@ serialized_valid_test_message = [
  "00000002", # 12: select_uint32
  "0003",  # 13: embedded_message.uint_32
  "0400040100",  # 14: repeated_message
+ "08ff00ee01fe00a003", # 15: vector_fixed_bytes
 ]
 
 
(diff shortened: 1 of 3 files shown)


phad:
LGTM although I'm not too familiar with proto annotations or the TLS encoding.  But in offline discussion with Eran it looks good.

python/ct/serialization/tls_message_test.py
```diff
@@ -209,6 +222,12 @@ def test_encode_repeated_too_short_fails(self):
         test_message.ClearField("vector_uint32")
         self.verify_encode_fails(test_message)
 
+    def test_encode_repeated_fixed_length_too_short_fails(self):
```
ekasper: I think you should also test that the encoder rejects components that aren't exactly `fixed_bytes` long.

And also test `test_encode_repeated_fixed_length_too_long_fails`

ekasper:
I have no recollection of what this code does, or why this was a TODO, but if it also passes the suggested added tests, then it's good to go I guess :)

python/ct/serialization/tls_message_test.py
eranmes: Done.

eranmes:
Thanks for the review both - Added tests as requested, PTAL.

python/ct/serialization/tls_message_test.py
```diff
@@ -209,6 +222,25 @@ def test_encode_repeated_too_short_fails(self):
...
+        test_message.ClearField("vector_fixed_bytes")
+        self.verify_encode_fails(test_message)
+
+    def test_encode_repeated_fixed_length_wrong_length_fails(self):
+        test_message = test_message_pb2.TestMessage()
+        test_message.CopyFrom(valid_test_message)
+        test_message.vector_fixed_bytes.append("\xab\x00")
+        self.verify_encode_fails(test_message)
+
+    def test_encode_repeated_fixed_length_too_long_fails(self):
```
phad: using both 'fixed_length' and 'too_short'/'too_long' is a bit confusing.    for the former, fixed_length, you mean the 4 byte length.  For the latter you mean how many repetitions can be permitted.

how about instead naming these:
 test_encode_repeated_fixed_length_too_few_fails 
and
 test_encode_repeated_fixed_length_too_many_fails

eranmes: Done.

eranmes:
Done, PTAL.

python/ct/serialization/tls_message_test.py
phad: I don't see the change?

eranmes: My bad, I haven't force-pushed the commit, it should be in now.

PR #1299, opened 2016-08-24
Remove the unused objecthash bits from C++ code.

pphaneuf:
When you want to use objecthash, use it, but until then... :stuck_out_tongue_closed_eyes:


eranmes:
LGTM - thanks for removing dependencies.

eranmes:
(do we still have to build the objecthash library?)

pphaneuf:
If you're using gbuild, yes. I'm not, and the `configure` script was disabling the XJSON part of the code, for (currently) no good reason.

PR #1304, opened 2016-09-05
Remove unused fields from OpenSSL log list output

RJPercival:
Simplify the code by not outputting fields that OpenSSL doesn't currently use.


pphaneuf:
LGTM

PR #1305, opened 2016-09-07
Add configs for a startcom mirror

AlCutter:


cloud/google/configs/ct-mirror-startcom.sh
```diff
@@ -0,0 +1,9 @@
+export INSTANCE_TYPE="mirror"
+export PROJECT="ct-mirror-startcom"
```
RJPercival: Following the advice for naming our DNS frontends, I've been calling this log "startcom1".

AlCutter: oke doke - changed

PR #1308, opened 2016-09-14
Crash if LevelDB node ID lookup fails in an unexpected way

RJPercival:
This can happen if the database is corrupt. Crashing here will result in a more helpful error message than the current error message, which is "leveldb_db.cc:336] Attempting to initialize DB beloging to node
with node_id:".


Martin2112:
OK I didn't realize you were already doing this. I wrote a PR to log the status, which we probably don't need now.

PR #1310, opened 2016-09-20
Have CertChecker::CheckIssuerChain log better.

pphaneuf:
If we have a non-OK status and we're logging a message, it should have the status in it!

### DIFF
--- a/cpp/log/cert_checker.cc
+++ b/cpp/log/cert_checker.cc
@@ -152,7 +152,7 @@ Status CertChecker::CheckIssuerChain(CertChain* chain) const {
   // because we will later check that it is trusted.
   Status status = chain->IsValidCaIssuerChainMaybeLegacyRoot();
   if (!status.ok()) {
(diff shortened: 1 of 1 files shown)


cpp/log/cert_checker.cc
```diff
@@ -152,7 +152,7 @@ Status CertChecker::CheckIssuerChain(CertChain* chain) const {
   // because we will later check that it is trusted.
   Status status = chain->IsValidCaIssuerChainMaybeLegacyRoot();
   if (!status.ok()) {
-    LOG(ERROR) << "Failed to check issuer chain";
+    LOG(ERROR) << "Failed to check issuer chain: " << status;
```
eranmes: I assume it prints the string as well?

pphaneuf: The string? `status` has an `operator<<` that will invoke `ToString()` on it, which looks like this:

https://github.com/google/certificate-transparency/blob/master/cpp/util/status.cc#L70

PR #1311, opened 2016-09-22
Log status when we fail to init leveldb

Martin2112:
Mirrors are crashing at this point, collecting more info.


Martin2112:
Obsoleted by #1308.

PR #1312, opened 2016-10-11
Add a script to update etcd VM images

AlCutter:


cloud/google/update_etcd_vm_image.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+# TODO(alcutter): Factor out common code with update_mirror_vm_image.sh script
```
Martin2112: Not sure if thats > 80 chars or just how GH is displaying it.

Martin2112: This script just recreates what's there? Might be worth a comment on how safe it is to run.

AlCutter: 79 chars I think

AlCutter: I've added a bunch more comments about what the script does.

PR #1318, opened 2016-10-18
Use LoggedEntryPB::CopyFrom directly.

ckennelly:


pphaneuf:
The change itself looks all right, could you add yourself to the `CONTRIBUTORS`, as per the comment at the top?

ckennelly:
I'm already in `CONTRIBUTORS`: https://github.com/google/certificate-transparency/blob/master/CONTRIBUTORS#L28

pphaneuf:
Oh, my bad! I need to put on my glasses, apparently...

It seems the Mac build is having problems, but I think this is good, so I'll be merging it. Thanks!

PR #1323, opened 2016-10-27
Update contributors file with Kiril Nikolov

dejandb:


dejandb:
The CLA was finalized in DocuSign after the last message so this is just to restart the check.

PR #1334, opened 2016-11-08
Golang: Fix go/jsonclient test for Go 1.8

dsnet:
It is considered bad practice matching on error strings since the exact
string output is not generally covered by the Go compatibility agreement.
In order for this code to work on both Go 1.7 and Go 1.8, we use a regexp
that matches on both versions of the output.


dsnet:
\cc @daviddrysdale

PR #1342, opened 2016-11-21
Do not treat loading an empty certificate file as an error

RJPercival:
This will cause `CertChecker::LoadTrustedCertificatesFromBIO` to return true when loading an empty file. It previously returned false, indicating an error.

### DIFF
--- a/cpp/log/cert_checker.cc
+++ b/cpp/log/cert_checker.cc
@@ -75,34 +75,33 @@ bool CertChecker::LoadTrustedCertificatesFromBIO(BIO* bio_in) {
 
   while (!error) {
     ScopedX509 x509(PEM_read_bio_X509(bio_in, nullptr, nullptr, nullptr));
-    if (x509) {
-      // TODO(ekasper): check that the issuing CA cert is temporally valid
-      // and at least warn if it isn't.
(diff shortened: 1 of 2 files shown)


cpp/log/cert_checker.cc
```diff
@@ -71,47 +71,36 @@ bool CertChecker::LoadTrustedCertificates(
...
         ClearOpenSSLErrors();
         break;
-      } else {
-        // A real error.
-        LOG(ERROR) << "Badly encoded certificate file.";
-        LOG_OPENSSL_ERRORS(WARNING);
-        error = true;
-        break;
       }
+      // A real error.
```
phad: So if I've followed the restructure right, the only functional change is checking for BIO_eof which previously would have fallen into this 'real error' block.  In that case, looks fine.

phad: Ah, according to Travis, boringssl doesn't like the BIO_eof you've used.
Can you fix that or find a workaround?

RJPercival: Since we don't actually need this change, I'm not going to spend the time figuring out a BoringSSL workaround. Cancelling PR...

PR #1344, opened 2016-11-23
Add synchronize_signing as start option for ct-server

knikolov82:


eranmes:
Thanks for your pull request.
I've got a few questions before we start reviewing it:
(1) What's the motivation behind it? Please add it to the description of the commit.
(2) You seem to be using a global lock - can it be a field of the signer?
(3) You seem to protect only the EVP_SignFinal call with the mutex. Why is that? It's hard to figure out if it's correct without context for the pull request (and given the use of the ctx object, why it's at all necessary).

cpp/log/signer.h
```diff
@@ -34,6 +34,7 @@ class Signer {
   ct::DigitallySigned::HashAlgorithm hash_algo_;
   ct::DigitallySigned::SignatureAlgorithm sig_algo_;
   std::string key_id_;
+  bool synchronize_signing_;
```
pphaneuf: Make this `const`.

cpp/log/signer.cc
```diff
@@ -59,7 +63,12 @@ std::string Signer::RawSign(const std::string& data) const {
   unsigned int sig_size = EVP_PKEY_size(pkey_.get());
   unsigned char* sig = new unsigned char[sig_size];
 
-  CHECK_EQ(1, EVP_SignFinal(&ctx, sig, &sig_size, pkey_.get()));
+  if (synchronize_signing_) {
+    std::lock_guard<std::mutex> lock(lock_);
+    CHECK_EQ(1, EVP_SignFinal(&ctx, sig, &sig_size, pkey_.get()));
```
pphaneuf: Is this because some OpenSSL engines are not thread-safe? Is it sufficient to only protect the call to `EVP_SignFinal`?

cpp/log/signer.cc
```diff
@@ -18,7 +19,9 @@ using cert_trans::Verifier;
 
 namespace cert_trans {
 
-Signer::Signer(EVP_PKEY* pkey) : pkey_(CHECK_NOTNULL(pkey)) {
+std::mutex lock_;
```
pphaneuf: This really needs to be a global singleton?!? In any case, it should be in an anonymous namespace, and the name changed to comply with the style guide (no trailing underscore, that's for member variables): https://google.github.io/styleguide/cppguide.html#Variable_Names

I'd change the name to be a bit more descriptive than just "lock", maybe "signer_lock"?

dejandb:
@eranmes I briefly mentioned the reasoning behind the option to @pphaneuf a few weeks ago. The SafeNet CloudHSMs that we use in AWS have threading issues and under load this causes the library to start misbehaving. There is no way to control locking externally other than to synchronize the calls to the library and hence the lock in the patch.

This behavior is probably not unique to SafeNet and the flag could benefit others too.

eranmes:
@dejandb thanks for the context. Would be very helpful if the context is in the commit description.

But wouldn't all users of OpenSSL suffer? Shouldn't the OpenSSL engine handle that?

dejandb:
@eranmes - This is an issue with the SafeNet PKCS11 library implementation, not OpenSSL. I agree that it can probably be fixed in OpenSSL too, however here looked easier and less moving parts.

eranmes:
@dejandb just to make sure I correctly understand the situation:
There's a problem running the open-source CT log with SafeNet's CloudHSM because:
* The SafeNet library PKCS11 library for interacting with the HSM has threading issues.
* The open-source code doesn't call it directly, but through its integration with OpenSSL.
* To work around the threading issues in SafeNet's PKCS11 library, the proposed change is to have a global mutex in the CT code, working around it two layers above where the problem is.

Is that right? Could you please help me understand if the library is provided by the vendor to AWS and you're forced to use it or if it's something you obtain and install yourself? I would hope that you could ask SafeNet to fix their library.

cpp/log/signer.cc
eranmes: I'm wondering if it's worth sub-classing the Signer class to have a SingleThreadedSigner that will lock before calling EVP_SignFinal (so only the call to EVP_SignFinal would be delegated).

That way the right signer would be created from configuration during initialization, it will not change, and there won't be two flows to check/fix should any bugs be discovered here.

pphaneuf: I disagree with this, if `synchronize_signing_` is `const` (as I suggested earlier), there is effectively no difference, and this will only add complexity (there will *still* be two flows to check or fix if a bug is discovered!).

I think there should be some good comment here explaining why this is done, how this is not necessary in theory but is a workaround for the HSM threading issue, etc...

pphaneuf:
I think we should accept this (once other comments above are addressed, of course!), but we would need the flag documentation string to include some explanation on why/when this flag is necessary, not have the explanation just in the pull request or the commit!

dejandb:
@eranmes - The problem is both in PKCS11 and SafeNet - PKCS11 doesn't define well a threading model and SafeNet doesn't do anything additional to fix potential MT problems.

We can ask SafeNet to fix their library but this is a very long process with potentially uncertain result.

The SafeNet library is provided by SafeNet and is the only one that they would support for using with the AWS HSMs.

eranmes:
@dejandb please ask SafeNet to fix their library and, if possible at all, include a link to the open issue in their bug tracking system / customer issues / product forum page.

This work-around should be removed when the SafeNet library is fixed.

homepc91:
Hi, Everyone's please please help me to solve the problem

My 1st Problem is :- When I run ./prepare_etcd.sh, I am getting error in last line :-> "W1013 07:02:03.298388 528384 etcd.cc:798] Got invalid JSON:**"

My 2nd Problem is :- How to get Base64 Log ID..?

32977192-652c676c-cc4e-11e7-8477-aa195f51f1b6
Please Please help me

thanks

PR #1347, opened 2016-12-01
Use Python's unittest ability to detect tests

RJPercival:


eranmes:
LGTM if that also handles the twisted tests .

RJPercival:
Seems to be causing 3 of the 4 Travis build bots to stall (it seems "make -C python test" is not terminating on those bots, for no obvious reason). Works fine on 1 of the bots though, strangely.

PR #1348, opened 2016-12-01
Remove flag parsing from tests

RJPercival:
This stops flags being passed to the test runner, because gflags raises an exception when it encounters an unrecognised flag. Replaced flags with constants.

### DIFF
--- a/python/ct/client/monitor_test.py
+++ b/python/ct/client/monitor_test.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env trial
 import copy
 import difflib
-import gflags
 import logging
 import mock
 import os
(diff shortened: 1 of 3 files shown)


python/ct/client/monitor_test.py
```diff
@@ -20,10 +19,8 @@
 from twisted.web import iweb
 from zope.interface import implements
 
-FLAGS = gflags.FLAGS
-
-#TODO(ekasper) to make this setup common to all tests
-gflags.DEFINE_bool("verbose_tests", False, "Print test logs")
+# Print test logs
+VERBOSE_TESTS=False
```
eranmes: Simply remove it? this constant is used in exactly one place.

RJPercival: Done.

RJPercival:
It turns out we use --testdata_dir, so closing this PR in favour of #1350.

PR #1349, opened 2016-12-01
Java: Implement verification of SCTs for logs with RSA keys

samuellb:
This is needed to be able to submit certificates/precerts to logs that use RSA keys.


eranmes:
Looks good - can you please add a test demonstrating verification of an SCT from a log with RSA signatures?

samuellb:
Good point. I added two tests of SCTs (pre-cert and normal cert) and a test of the log ID calculation.

java/src/org/certificatetransparency/ctlog/serialization/CryptoDataLoader.java
```diff
@@ -83,21 +83,28 @@ static PublicKey parsePublicKey(List<String> pemLines) {
...
-      kf = KeyFactory.getInstance("EC");
-      return kf.generatePublic(spec);
-    } catch (NoSuchAlgorithmException e) {
-      // EC is known to be missing from openjdk; Oracle's JDK must be used.
-      throw new UnsupportedCryptoPrimitiveException("EC support missing", e);
-    } catch (InvalidKeySpecException e) {
-      throw new InvalidInputException("Log public key is invalid", e);
+        // First try to parse as an RSA key. If it's an EC key then generatePublic throws.
+        // On OpenJDK it throws NPE, but a different exception might be thrown on other Java implementations
+        kf = KeyFactory.getInstance("RSA");
```
eranmes: Could you use EncodedKeySpec.getFormat to figure out which key has to be decoded?
https://docs.oracle.com/javase/7/docs/api/java/security/spec/EncodedKeySpec.html#getFormat()

java/src/org/certificatetransparency/ctlog/serialization/CryptoDataLoader.java
```diff
@@ -83,21 +83,28 @@ static PublicKey parsePublicKey(List<String> pemLines) {
...
+        return kf.generatePublic(spec);
+    } catch (Exception e1) {
```
eranmes: (1) This catch is too broad.
(2) Seems like the code is re-trying with EC even if a fatal exception was thrown from generatePublic.

It seems like we could have a better flow here:
* Figure out which key specification to use.
* try to generatePublic according to that key.
* Fail according to the exception rather than re-try with the other key type.

What do you think?

java/test/org/certificatetransparency/ctlog/TestData.java
```diff
@@ -15,14 +15,17 @@
   private static final String DATA_ROOT = "test/testdata/";
   // Public log key
   public static final String TEST_LOG_KEY = DATA_ROOT + "ct-server-key-public.pem";
+  public static final String TEST_LOG_KEY_RSA = DATA_ROOT + "ct-server-key-public-rsa.pem";
```
eranmes: Could you put all the new RSA-related test data under DATA_ROOT/rsa ?
Otherwise it might confuse other developers looking at the test/testdata directory to see two keys (because everything under test/testdata was generated/signed by the same log key).

java/src/org/certificatetransparency/ctlog/serialization/CryptoDataLoader.java
samuellb: Will have a look at it tomorrow. I looked quickly at the OpenJDK source now, and it looks like it always returns the string "X.509". So not sure if it will work, but if not I'll try to find some alternative

java/test/org/certificatetransparency/ctlog/TestData.java
samuellb: Good point. Will change this and the catch Exception tomorrow

java/src/org/certificatetransparency/ctlog/serialization/CryptoDataLoader.java
samuellb: I couldn't find any method to check the key algorithm, so I replaced the "catch Exception" with a check of the OID.

java/test/org/certificatetransparency/ctlog/TestData.java
samuellb: Done

eranmes:
Looks good, thanks for taking the time to add tests and test data!
Please squash before merging.

samuellb:
Sorry but I don't have write access to the repo, so I can't merge it

samuellb:
Thanks for reviewing and merging!

PR #1355, opened 2016-12-07
Python: Export the TBSCertificate

eranmes:
Add a method to export the TBSCertificate part of the certificate.

In RFC6962-bis, the signature in the SCT for X.509 certs does not cover
the entire certificate, but the TBSCertificate and the issuer key hash.

To implement issuance of V2 SCTs, access to the TBSCertificate is necessary.

### DIFF
--- a/python/ct/crypto/cert.py
+++ b/python/ct/crypto/cert.py
@@ -221,6 +221,14 @@ class Certificate(object):
         """
         return self._asn1_cert["tbsCertificate"]["extensions"] or []
 
+    def tbscertificate(self):
+        """Returns the underlying tbsCertificate
+
+        Returns:
+            An x509.TBSCertificate instance.
+        """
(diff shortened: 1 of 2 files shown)


python/ct/crypto/cert.py
```diff
@@ -221,6 +221,14 @@ def get_extensions(self):
         """
         return self._asn1_cert["tbsCertificate"]["extensions"] or []
 
+    def get_tbscertificate(self):
```
RJPercival: Python getters don't typically start with "get_" (the `get_extensions` method above is an anomaly, even within this class).

RJPercival: Test?

eranmes: Done and done.

eranmes:
@RJPercival PTAL

PR #1356, opened 2016-12-07
Python: Add ECDSA signing code.

eranmes:
Add a module for producing ECDSA signatures that are suitable for use in
CT log implementations (both 6962 and 6962-bis):

* Produce deterministic signatures.
* Signature bytes are DER-encoded.
* Signature is wrapped in a DigitallySigned protobuf for easy TLS
  serialization.

This is necessary for a reference implementation of a 6962-bis log.

### DIFF
--- a/python/ct/crypto/signing/signer_ecdsa.py
+++ b/python/ct/crypto/signing/signer_ecdsa.py
@@ -1,0 +1,37 @@
+"""ECDSA signatures suitable for CT logs"""
+from ct.proto import client_pb2
+import hashlib
+import ecdsa
+
+class EcdsaSigner(object):
+    """Signs using deterministic ECDSA signatures."""
+
+    def __init__(self, key_pem):
+        """Creates a signer from a PEM-encoded ECDSA private key.
+
+        Args:
+            - key_pem: Private key, in PEM format.
+        """
+        self.__key = ecdsa.SigningKey.from_pem(key_pem)
+
+    def sign_raw(self, data_to_sign):
+        """Returns the raw signature bytes over |data_to_sign|.
+
+        The signature is deterministic and output is DER encoded.
+        """
+        return self.__key.sign_deterministic(
+                data_to_sign, hashfunc=hashlib.sha256,
+                sigencode=ecdsa.util.sigencode_der)
+
+    def sign(self, data_to_sign):
+        """Returns a signature over |data_to_sign|.
+
+        The returned value is a DigitallySigned protobuf instance with the
+        right signature and hash algorithms set, in addition to the signature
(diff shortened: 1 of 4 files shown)


RJPercival:
I think we should merge PR #953 first, as that ecdsa module has issues and we'd be better off not expanding its use.

eranmes:
To follow up:
* ECDSA signatures produced by the cryptography.io package are not deterministic.
* Cryptography.io uses openssl underneath so OpenSSL has to support it first:
https://mail.python.org/pipermail/cryptography-dev/2016-December/000698.html
* I've requested this feature in OpenSSL, but even once implemented, will take a while to make it to a stable release.

So I think we should stay with the ecdsa package here for now.

RJPercival:
That's a shame. I've filed an issue with OpenSSL to track this (https://github.com/openssl/openssl/issues/2078). I guess the ecdsa package is the best option we have.

python/ct/crypto/signing/signer_ecdsa.py
```diff
@@ -0,0 +1,37 @@
+"""ECDSA signatures suitable for CT logs"""
+from ct.proto import client_pb2
+import hashlib
+import ecdsa
+
+class EcdsaSigner(object):
+    """Signs using ECDSA signatures."""
+
+    def __init__(self, key_pem):
+        """Creates a signer from the a PEM-encoded ECDSA public key.
```
RJPercival: typo: "from the a" -> "from a"

python/ct/crypto/signing/signer_ecdsa.py
```diff
@@ -0,0 +1,37 @@
+"""ECDSA signatures suitable for CT logs"""
+from ct.proto import client_pb2
+import hashlib
+import ecdsa
+
+class EcdsaSigner(object):
+    """Signs using ECDSA signatures."""
```
RJPercival: Signs using **deterministic** ECDSA signatures.

python/ct/crypto/signing/signer_ecdsa.py
RJPercival: typo: "public key" -> "private key"

python/ct/crypto/signing/signer_ecdsa.py
eranmes: Done.

python/ct/crypto/signing/signer_ecdsa.py
eranmes: Done & Done.

eranmes:
@RJPercival PTAL.

PR #1362, opened 2017-02-09
Remove cgo merkle tree

gdbelvin:
I couldn't find any code that depends on the C++ version of the merkle tree through Go. 

Removing this removes the dependency on OpenSSL for the Go libraries, making the Go repo `go get`'able.


gdbelvin:
Checked that this is not used internally. Merging.

PR #1363, opened 2017-02-10
Move cgo merkle tree to alternative package.

gdbelvin:
This allows 3rd parties to use merkle tree verification algorithms
without aquiring a dependency on openssl.


AlCutter:
Would the `fulltree` package be better moved under the `merkletree/` package?

gdbelvin:
I found it easier to compile at the root level. I can try again under the merkletree package if you prefer.

AlCutter:
Yeah, it's definitely related to merkletree so I think it's the right place for it. I can give you a hand on Monday wrangling the compile if you'd like.

PR #1364, opened 2017-02-11
Fix build issues with LibreSSL

Sp1l:
1. Fix CMS detection
2. Fix OPENSSL_VERSION_NUMBER checks


eranmes:
Pierre has kindly agreed to take a look.

digitalist:
how are we doing here?

PR #1375, opened 2017-03-06
Use maven to build java code

alexanderkjall:
Some remarks:

The .idea directory is deleted and added to .gitignore, please backup your private copy of it before you pull this in order to not loose any settings.

The /test/testdata directory is symlinked in java/src/test/resources. This works fine on my linux machine, but might cause trouble with windows maybe? I don't really have any experience with windows.

I have signed the CLA.

Should I maybe collapse the commits?

feedback is welcome :)


pphaneuf:
If there's going to be some work on the Java code (which is quite welcome!), could I ask for a temporary "freeze" after this PR is merged, to do some repository reorganisation? It would likely break PRs that are "in flight" for this area, but I think it would make things easier for you after... :smiley:

java/pom.xml
```diff
@@ -0,0 +1,273 @@
...
+            </plugin>
+
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.6.0</version>
+                <inherited>true</inherited>
+                <configuration>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                    <encoding>UTF-8</encoding>
```
codingllama: Use `${project.build.sourceEncoding}`?

java/src/test/java/org/certificatetransparency/ctlog/LogSignatureVerifierTest.java
```diff
@@ -56,13 +32,13 @@
 public class LogSignatureVerifierTest {
   /** Returns a LogSignatureVerifier for the test log with an EC key */
   private LogSignatureVerifier getVerifier() {
-    LogInfo logInfo = LogInfo.fromKeyFile(TEST_LOG_KEY);
+    LogInfo logInfo = LogInfo.fromKeyFile(LogSignatureVerifierTest.class.getResource(TEST_LOG_KEY).getFile());
```
codingllama: nit: Do `getClass().getResource(...)` instead of using the class name? The method is not static and `getClass()` tends to induce less copy-paste errors.

Same for others.

nit2: Please break lines at 100 columns. Eg:

```java
LogInfo logInfo =
    LogInfo.fromKeyFile(LogSignatureVerifierTest.class.getResource(TEST_LOG_KEY).getFile());
```

Google's line breaking rules are pretty particular, so your IDE autoformat is unlikely to be helpful here. (The rule is [break at a higher syntactic level](https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break)).

In this particular case it'll go under 100 chars if you do `getClass()`, but please check the others as well.

java/src/test/resources/testdata
```diff
@@ -0,0 +1 @@
+../../../../test/testdata/
```
codingllama: Oh, I see what you mean by by the symlink. I don't think that will work well for Windows.

An option is to do the linking using Ant (via antrun-plugin). Eg:

```xml
<execution>
    <id>symlink-testdata</id>
    <phase>generate-test-resources</phase>
    <configuration>
        <target>
            <mkdir dir="${project.basedir}/src/test/resources"/>
            <symlink link="${project.basedir}/src/test/resources/testdata"
                resource="${project.basedir}/../test/testdata"
                overwrite="true"/>
        </target>
    </configuration>
    <goals>
        <goal>run</goal>
    </goals>
</execution>
```

Another option might be to simply copy the files to `${project.build.testOutputDirectory}/testdata`. I don't think I have a strong preference; none of it makes me too happy TBH.

java/src/test/java/org/certificatetransparency/ctlog/LogSignatureVerifierTest.java
```diff
@@ -47,6 +20,9 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import static org.certificatetransparency.ctlog.TestData.*;
```
codingllama: Please don't change the order of the imports, nor use wildcard imports.

https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

java/src/test/java/org/certificatetransparency/ctlog/comm/HttpLogClientTest.java
```diff
@@ -39,15 +17,30 @@
 import org.junit.runners.JUnit4;
 import org.mockito.Matchers;
 
-import com.google.protobuf.ByteString;
+import java.io.ByteArrayInputStream;
```
codingllama: Ditto for imports.

codingllama:
OK, I had a look at the Travis failure and the problem is that [ant build test](https://github.com/google/certificate-transparency/blob/master/.travis.yml#L128) is failing, as [build.xml](https://github.com/google/certificate-transparency/blob/master/build.xml) is now outdated.

My suggestion is to kill it with fire (along with `build.properties`) and replace ant with a `cd java && mvn test install` at `.travis.yml` (plus adding the Maven dependency to it, doing a bit of a cleanup, etc).

@pphaneuf , any objections?

pphaneuf:
If you're pulling the carpet out from under Ant, and replacing it with Maven, then it would make sense to fix up the `.travis.yml` to run the appropriate Maven commands there.

As long as the tests are being run (and their failure cause Travis to fail as well, which hasn't always been true!), I'm happy. :wink:

java/pom.xml
```diff
@@ -0,0 +1,273 @@
...
+
+
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>1.5.0</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>exec</goal>
```
alexanderkjall: I didn't really understand this one. My thinking was that it's both useful to be able to do mvn exec:java and to produce an jar with a correct manifest so that it's executable for non-developers?

java/pom.xml
```diff
@@ -0,0 +1,273 @@
...
+                    <forkCount>1</forkCount>
+                    <reuseForks>true</reuseForks>
+                    <includes>
+                        <include>**/*.class</include>
+                    </includes>
+                </configuration>
+            </plugin>
+
+
+            <plugin>
```
alexanderkjall: Copy-paste is dangerous :)

java/src/test/java/org/certificatetransparency/ctlog/LogSignatureVerifierTest.java
alexanderkjall: Automatically done by intellij, i think i have gotten it to conform with the google style now.

java/src/test/java/org/certificatetransparency/ctlog/LogSignatureVerifierTest.java
alexanderkjall: I improved the line breaks, some went a bit over 100, but my thought was that it was better that the code pattern repeated between the tests than all lines was below 100 chars.

I don't have any particular opinion about this, just notify me if it should be changed more.

java/src/test/resources/testdata
alexanderkjall: I agree on this not making me happy, none of the three different solutions are very clean.

I went for adding an antrun execution, as the project already used that for protobuf.

alexanderkjall:
I don't have any experience with travis, so i haven't made any changes to the .travis.yml file as I didn't have any easy way to verify that it work.

Is there some way to run the travis file, so that I don't have to do an commit in order to do a testrun?

alexanderkjall:
The travis build system seem to have defeated me.

This is my theory of the problem: The apt addon doesn't update it's index before it tries to download stuff, therefor it doesn't find the java-8 packages and the setting of JAVA_HOME points to an invalid path.

If i try to run "sudo apt-get update" in the before_install section, the build fails on OS X.

And i can't find a way to run apt-get update in the apt section, but maybe my googling skills are too weak.

pphaneuf:
I was hoping this wouldn't happen and it would all Just Work, but I had a feeling it could... Hence my clever wording that "all I needed to be happy" was for Travis to run all the tests properly. :wink: 

I'm sure it's possible, but that stuff always needs some fiddling around... Good luck!

codingllama:
Sorry about Travis, it can be a bit painful sometimes. No promises, but I'll see if I can have a go at it later today.

alexanderkjall:
I have thought about the problem a bit since yesterday. And I think it boils down to that the maven package that gets installed by the apt section uses the wrong java version.

The next things I will try: 
1) There might be some other way to install maven
2) Maybe it's already installed.
3) Maybe there is some way to conditionally set JAVA_HOME

codingllama:
I had a JAVA_HOME mismatch when I was playing around with the Docker container. I think the box comes with java7 and sets JAVA_HOME to it, which maven respects.

Could you add a `java -version` and a `mvn -version` at the `before_install` section?

You could unset JAVA_HOME and see if it helps. This seemed to work in the Docker container:

```bash
travis@91be326c05ab:~$ java -version
java version "1.8.0_121"
...
travis@91be326c05ab:~$ mvn -version
Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-14T17:29:23+00:00)
Maven home: /usr/local/maven
Java version: 1.7.0_76, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-7-oracle/jre
...
travis@91be326c05ab:~$ unset JAVA_HOME
travis@91be326c05ab:~$ mvn -version
Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-14T17:29:23+00:00)
Maven home: /usr/local/maven
Java version: 1.8.0_121, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-oracle/jre
...
```

If the above doesn't work, you can also try to set JAVA_HOME with this monster I put together here. It should be relatively portable, but I haven't tested under OSX.

```bash
export JAVA_HOME="$(java -XshowSettings:properties -version 2>&1 | grep 'java.home' | awk -F' = ' '{print $2}')"
```

codingllama:
On the plus side, we know the build fails when maven goes wrong, so there's that. :)

pphaneuf:
That was a specific problem in the past, where I *knew* I had broken something, and it turned out the test command still returned a zero status? See commit 0675b70f6b4355de5650dd5e8c5cf8d25a06604e... :stuck_out_tongue:

eranmes:
Out of curiosity, why the proposal to remove the .idea directory?
What's the alternative?

alexanderkjall:
My motivation for removing the .idea is that the pom.xml file should be the canonical way that the project gets built.

It's easy to configure intellij to use the pom.xml file when starting the project, file -> new -> Project from existing sources -> select the pom.xml file.

And also since not all user-specific settings where included in the .gitignore file I felt that it's about as much work removing it as adding more stuff to it.

codingllama:
@eranmes I stand behind removing IDE-specific from the repos for a few reasons:

1. It's not uncommon for IDE settings to have assumptions about the machine they were created or user specific settings (local paths, references to your favorite code style, so on).
1. Maven becomes the canonical source of configuration, so there's no room for mismatch.
1. Maven is well supported by Java IDEs (Eclipse, IntelliJ). As Alexander said, you can just open it in IntelliJ.
1. IDE-specific settings are poor form and personal. What if someone uses Eclipse? Or vim? Or notepad?

Maven is standard in the Java community. We should make the move to it.

codingllama:
I've been following your attempts at Travis, but it doesn't seen to be working for linux (java -version is 1.7 regardless of the jdk setup). I think it boils to the top-level language on the travis.yml file, but I don't want to mess with that.

We could set the project to use Java 1.7 - this way we don't need special JDKs or JAVA_HOME magic.

We can fight Travis for Java8 support on another PR, then.

pphaneuf:
Could the code changes be merged in while keeping Ant for now? Because I might be able to help with this, at the cost of having to re-do this PR, and I'm thinking that the build system part of it (which is what's not working here without the correct Java version, I think?) might not be bad to re-do, so if the code changes could be salvaged (changing this PR from "use Maven" to "reorganise the Java code a bit"), it might be the easiest way?

alexanderkjall:
The only code changes I made was to replace the access to the test files to be done in a more maven friendly way, and I don't think those changes will work with ant.

But I don't think there is anything in the codebase that requires java 8, I simply defaulted to that since java 7 is EOL.

I'll reduce the language level to 1.7 and make a try with that right after work.

pphaneuf:
OK, makes sense. Give that a try, we'll see how it goes. Even if I do my own reorganisation, it should be possible to save this PR as a diff, and re-apply it cleanly enough after I've done my thing, but if you can get this PR to pass the tests and merge it, that'd be the best. :smiley:

alexanderkjall:
Seems like i got the tests running by hard coding the java 8 path for linux into the travis file.

This feels a bit like building a house with rubber bands and glue, but maybe it's good enough?

.travis.yml
```diff
@@ -125,7 +128,9 @@ install:
 script:
  - getconf _NPROCESSORS_ONLN
  - ${MAKE} -j$(getconf _NPROCESSORS_ONLN) check VERBOSE=1 V=${ENV_VERBOSE}
- - ant build test
+ - if [ -e "/usr/lib/jvm/java-8-oracle/" ]; then export JAVA_HOME=/usr/lib/jvm/java-8-oracle/; fi
```
codingllama: Could you move this to the before_install section, before the `java -version` line, and add an `echo "JAVA_HOME=$JAVA_HOME"` after it?

.travis.yml
```diff
@@ -125,7 +128,9 @@ install:
...
+ - mvn -version
```
codingllama: nit: Remove this line. The `mvn -version` at before_install should be the one that counts.

codingllama:
Btw, I think we're really close to the end here.

alexanderkjall:
So... that just looks like a bug somewhere in travis? I'll just make a whitespace commit to trigger a new build and then we can see if it corrects itself. Unless you have any other suggestion.

codingllama:
I didn't see what happened, but Travis is not 100% free of flakes, so there's that.

Alexander, many thanks for the PR! It's a great improvement to our Java client and I'm happy to see it get through.

alexanderkjall:
At last! Thanks for all the work guiding this pull request to completion, this is one of the most professional and pleasant open source experiences I have had.

Regarding travis, I checked their incident log on their homepage, and they had an incident with a partition running out of space for the mac cluster yesterday.

pphaneuf:
FYI, there is a "retry build" button somewhere in Travis, but I think you might have to be an owner of the repository (so you wouldn't have been able to trigger it).

PR #1381, opened 2017-03-29
Add logging to prober.py to identify problematic log configs

RJPercival:
If an exception is raised while starting a prober, it may be useful to know which CT log config it was processing at the time.

### DIFF
--- a/python/ct/client/prober.py
+++ b/python/ct/client/prober.py
@@ -35,22 +35,26 @@ class ProberThread(threading.Thread):
                 raise RuntimeError("Cannot start monitor: log proto has "
                                    "missing or empty fields: %s" % log)
 
-            temp_db = temp_db_factory.create_storage(log.log_server)
-            client = async_log_client.AsyncLogClient(agent,
-                                               log.log_server,
-                                               temp_db)
-            hasher = merkle.TreeHasher()
-            verifier = verify.LogVerifier(log.public_key_info,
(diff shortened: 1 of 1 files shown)


python/ct/client/prober.py
```diff
@@ -35,22 +35,26 @@ def __init__(self, ct_logs, db, cert_db, temp_db_factory, monitor_state_dir,
...
+                # log ID can be used as part of a file path.
+                log_id_urlsafe = log.log_id.replace('/', '_').replace('+', '-')
+                state_keeper = state_keeper_class(monitor_state_dir +
+                                                 "/" + log_id_urlsafe)
+                log_key = db.get_log_id(log.log_server)
+                self.__monitors.append(monitor.Monitor(client, verifier, hasher,
+                                                       db, cert_db, log_key,
+                                                       state_keeper))
+            except:
+                logging.error("Error starting monitor for log: %s" % log)
```
phad: Can you output anything about the exception itself?

Also, you might just output log.log_id (upto you; outputting the whole proto might make any fault obvious straightaway)

RJPercival: Since I re-raise the exception, it should be the next thing to appear in the logs. I wanted to log the entire proto since, as you say, it may make the issue obvious.

phad: OK, sounds fine.

PR #1383, opened 2017-03-29
Add mirror config for PuChuangSida CT log

RJPercival:
It was recently accepted by Chrome (see https://crbug.com/667663).


phad:
LGTM

PR #1391, opened 2017-05-10
Configure google-fluentd correctly for ct-mirror

RJPercival:
Previously, collection of logs was only correctly configured for ct-server.


RJPercival:
There may well be a better way to determine what kind of node `node_init.sh` is running on than inspecting the Docker images. Any suggestions? The GCE documentation suggests setting custom metadata on the VM instance when it's created and reading that back in `node_init.sh`, so I could do that if you'd prefer it.

PR #1395, opened 2017-05-15
Delete the Cert default constructor with the others.

pphaneuf:


cpp/log/cert.h
```diff
@@ -30,6 +30,7 @@ class Cert {
   // Will return null if |x509| is null.
   static std::unique_ptr<Cert> FromX509(ScopedX509 x509);
 
+  Cert() = delete;
```
RJPercival: While redundant (the default constructor is suppressed by the constructor on line 201), I suppose this makes it easier to see at a glance that there is no default constructor.

pphaneuf: Yeah, I considered just deleting it, but I think it's nice that there's no ambiguity about not being able to construct one (the other constructor is more of an implementation detail).

PR #1401, opened 2017-06-02
Change machine type for aviator,pilot,rocketeer mirrors

phad:
n1-highmem-4 has more memory and CPU than n1-highmem-2.


Martin2112:
Do we need the cpu as well as more RAM? We could have a custom machine type if not but it's probably more work. Let's go with this for now.

phad:
Yeah, this was a simple label change - I think fully customising the memory/CPU spec would be a bit more involved.

PR #1402, opened 2017-06-05
Stop etcd VMs trying to monitor ct-mirror log files

RJPercival:
The previous grep would match "ct-mirror" part was through the etcd image name, e.g. "gcr.io/ct-mirror-pilot/etcd". Now, it should only match the intended image, e.g. "gcr.io/ct-mirror-pilot/ct-mirror".

### DIFF
--- a/cloud/google/node_init.sh
+++ b/cloud/google/node_init.sh
@@ -42,11 +42,16 @@ fi
 sudo bash ./${AGENT_INSTALL_SCRIPT}
 
 # Examine what kind of Docker image is on this machine to determine what to log.
-if docker images | grep ct-server; then
+DOCKER_REPOS_FILE=$(mktemp)
+docker images --format '{{ .Repository }}' > "${DOCKER_REPOS_FILE}"
+
+if grep '/ct-server$' "${DOCKER_REPOS_FILE}"; then
   CT_LOGS_PREFIX="${DATA_DIR}/ctlog/logs/ct-server"
-elif docker images | grep ct-mirror; then
+elif grep '/ct-mirror$' "${DOCKER_REPOS_FILE}"; then
   CT_LOGS_PREFIX="${DATA_DIR}/ctmirror/logs/ct-mirror"
 fi
+
+rm "${DOCKER_REPOS_FILE}"
 
 if [[ -n "${CT_LOGS_PREFIX}" ]]; then
   sudo cat > /etc/google-fluentd/config.d/ct-info.conf <<EOF


cloud/google/node_init.sh
```diff
@@ -42,9 +42,9 @@ fi
 sudo bash ./${AGENT_INSTALL_SCRIPT}
 
 # Examine what kind of Docker image is on this machine to determine what to log.
-if docker images | grep ct-server; then
+if docker images | grep '/ct-server$'; then
   CT_LOGS_PREFIX="${DATA_DIR}/ctlog/logs/ct-server"
-elif docker images | grep ct-mirror; then
+elif docker images | grep '/ct-mirror$'; then
```
phad: Are you sure that 'docker images' output will have a line ending ct-mirror ?
When I tried, it didn't:

(on a mirror host)
$ sudo docker images
REPOSITORY TAG IMAGE ID CREATED SIZE
gcr.io/ct-mirror-pilot/ct-mirror test 7643c5ff45ac 11 days ago 273.4 MB
google/cadvisor latest f9ba08bafdea 12 weeks ago 57.34 MB

$ sudo docker images | grep '/ct-mirror$'
$

(on an etcd host)
$ sudo docker images
REPOSITORY                             TAG                 IMAGE ID            CREATED             SIZE
gcr.io/ct-mirror-pilot/etcd            test                bf26662588bc        3 days ago          194.9 MB
google/cadvisor                        latest              f9ba08bafdea        12 weeks ago        57.34 MB
gcr.io/google_containers/pause-amd64   3.0                 99e59f495ffa        13 months ago       746.9 kB
$ sudo docker images | grep '/ct-mirror$'
$ 

Same would apply to the ct-server regexp above.

RJPercival: Good point, I've updated this code to to make `docker images` write only the repository names (since that's what we're examining) to a file and then grep that.

cloud/google/node_init.sh
```diff
@@ -42,12 +42,17 @@ fi
 sudo bash ./${AGENT_INSTALL_SCRIPT}
 
 # Examine what kind of Docker image is on this machine to determine what to log.
-if docker images | grep ct-server; then
+DOCKER_REPOS_FILE=$(mktemp)
+docker images --format '{{ .Repository }}' > "${DOCKER_REPOS_FILE}"
```
phad: Looks good, but does this require a newer Docker than we have on our VMs at present?

$ docker -v
Docker version 1.9.1, build a34a1d5

$ docker images --format '{{ .Repository }}'
flag provided but not defined: --format
See 'docker images --help'.

RJPercival: Only VMs running old images will have too old a version of Docker. Since node_init.sh is only run when a node is being initialized, we can be sure that it will have a new image and therefore a recent version of Docker. For example, running `docker -v` on a more recent VM prints `Docker version 1.11.2, build b9f10c9`.

phad: OK, sounds fine in that case.

PR #1408, opened 2017-08-10
Mutable tree based on API proposal originated by Google

Martin2112:
The new API supports applications that need to be able to modify leaf data. This never happens in CT hence there is an additional interface for those that need it.

An example usage is committing to a large data set that evolves slowly over time. Instead of building a tree each time and then throwing it away and rebuilding it next time the tree can be updated to the current state and the new root hash obtained.

### DIFF
--- a/cpp/merkletree/merkle_tree.cc
+++ b/cpp/merkletree/merkle_tree.cc
@@ -295,11 +295,6 @@ MutableMerkleTree::MutableMerkleTree(unique_ptr<SerialHasher> hasher)
 
 MutableMerkleTree::~MutableMerkleTree() {}
 
-bool MutableMerkleTree::UpdateLeaf(size_t leaf, const string& data) {
-  // Optimized for success: hash computed before validating the leaf's position.
-  return UpdateLeafHash(leaf, treehasher_.HashLeaf(data));
-}
-
 bool MutableMerkleTree::UpdateLeafHash(size_t leaf, const string& hash) {
   if (leaf == 0 || leaf > LeafCount()) return false;
 
--- a/cpp/merkletree/merkle_tree.cc
+++ b/cpp/merkletree/merkle_tree.cc
@@ -336,11 +336,9 @@ bool MutableMerkleTree::Truncate(size_t leaf) {
   if (leaf > LeafCount()) return false;
 
   if (leaf == 0) {
-    if (leaves_processed_ > 0) {
       tree_.clear();
       leaves_processed_ = 0;
       level_count_ = 0;
-    }
 
     return true;
   }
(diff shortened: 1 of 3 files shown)


cpp/merkletree/merkle_tree.h
```diff
@@ -209,4 +209,25 @@ class MerkleTree : public cert_trans::MerkleTreeInterface {
...
+class MutableMerkleTree : public MerkleTree {
+ public:
+  // The constructor takes a pointer to some concrete hash function
+  // instantiation of the SerialHasher abstract class.
+  explicit MutableMerkleTree(std::unique_ptr<SerialHasher> hasher);
+  virtual ~MutableMerkleTree();
+
+  // Update |leaf|th leaf in the tree. Indexing starts from 1.
+  // Returns false if |leaf| is out of bounds.
+  bool UpdateLeaf(size_t leaf, const std::string& data);
```
pphaneuf: Could we leave that one out? It's really easy to do yourself, doesn't bring much to the class (I know the parent has an equivalent, I wish to delete it too!).

cpp/merkletree/merkle_tree.cc
```diff
@@ -289,3 +289,99 @@ void MerkleTree::AddLevel() {
...
+  }
+
+  return true;
+}
+
+bool MutableMerkleTree::Truncate(size_t leaf) {
+  if (leaf > LeafCount()) return false;
+
+  if (leaf == 0) {
+    if (leaves_processed_ > 0) {
```
pphaneuf: Wouldn't this stuff need to be done all the time?

Martin2112: I think the case is truncating the tree to size zero when nothing has been propagated through the lazy evaluation. I'm now suspicious that it doesn't clear the tree in this case because you might have added stuff that has not yet been propagated.

Martin2112: Actually I now don't think that's what happens. When the first leaf is added it sets leaves_processed to 1. So it can only be 0 if nothing has been added, the tree is empty and there are no levels. I think the code is right.

pphaneuf: But why even check this, then? If `leaf == 0`, you can just blow away everything, to get it back to the initial state, that should be straightforward?

I'm not saying that this is wrong, just that this is making me think I don't understand it?

Martin2112: I'm reasonably sure that removing the if here has no effect. It is just avoiding setting up a state that is already present. Arguably removing it helps improve clarity.

cpp/merkletree/merkle_tree.h
Martin2112: Done.

Martin2112:
I think it's all fixed up now?

Martin2112:
How's that? I just ran clang-format with what I think is the right style.

PR #1409, opened 2017-08-10
Temporary workaround for issue with gclient and DEPS

Martin2112:
For the moment just build with OpenSSL until a proper fix can be developed.


Martin2112:
Merging with some un-reviewed fixes to unblock the build.

PR #1450, opened 2018-08-21
Fix issues preventing build on OSX and Travis tests on Linux

Martin2112:
* Fix unused lambda 'this' param error on OSX build
* Build openssl with `no-asm` to fix illegal instruction error on AES tests on Travis

### DIFF
--- a/build/Makefile.openssl
+++ b/build/Makefile.openssl
@@ -11,6 +11,6 @@ $(INSTALL_DIR)/bin/openssl: Makefile
 
 Makefile: config Configure Makefile.org
 	# Force a 64 bit build on MacOS
-	KERNEL_BITS=64 ./config --install_prefix=$(INSTALL_DIR) --prefix=/ --libdir=lib --openssldir=$(OPENSSLDIR) no-shared enable-static-engine -fPIC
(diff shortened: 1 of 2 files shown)


build/Makefile.openssl
```diff
@@ -11,6 +11,6 @@ $(INSTALL_DIR)/bin/openssl: Makefile
 
 Makefile: config Configure Makefile.org
 	# Force a 64 bit build on MacOS
-	KERNEL_BITS=64 ./config --install_prefix=$(INSTALL_DIR) --prefix=/ --libdir=lib --openssldir=$(OPENSSLDIR) no-shared enable-static-engine -fPIC
+	KERNEL_BITS=64 ./config --install_prefix=$(INSTALL_DIR) --prefix=/ --libdir=lib --openssldir=$(OPENSSLDIR) no-asm no-shared enable-static-engine -fPIC
```
AlCutter: Should we try to contain this to Travis builds?
(See https://github.com/google/certificate-transparency/blob/master/build/Makefile.glog for a possible approach.)

Martin2112: Yeah I suppose we should.

Martin2112: I think that should do it.

PR #1453, opened 2018-09-06
Use absl.app in print_log_list.py and improve flag checks

RJPercival:
Using absl.app causes flags to be handled automatically and gives us nice things like `app.UsageError`. I've added extra checks for unspecified flags, since otherwise the error messages caused by these flags being unset were quite obtuse.

I've removed the default value for `--log_list_schema` since there is no schema file found at "data/log_list_schema.json".

### DIFF
--- a/python/utilities/log_list/print_log_list.py
+++ b/python/utilities/log_list/print_log_list.py
@@ -8,6 +8,7 @@ import os
 import sys
 import time
 
+from absl import app
 from absl import flags as gflags
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes, serialization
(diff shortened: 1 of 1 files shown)


python/utilities/log_list/print_log_list.py
```diff
@@ -8,7 +8,7 @@
 import sys
 import time
 
-from absl import flags as gflags
+from absl import app, flags as gflags
```
pphaneuf: I think we'd want those imports on separate lines.

RJPercival: Done.

python/utilities/log_list/print_log_list.py
```diff
@@ -89,19 +87,27 @@ def print_formatted_log_list(json_log_list):
                 final_sth["tree_size"])
         print "-" * 80
 
-def run():
+
+def main(argv):
+    del argv  # Unused; all arguments provided via flags.
```
pphaneuf: Usually, what we do with unused flags like here is name them `unused_argv` (which shuts up the linter).

RJPercival: My linter sadly does complain if I name it `unused_argv`. If I prefix it with an underscore, it's happy though.

